PR #2, opened 2019-11-11
Remove redundant Openshift refs

skonto:
-  removes `install-oc-plugin` command from the `kubectl-cloudflow` plugin. We should have more platform agnostic code, that is the rationale behind this removal.
- fixes output in the ImagePlugin and in several examples.

Notes: 
1) this [part](https://github.com/lightbend/cloudflow/blob/89ce8a7416156ec2a05706e2303f2c0831ed4ac5/core/sbt-cloudflow/src/main/scala/cloudflow/sbt/ImagePlugin.scala#L131-L137)  will be fixed in another PR where we should remove the default lightbend namespace dependency. The latter causes problems since now the the namespace for the protocol version configmap cloudflow-protocol-version is hardcoded to `lightbend` as @skyluc has mentioned before.

2) We have some [special logic](https://github.com/lightbend/cloudflow/search?l=Scala&q=openshift) for handling Openshift cases, we should not remove that as we will support Openshift in the future as part of the deployment options.

@maasg @debasishg what are your thoughts for 2?


debasishg:
I agree with the philosophy. We decided to make cloudflow platform agnostic.

maasg:
+1

PR #9, opened 2019-11-20
Publish to Sonatype

debasishg:
This PR does the necessary changes to publish artifacts to Sonatype. However before this can be used, a few local settings need to be in place:

1. In folder `~/.sbt/1.0` create a file named `sonatype.sbt`:
```
$ cat sonatype.sbt
credentials += Credentials(Path.userHome / ".sbt" / "sonatype.credentials-lb")
```

2. In the path mentioned in the above `sonatype.sbt` create the credentials file with your sonatype credentials:

```
$ cat sonatype.credentials-lb
realm=Sonatype Nexus Repository Manager
host=oss.sonatype.org
user=<your sonatype user id>
password=<your sonatype password>
```
Then the publishing can be done as ..

```
$ sbt publishSigned
```
* In order for the above to work you need to have PGP signing set up as per the instructions in https://www.scala-sbt.org/1.x/docs/Using-Sonatype.html
* Also in order for the whole process to work you need to have your public keys uploaded to a key server and it may take a day or so to sync your keys across.

**Note:** I am still facing some problems with entering the passphrase from within sbt. The outputs of gpg and sbt get garbled. Not sure if it is a Mac specific issue - I tried using `pinentry-mac`, `pinentry-tty` but getting the same result.


core/sonatype.sbt
```diff
@@ -0,0 +1,48 @@
...
+
+ThisBuild / developers := List(
+  Developer(id="agemooij", name="Age Mooij", email="age.mooij@gmail.com", url=url("http://lightbend.com")),
+  Developer(id="RayRoestenburg", name="Raymond Roestenburg", email="raymond.roestenburg@gmail.com", url=url("http://lightbend.com")),
+  Developer(id="olofwalker", name="Robert Walker", email="robert@walker.st", url=url("http://lightbend.com")),
+  Developer(id="maasg", name="Gerard Maas", email="gerard.maas@gmail.com", url=url("http://lightbend.com")),
+  Developer(id="debasishg", name="Debasish Ghosh", email="dghosh@acm.org", url=url("http://lightbend.com")),
+  Developer(id="seglo", name="Sean Glover", email="sean@seanglover.com", url=url("http://lightbend.com")),
+  Developer(id="skyluc", name="Luc Bourlier", email="luc.bourlier@typesafe.com", url=url("http://lightbend.com")),
+  Developer(id="skonto", name="Stavros Kontopoulos", email="skonto@users.noreply.github.com", url=url("http://lightbend.com")),
```
skonto: @debasishg my email is st.kontopoulos@gmail.com on github now visible. Could you update this entry?

debasishg: sure .. will do that @skonto ..

core/build.sbt
```diff
@@ -451,11 +444,7 @@ def cloudflowModule(moduleID: String): Project = {
 
 lazy val commonSettings = Seq(
   organization := "com.lightbend.cloudflow",
-  bintrayOrganization := Some("lightbend"),
-  bintrayRepository := altBintrayRepository.value.getOrElse("cloudflow"),
   headerLicense := Some(HeaderLicense.ALv2("(C) 2016-2019", "Lightbend Inc. <https://www.lightbend.com>")),
```
maasg: [OT for this PR] is this a left-over from the header plugin? Do we still need this?

debasishg: @maasg I will take a look and fix it in a future PR

PR #23, opened 2019-11-26
Adds a link to the cluster installer

laylamaas:

Add a link to the installer repo

Need info on how to get the cluster setup

No

I read it over

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,5 +74,6 @@ Cloudflow aims to make this easier by integrating the most popular streaming fra
 
 * [Getting Started](docs/Getting%20Started.md)
 * Example / walk-through
+* [Creating a Cloudflow cluster on Kubernetes](https://github.com/lightbend/cloudflow-installer) 
 * Deploying to a cluster
 * Contributors


lightbend-cla-validator:
Hi @laylamaas,

Thank you for your contribution! We really value the time you've taken to put this together.

Before we proceed with reviewing this pull request, please sign the Lightbend Contributors License Agreement:

[http://www.lightbend.com/contribute/cla](http://www.lightbend.com/contribute/cla)

README.md
```diff
@@ -74,5 +74,6 @@ Cloudflow aims to make this easier by integrating the most popular streaming fra
 
 * [Getting Started](docs/Getting%20Started.md)
 * Example / walk-through
+* [Creating a Cloudflow cluster on Kubernetes](https://github.com/lightbend/cloudflow-installer) 
 * Deploying to a cluster
```
skonto: maybe then we should remove this one. My thinking was to leave only the Getting started document which has a section for this [here](https://github.com/lightbend/cloudflow/blob/master/docs/Getting%20Started.md#install-a-gke-cluster). Alternatively we could break down the `Getting Started` guide to separate documents:
a) `Prerequisites` eg. Cloudflow CLI installation and setting up Cloudflow runtime on GKE.
b) Build/deploy/undeploy an app.

debasishg: @skonto I agree .. we should not repeat. Deploying to a Cluster is explained in details in Getting Started - we can refer to that. But if the deployment for this app needs some special considerations like `ConfigParams` or pvc-claim, then I think we should add the deploy command here (in addition to the link). WDYT ?

maasg: Agree. I'll clean that up.

maasg:
Thanks for your contribution. We have removed the need for the installer repo, which is now under `/installer`. I'll close this PR and update the docs accordingly.

PR #26, opened 2019-11-28
Adds a custom release pipeline that enables single shot publish to Sonatype

debasishg:

* Adds a custom release pipeline through `sbt-release`
* Now uses single shot Sonatype publish using `sonatypeBundleRelease` capability of `sbt-sonatype`
* Upgrades version to `1.3.0-M4` (`1.3.0-M3` is published to maven Central)
* Removed doc on publishing artifacts - will be moved to the release process of cloudflow

* The changes done make the Sonatype publish process much easier. Previously the process of promoting the repository from staging at Sonatype was completely manual. This new approach makes it fully automated. Also this makes the release process much easier.

* No

* The patch has been tested as part of the release process


docs/publishing-artifacts.md
```diff
@@ -49,10 +49,9 @@ credentials += Credentials(Path.userHome / ".sbt" / "sonatype.credentials")
...
-* Once you have verified all contents of the staging repository, close the repository using the close button in the UI. Closing the staging repository will start a process to validate all contents and check for compliance with Maven Central. If all checks are ok, the staging repository will be closed successfully.
-* The next step is to do the release using the release button. This will release all artifacts to the Sonatype release repository. The released artifacts are now ready to be transferred to Maven Central. This may take some time and will be synced with Maven Central the next time this sync process runs.
+* In order to publish signed artifacts locally, just do `sbt publishLocalSigned`
+* In order to do a publish of signed artifacts to Sonatype and then subsequently getting it sync'd with Maven Central, just do the following 2 steps:
+ * `sbt publishSigned`, which will ask for the GPG passphrase before signing
+ * `sbt sonatypeBundleRelease`, which will do the actual publish to staging repository, promote it to final repository and subsequently making the artifacts available for Maven Central sync
```
agemooij: I'm a little bit confused by this. If we tell ourselves that these are the two steps, then what is sbt-release for? Should we always use `sbt release` to publish signed artifacts to Maven Central?

debasishg: @agemooij These 2 steps are only for the streamlet libraries to be pushed to Maven Central. But the release process involves a lot more .. in fact this is **one of the steps** in the release .. 

* pushing operator image to dockerhub
* releasing `kubectl-cloudflow` to bintray
* preparing docker images of `examples` with appropriate plugin versions and pushing them to dockerhub
* releasing helm charts
* releasing docs

agemooij: So perhaps the READMe should be a bit more clear about when to use what.

debasishg: @agemooij Added a note on this at the end of this documentation

debasishg: @agemooij The steps are exactly similar to what we do in `pipelines-release` where we use the default pipeline from `sbt-release`. We need this customization here for the Sonatype dance

PR #34, opened 2019-11-29
Simplified steps to push data to the application

jeroenr:
This gets rid of some obsolete steps when pushing data to the application. 

1. Just using the provided bash script with the provided test-data is sufficient
2. Use a one-liner for the port-forwarding to remove the need for looking up and copying the pod id

### DIFF
--- a/core/cloudflow-akka/src/main/scala/cloudflow/akkastream/AkkaStreamlet.scala
+++ b/core/cloudflow-akka/src/main/scala/cloudflow/akkastream/AkkaStreamlet.scala
@@ -33,6 +33,24 @@ import net.ceedubs.ficus.Ficus._
 abstract class AkkaStreamlet extends Streamlet[AkkaStreamletContext] {
   final override val runtime = AkkaStreamletRuntime
 
+  // ctx is always first set by runner through `init` so this is safe.
+  @volatile private var ctx: AkkaStreamletContext = null
+
+  /**
+   * Returns the [[StreamletContext]] in which this streamlet is run. It can only be accessed when the streamlet is run.
+   */
+  protected final implicit def context: AkkaStreamletContext = {
+    if (ctx == null) throw new StreamletContextException()
+    ctx
+  }
+
+  /**
+   * Java API
+   *
+   * Returns the [[StreamletContext]] in which this streamlet is run. It can only be accessed when the streamlet is run.
+   */
+  protected final def getStreamletContext(): AkkaStreamletContext = context
+
   /**
    * Initialize the streamlet from the config. In some cases (e.g. the tests) we may pass a context
    * directly to be used instead of building it from the config.
--- a/core/cloudflow-flink/src/main/scala/cloudflow/flink/FlinkStreamlet.scala
+++ b/core/cloudflow-flink/src/main/scala/cloudflow/flink/FlinkStreamlet.scala
@@ -71,7 +71,10 @@ import cloudflow.streamlets._
 abstract class FlinkStreamlet extends Streamlet[FlinkStreamletContext] with Serializable {
   final override val runtime = FlinkStreamletRuntime
 
-  private val readyPromise      = Promise[Dun]()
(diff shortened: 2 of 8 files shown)


docs/Getting Started.md
```diff
@@ -702,51 +702,16 @@ http-ingress      sensor-data-http-ingress-fd9cdb66f-jbsrm    Running
...
-  sensor-data-http-ingress:
-    Container ID:  docker://9149cd757094e7ea1b943076048b7efc7aa343da8c2d598bba31295ef3cbfd6b
-    Image:         eu.gcr.io/bubbly-observer-178213/sensor-data@sha256:ee496e8cf3a3d9ab71c3ef4a4929ed8eeb6129845f981c33005942314ad30f18
-    Image ID:      docker-pullable://eu.gcr.io/bubbly-observer-178213/sensor-data@sha256:ee496e8cf3a3d9ab71c3ef4a4929ed8eeb6129845f981c33005942314ad30f18
-    Ports:         3003/TCP, 2048/TCP, 2049/TCP, 2050/TCP
-...
-```
-
-So one of the POD ports is `3003` - let's set up a port forwarding on it.
-
```
agemooij: The following command assumes that the user knows that the port to forward is 3003, so we cannot remove this section, right?

jeroenr: Right, I assumed that the port in the example would be fixed

jeroenr: Do you think it makes sense to clearly label the port so we can filter for it? Would be nice to prevent the user from having to scroll through the describe output to find the port

agemooij: We could probably simplify the command to only extract the container ports so the output is more readable.

jeroenr: Wouldn't something like this be sufficient?
```
$ kubectl -n sensor-data get po -lcom.lightbend.cloudflow/streamlet-name=http-ingress -o jsonpath="{.items[0].spec.containers[0].ports[0].containerPort}"
```

lightbend-cla-validator:
At least one pull request committer is not linked to a user. See https://help.github.com/en/articles/why-are-my-commits-linked-to-the-wrong-user#commits-are-not-linked-to-any-user

docs/Getting Started.md
olofwalker: @jeroenr Picking this up, I think this is sufficient

olofwalker:
@jeroenr If you can rebase this PR to master I think we can merge it! Also, very sorry about the lead time on this. We are now better organized internally to follow up PR's with a proper support rotation starting this week, so this should be avoidable in the future.

jeroenr:
@olofwalker Thanks for picking this up. I could rebase, but it looks like your docs are moved

olofwalker:
> @olofwalker Thanks for picking this up. I could rebase, but it looks like your docs are moved

@jeroenr It should be in the same file (`Getting Started.md`) on line 747:ish ?

maasg:
@jeroenr @olofwalker I took a brief look last Thursday, wanting to move this forward but it was more work than the time I had. 
To merge the value of this PR, we need two things:
- indeed the docs are moved. These updated instructions should go in https://cloudflow.io/docs/current/get-started/exercise-example.html
- we need an example in our repo that's exactly replicates the documentation. The current sibling, `sensor-data-scala` contains an additional file ingress.
The idea would be to separate `sensor-data-scala` into two examples: 
- one that matches the docs, so that people can just `clone` the example, and
- one 'file ingress sample' that shows how to use the ingress

We also discussed that we should split the current `examples/` dir into _sample code_ and _sample apps_ (names to be discussed. We know that that's the hard part ;-) )
`sensor-data-scala` and `file-ingress-sample` should become part of that _sample code_ while `call-record-app` should be in the applications.

We probably need to break this work into smaller, individual chunks.

jeroenr:
Closing this since the changes involved don't seem to outweigh the rebasing work

Issue #43, opened 2019-11-29
[operator] NoClassDefFoundError: com/fasterxml/jackson/databind/Module at runtime


agemooij:
**Describe the bug**
After building the operator docker image, pushing it to a repo, and adjusting the installation script to use it, Gitter user @mrooding ran into a runtime error: `java.lang.NoClassDefFoundError: com/fasterxml/jackson/databind/Module`

**To Reproduce**
- build the operator docker image using `sbt docker` in the operator sub-project
- tag and push the resulting image to a registry
- adjust the installation script to use the above image
- install

Check the logs

**Expected behavior**
No errors

**Screenshots**

**Additional context**

skonto:
I was able to reproduce this by building a custom imag and trying to deploy Cloudflow:
```
2019-11-29 17:45:18,117 INFO  [ActorSystemImpl] - Connected to Kubernetes cluster: https://10.3.240.1:443
Uncaught error from thread [default-akka.actor.default-dispatcher-2]: com/fasterxml/jackson/databind/Module, shutting down JVM since 'akka.jvm-exit-on-fatal-error' is enabled for ActorSystem[default]
java.lang.NoClassDefFoundError: com/fasterxml/jackson/databind/Module
	at play.api.libs.json.StaticBinding$.parseJsValue(StaticBinding.scala:16)
	at play.api.libs.json.Json$.parse(Json.scala:171)
	at skuber.json.PlayJsonSupportForAkkaHttp.$anonfun$unmarshaller$2(PlayJsonSupportForAkkaHttp.scala:83)
	at akka.http.scaladsl.util.FastFuture$.$anonfun$map$1(FastFuture.scala:23)
	at akka.http.scaladsl.util.FastFuture$.strictTransform$1(FastFuture.scala:41)
	at akka.http.scaladsl.util.FastFuture$.transformWith$extension1(FastFuture.scala:45)
	at akka.http.scaladsl.util.FastFuture$.map$extension(FastFuture.scala:23)
	at akka.http.scaladsl.unmarshalling.Unmarshaller.$anonfun$map$3(Unmarshaller.scala:26)
	at akka.http.scaladsl.unmarshalling.Unmarshaller.$anonfun$transform$3(Unmarshaller.scala:23)
	at akka.http.scaladsl.unmarshalling.Unmarshaller$$anon$1.apply(Unmarshaller.scala:58)
[...]

Issue #6, opened 2019-11-19
Namespace for protocol version configmap is hardcoded


skonto:
**Describe the bug**
The namespace for the protocol version configmap cloudflow-protocol-version is hardcoded to lightbend.
If the lightbend namespace doesn't exist, the configmap is not created, and the CLI cannot be used.
It should not be hardcoded.

**To Reproduce**
Install Cloudflow on a namespace different than lighbend (can be changed in [install-gke.sh](https://github.com/lightbend/cloudflow-installer/blob/ff701f9e4aa313f55712d7fabba16c44eec852be/install-gke.sh#L36)).

Run kubectl cloudflow list.

It fails with
```
[Error] Cannot find the 'cloudflow-protocol-version' configmap, please make sure that the Cloudflow operator is installed
```
***workaround***: create a `lightbend` namespace, and kill the `cloudflow-operator` pod to restart it.

**Expected behavior**
The configmap to be installed and looked for in the current namespace.

**Screenshots**
N/A
**Additional context**
N/A

Thank you @skyluc for reporting this.

mrooding:
This can be closed since #45 was merged?

RayRoestenburg:
I agree, this can be closed. @skonto @debasishg

PR #44, opened 2019-11-29
Fix operator dependency

debasishg:
### What changes were proposed in this pull request?
* Fix reported operator dependency [problem](https://github.com/lightbend/cloudflow/issues/43)

### Why are the changes needed?
* The changes are needed to fix the `NoClassDefFoundError` as reported in the issue (linked above)

### Does this PR introduce any user-facing change?
*  Fixes the problem of operator not starting

### How was this patch tested?
* Tested by installing cloudflow in a GKE cluster and ensuring that `cloudflow-operator` runs ok

PR #45, opened 2019-11-29
Remove hardcoded protocol version configmap namespace

mrooding:
I removed the hardcoded namespace for the protocol version config map. I've also added a small section in the docs about building and testing a modified `cloudflow-operator` container on GKE.

As described in #6, if a user changes the namespace used during the installation the protocol version config map will not be created and the installation will be unusable without applying a workaround.

No

As discussed on Gitter, I wanted to test the updated `cloudflow-operator` image but couldn't because of #43


RayRoestenburg:
@mrooding you also need to change the kubectl cloudflow plugin, please see `/Users/ray/src/cloudflow/kubectl-cloudflow/version/version.go`, `FailOnProtocolVersionMismatch` function. The protocol version is a contract between the kubectl plugin and the operator.

The namespace was fixed because of this reason. `kubectl cloudflow` just checks for the configmap in the lightbend namespace. 
Another way to do this is for kubectl to find the configmap by name or by some label, in all namespaces. It does not know which namespace the operator is installed in.

Another thing I was not sure about, is where the configmap will get created if no namespace is provided to skuber. but it looks like this will be the namespace the pod is running in, based on https://github.com/doriordan/skuber/blob/c8c64ae753016d0b683ca765454d8158b9ac9281/client/src/main/scala/skuber/api/Configuration.scala#L241 and the links in the comments above it, so that is all good, (and you probably tested that it did get created there)

core/README.md
```diff
@@ -1,5 +1,34 @@
 # Cloudflow Streamlets Library
 
+## Building and testing the Cloudflow Operator image
```
RayRoestenburg: 👍

mrooding:
Hi Ray, thanks for pointing out the reference in the CLI tool. From what I read about `client-go`, looking for the config map in all namespaces should be as simple as changing `k8sClient.CoreV1().ConfigMaps(ProtocolVersionNamespace)` to `k8sClient.CoreV1().ConfigMaps("")`. 

I wasn't able to end-to-end test this feature branch yet due to #44 as discussed on Gitter. I'll update the cloudflow CLI and test it.

W.r.t. the namespace used by skuber, I think you're actually right, and when I run the test we'll get this confirmed. When I was working on the changes to the operator image, I noticed that the `DeploymentContext` setting knows the pod namespace, so I actually thought about using that to construct the skuber client.

I'll get back to you when I was able to run the test

RayRoestenburg:
Thanks @mrooding ! sounds like you got a solution, let's see what happens when you test it.

olofwalker:
@mrooding Yesterday we had to do a quick fix for the hardcoded namespace, changing it from `lightbend` to `cloudflow`, due to some internal scheduling issues. I also tested this PR and I think we have to change it a bit, first making a list of all the configmaps in the system

```
configMaps, err := k8sClient.CoreV1().ConfigMaps("").List(metav1.ListOptions{})
```

and then finding the one we are looking for (or not).

mrooding:
@olofwalker thanks for testing the PR. I didn't get to it yet. Can you tell me what you ran into when testing the PR. As mentioned in a comment before, looking for the config map in all namespaces should be as simple as using `k8sClient.CoreV1().ConfigMaps("")`. However, I've updated the PR to use your proposed solution.

I've also used the `podNamespace` to set up the config map in the operator.

I most likely won't be able to test it until this weekend, at the earliest. Let me know if you do have a chance to try it out.

core/cloudflow-operator/src/main/scala/cloudflow/operator/Main.scala
```diff
@@ -121,16 +121,14 @@ object Main extends {
     )
   }
 
-  private def installProtocolVersion(client: skuber.api.client.KubernetesClient)(implicit ec: ExecutionContext): Unit = {
+  private def installProtocolVersion(client: skuber.api.client.KubernetesClient, podNamespace: String)(implicit ec: ExecutionContext): Unit = {
```
olofwalker: This function now uses two different K8s clients, if we re-write it this way, we can do away with the risk of this problem repeating in the future.
```
private def installProtocolVersion(client: skuber.api.client.KubernetesClient)(implicit ec: ExecutionContext): Unit = {
    val protocolVersionTimeout = 20.seconds

    Await.ready(
      client.getOption[ConfigMap](Operator.ProtocolVersionConfigMapName).map {
        _.fold(client.create(Operator.ProtocolVersionConfigMap)) { configMap ⇒
          if (configMap.data.getOrElse(Operator.ProtocolVersionKey, "") != Operator.ProtocolVersion) {
            client.update(configMap.copy(data = Map(Operator.ProtocolVersionKey -> Operator.ProtocolVersion)))
          } else {
            Future.successful(configMap)
          }
        }
      },
      protocolVersionTimeout
    )
  }
```

kubectl-cloudflow/version/version.go
```diff
@@ -40,26 +36,34 @@ func FailOnProtocolVersionMismatch() {
...
-						util.LogAndExit("This version of kubectl cloudflow is not compatible with the Cloudflow operator, please upgrade kubectl cloudflow")
-					}
-					if version > supportedVersion {
-						util.LogAndExit("This version of kubectl cloudflow is not compatible with the Cloudflow operator, please upgrade the Cloudflow operator")
-					}
-				}
+	if configMaps, err := k8sClient.CoreV1().ConfigMaps("").List(metav1.ListOptions{}); err == nil {
+		for _, v := range configMaps.Items {
+			if v.Name == ProtocolVersionConfigMapName {
+				cm = &v
```
olofwalker: We need to change this, this actually takes the address of `v`, which will change during the iterations of the loop, so it will always point towards the last item in the slice.

We can fix this issue by instead taking the address of the item in the slice, like this:

```
	for i, v := range configMaps.Items {
			fmt.Println(v.Name)
			if v.Name == ProtocolVersionConfigMapName {
				cm = &configMaps.Items[i]
			}
		}```

**Comment by olofwalker** [comment]
@mrooding I've tested your PR now and found a few issues preventing the application from deploying correctly. If these changes are made to the PR we can go ahead and merge it.

File: core/cloudflow-operator/src/main/scala/cloudflow/operator/Main.scala (hunk 2 above)
Comment (olofwalker) [reply core/cloudflow-operator/src/main/scala/cloudflow/operator/Main.scala]: The function call would look like this after the change:

```
installProtocolVersion(client.usingNamespace(settings.podNamespace))
```

mrooding:
@olofwalker thank you for taking the time to test it. Sharp observations, I've just updated the PR

olofwalker:
@mrooding Thank you, I have tested the PR and it worked just fine. Thanks again for your contribution!

PR #49, opened 2019-12-02
Made ctx volatile.

RayRoestenburg:
Minor fix, making `ctx` volatile in `AkkaStreamlet`. Thanks to @hepin1989 for the suggestion.

It is a (minor) improvement (it's more safe).

No

Compiled.

### DIFF
--- a/core/cloudflow-akka/src/main/scala/cloudflow/akkastream/AkkaStreamlet.scala
+++ b/core/cloudflow-akka/src/main/scala/cloudflow/akkastream/AkkaStreamlet.scala
@@ -36,7 +36,7 @@ abstract class AkkaStreamlet extends Streamlet {
   final override val runtime = AkkaStreamletRuntime
 
   // ctx is always first set by runner through `init` so this is safe.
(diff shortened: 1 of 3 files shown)


core/cloudflow-akka/src/main/scala/cloudflow/akkastream/AkkaStreamlet.scala
```diff
@@ -36,7 +36,7 @@ abstract class AkkaStreamlet extends Streamlet {
   final override val runtime = AkkaStreamletRuntime
 
   // ctx is always first set by runner through `init` so this is safe.
-  private var ctx: AkkaStreamletContext = null
+  @volatile private var ctx: AkkaStreamletContext = null
```
debasishg: @RayRoestenburg What about Spark and Flink streamlets ? Should we not make the context `volatile` there ?

agemooij: Good point!

Issue #53, opened 2019-12-03
Add Bash version 4 as cloudflow install requirement


nolangrace:
**Describe the bug**
MacOS default version of bash is still v3 which causes a problem when running the install-gke.sh script.

**To Reproduce**
Switch to any v3 version of bash and run the install-gke.sh script.

**Expected behavior**
The documentation should reflect that v4 or higher of bash is required to run the cloudflow installer and if possible the installer should check the bash version

[Cloudflow-on-bashv3.2.57(1)-release.txt](https://github.com/lightbend/cloudflow/files/3918617/Cloudflow-on-bashv3.2.57.1.-release.txt)

PR #56, opened 2019-12-04
Checks if bash version >= 4 for gke install of cloudflow

debasishg:

* Checks if bash version is >= 4 for gke install of cloudflow

* Fixes https://github.com/lightbend/cloudflow/issues/53

* Yes. It changes some validations in the installer

* Tested by installing cloudflow on a gke cluster

### DIFF
--- a/installer/common/requirements.sh
+++ b/installer/common/requirements.sh
@@ -16,4 +16,8 @@
 
 # Perform checks of utilities required to perform the installation
 
+if [ "${BASH_VERSINFO}" -lt 4 ] | [ "${BASH_VERSINFO[0]}" -lt 4 ]; then
+	echo "Bash version must be >= 4 in order to run the Cloudflow installer script for GKE"
+	exit 1
+fi
 hash jq 2>/dev/null || { print_error_message >&2 "'jq' is required but it's not installed."; exit 1; }


installer/common/requirements.sh
```diff
@@ -16,4 +16,8 @@
 
 # Perform checks of utilities required to perform the installation
 
+if [ "${BASH_VERSINFO}" -lt 4 ] | [ "${BASH_VERSINFO[0]}" -lt 4 ]; then
```
skonto: LGTM. I assume the left part covers Linux and the right OS X.

agemooij: Here's the output of the check on MacOS after `brew install bash` to get the latest instead of the builtin version:

```
$ echo $BASH_VERSINFO
5
```

debasishg: @agemooij Does the install go through with bash 5 in your MacOS ?

RayRoestenburg: I get nothing when I do that (macOS):
```

Comment (RayRoestenburg) [reply installer/common/requirements.sh]: BASH_VERSINFO is not set for me @debasishg but I do have bash 5 installed (macOS), so I'm not sure if this will always work:
```
> bash --version
GNU bash, version 5.0.11(1)-release (x86_64-apple-darwin19.0.0)
Copyright (C) 2019 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>

This is free software; you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.
> echo $BASH_VERSINFO

```

Comment (debasishg) [reply installer/common/requirements.sh]: @RayRoestenburg Can u pls check what u get with `echo $BASH_VERSION` - I can use that too but just have to do some additional shell stuff to extract the exact version.

Comment (debasishg) [reply installer/common/requirements.sh]: I have version 4 installed and I get ..

```
$ echo $BASH_VERSINFO
4
```

Comment (debasishg) [reply installer/common/requirements.sh]: @RayRoestenburg Can u pls check what u get with `echo ${BASH_VERSINFO[0]}` ?

Comment (skonto) [reply installer/common/requirements.sh]: @RayRoestenburg you should have[ `${BASH_VERSINFO[0]}`](https://stackoverflow.com/questions/29060076/how-to-get-bash-version-number-in-os-x) set.

Comment (skonto) [reply installer/common/requirements.sh]: @debasishg another more robust option I suspect is: 
```BASHFULL=$(bash --version | head -1 | awk '{print $4}')
${BASHFULL%%.*}```

Comment (debasishg) [reply installer/common/requirements.sh]: @skonto yeah .. I tried that too .. also this .. 

```
$ ver=$(echo $BASH_VERSION | cut -d'.' -f1)
$ echo $ver
4
```

agemooij: The script is run using `/usr/bin/env bash` so then the context s bash and you will get those env vars set correctly. They will not work if you execute them from another shell, like my default shell, which is zsh.

Issue #31, opened 2019-11-29
Add support for Helm 3


mrooding:
**Is your feature request related to a problem? Please describe.**
Currently, the GKE installer relies on Helm 2. Since Helm 3, Tiller is no longer required and `helm init` no longer exists as a CLI command. I've already created a PR to include dependencies required to run the installer in #30
 
**Describe the solution you'd like**
Either offer support for both Helm 2 and 3 or migrate to 3 only. I'm willing to work on a PR but I'd first like to know which approach you prefer.

**Describe alternatives you've considered**
I haven't looked into how Helm is being used by Cloudflow extensively but if it's purely for templating then maybe switching to Kustomize is an option

olofwalker:
@mrooding Thanks for taking the time to write this issue. Things are currently in the works when it comes to improving the installation of Cloudflow, making this issue redundant. I will close it for now.

PR #63, opened 2019-12-06
Support both Helm 2 and 3

mrooding:
I've added support for both Helm v2 and v3 as proposed in #31

- I've moved all Helm 2 related bash functions from `tiller.sh` to `helm2.sh`
- I've introduced a `HELM_TIMEOUT` env var since Helm 3 changed how the `--timeout` flag is to be used. 
- Helm 3 also changed the output of `helm version`. The init script that I've added supports both and has an else case for when no version can be detected.
- I've removed the tiller check at the start of `shared.sh`. When using `install-gke.sh` the status of either Helm 2 or 3 is verified. The only use case where the additional check in `shared.sh` made sense if someone was to execute `gke-install-cloudflow.sh` directly. If you want, I can add a check to `shared.sh` to verify either Helm 2 or 3.
- I've generalised the Chart used for testing the installation from `tiller-test-chart` to `helm-test-chart` and applied it to verify both the Helm 2 and 3 installation.

To support users that have Helm v3 installed

Yes, it will allow users with both Helm v2 and v3 to use the GCP installer

I've tested the installation on a GCP cluster with both Helm v2 and v3


skonto:
thank you @mrooding, awesome!

skyluc:
FYI: the `uninstall-cloudflow.sh` script also need to be updated to support Helm 3.
The `helm delete` command doesn`t have a `--purge` option anymore. `purge` is the default behavior on Helm 3.

mrooding:
> FYI: the `uninstall-cloudflow.sh` script also need to be updated to support Helm 3.
> The `helm delete` command doesn`t have a `--purge`option anymore.`purge` is the default behavior on Helm 3.

Thank you for mentioning this. I've added support for both version in the uninstaller in the latest commit.

debasishg:
@mrooding This may take some time to 👀 .. we have a holidays shutdown till Jan 1, 2020.

mrooding:
@debasishg no problem. Enjoy the holidays!

yuchaoran2011:
@mrooding Thanks again for the PR. It LGTM. We are discussing Helm 3 migration strategy internally within the team. In the meantime, could you please update the PR to resolve the conflicts. We'll merge when ready.

yuchaoran2011:
@mrooding We have decided to support both Helm 2 and 3 binaries in Cloudflow as you implemented in this PR. Could you resolve the conflicts and then we'll merge?

mrooding:
@yuchaoran2011 I've just rebased on top of master.

Issue #51, opened 2019-12-03
Create Cluster GKE script should check for Zone and Region setup


nolangrace:
**Describe the bug**
For whatever reason my gcloud local configuration did not have Zone or Region setup although the project did have a default Zone and region.  This resulted in the error below being thrown multiple times when trying to setup the GKE Cluster

ERROR: (gcloud.container.clusters.create) One of [--zone, --region] must be supplied: Please specify location.

**To Reproduce**
Run the commands below to clear local Zone and Region settings
> gcloud config unset compute/zone
> gcloud config unset compute/region

Then run the GKE Cluster Setup Comand
> ./create-cluster-gke.sh cloudflow-cluster

**Expected behavior**

I would expect that the create-cluster-gke.sh script would check if the Zone and Region local configs are set.  If they are not present the script should indicate what the problem is and print how to fix it.

**Additional context**

Adding the missing configs involves running the commands below to set the Zone and Region
> gcloud config set compute/zone ZONE
> gcloud config set compute/region REGION

mrooding:
This can be closed after #64 right?

yuchaoran2011:
@mrooding Sorry for not noticing your reminder until today. Closing it now.

PR #64, opened 2019-12-06
Verify that a compute region and zone are set in the gcloud config

mrooding:
Check if the compute/zone and compute/region properties are set in the Gcloud configuration as mentioned in #51 

Make the installation more user-friendly by telling the user what to do instead of confronting them with a gcloud error

Yes, additional error messages that are possibly displayed during the creation of a GCP cluster

Verified the functionality by unsetting the gcloud config values.

### DIFF
--- a/installer/create-cluster-gke.sh
+++ b/installer/create-cluster-gke.sh
@@ -23,6 +23,22 @@ if [ $# -eq 0 ]
     exit 1
 fi
 
+gcloudZone=$(gcloud config get-value compute/zone)
+if [ "$gcloudZone" == "" ]
+  then
+    echo "No compute/zone set in your GCloud configuration"
+    echo "Please set a compute zone by running: gcloud config set compute/zone VALUE [optional flags]"
+    exit 1
+fi
+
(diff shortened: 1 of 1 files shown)


installer/create-cluster-gke.sh
```diff
@@ -23,6 +23,22 @@ if [ $# -eq 0 ]
     exit 1
 fi
 
+gcloudZone=$(gcloud config get-value compute/zone)
```
skonto: Maybe we could pass these values as parameters to the installer script and then you can always add `--zone` and `--region` to gcloud command. Then we could have multiple clusters setup independent of the default config per [project id](https://cloud.google.com/compute/docs/gcloud-compute/#default-properties).

mrooding: That would definitely make it more flexible. I can do another PR in which I add 2 mandatory parameters for `--zone` and `--region`. Would we then want to remove the gcloud check for zone and region again and simply restrict the user to always supply the parameters? Alternatively, we could make the 2 parameters optional and fallback to the gcloud configuration

PR #66, opened 2019-12-06
Unique group id support. 

blublinsky:

It introduces an additional streamlet type, that reads data from all partitions

This effectively implements broadcast to all instances of akka streams based streamlets. This is an important features to enable scalability for dynamically controlled streams with a single key.

It does not impact an existing code, but introduce a new API, thus allowing to control whether a streamlet reads from all partitions or using partitioning strategy based on scaling of the streamlet

No test are necessary. It basically creates a unique groupID for Kafka readers


core/build.sbt
```diff
@@ -124,7 +122,10 @@ lazy val akkastreamTestkit =
     .settings(
       (sourceDirectory in AvroConfig) := baseDirectory.value / "src/test/avro",
       (stringType in AvroConfig) := "String",
-      javacOptions += "-Xlint:unchecked"
+      javacOptions += "-Xlint:unchecked",
+      javaOptions += "-XX:MaxMetaspaceSize=512M",
```
RayRoestenburg: Please put these changes in a separate PR, they are not related to reading from all partitions.

core/cloudflow-spark-tests/src/test/scala/cloudflow/spark/SparkEgressSpec.scala
```diff
@@ -81,8 +81,8 @@ class SparkEgressSpec extends SparkScalaTestSupport {
       val r2 = asCollection[String](session, "allNamesUpper")
 
       // assert
-      r1 must contain("name1")
-      r2 must contain("NAME1")
+      r1 must contain("name2")
```
RayRoestenburg: This also does not look like it is related to read all partitions. Why was this change necessary? Did the test fail?

core/cloudflow-streamlets/src/main/scala/cloudflow/streamlets/avro/AvroInlet.scala
```diff
@@ -35,3 +36,18 @@ object AvroInlet {
   def create[T <: SpecificRecordBase](name: String, clazz: Class[T]): AvroInlet[T] =
     AvroInlet[T](name)(ClassTag.apply(clazz))
 }
+
+// Additional AvroInlet reading from all partitions of the topic. This is done by ensuring unique
+// groupID for Kafka listener
+case class AvroInletAllPartitions[T <: SpecificRecordBase: ClassTag](name: String) extends CodecInlet[T] {
```
RayRoestenburg: Instead it would be more consistent to add a `withUniqueGroupId` to `CodecInlet`. We already have a 'builder' API for `CodecOutlet`, which allows for setting a partitioner with `withPartitioner`. So doing it that way would be more consistent. Also, that way, this class is not necessary anymore.

RayRoestenburg:
Also, please also look at `SparkStreamletContext` and `FlinkStreamletContext`, if you want to be able to read as a unique group there too.

blublinsky:
I do not think we need to unique groupID in Flink and Spark. Both allow to bring all of the messages from all partitions together. This is something that is only required for Akka streams

core/build.sbt
blublinsky: Did this only to be able to run tests locally, removing

core/cloudflow-spark-tests/src/test/scala/cloudflow/spark/SparkEgressSpec.scala
blublinsky: Done

core/cloudflow-streamlets/src/main/scala/cloudflow/streamlets/avro/AvroInlet.scala
blublinsky: Redesign based on Ray's suggestion of implementing a builder

core/cloudflow-streamlets/src/main/scala/cloudflow/streamlets/StreamletPort.scala
```diff
@@ -61,6 +61,18 @@ trait CodecInlet[T] extends Inlet {
    * Describes the schema used to deserialize the data.
    */
   def schemaAsString: String
+
+  /**
+   * Returns if this inlet uses a unique group id.
+   */
```
agemooij: Can we please use the same, more informative comment as used for the Java version below?

core/cloudflow-akka/src/main/scala/cloudflow/akkastream/AkkaStreamletContextImpl.scala
```diff
@@ -73,7 +74,11 @@ final class AkkaStreamletContextImpl(
   }
 
   private val bootstrapServers = system.settings.config.getString("cloudflow.kafka.bootstrap-servers")
-  private def groupId[T](savepointPath: SavepointPath, streamletRef: String, inlet: CodecInlet[T]) = s"${savepointPath.appId}.${streamletRef}.${inlet.name}"
+  private def groupId[T](savepointPath: SavepointPath, streamletRef: String, inlet: CodecInlet[T]) = {
+    val base = s"${savepointPath.appId}.${streamletRef}.${inlet.name}"
+    if (inlet.hasUniqueGroupId) base + randomUUID.toString
+    else base
+  }
```
agemooij: Are there any length restrictions in group IDs we need to worry about here? A GUID is quite long...

RayRoestenburg: No:
https://github.com/apache/kafka/blob/72282ed1988e33000f91672a72c7356e26658241/core/src/main/scala/kafka/coordinator/group/GroupCoordinator.scala#L799

core/cloudflow-streamlets/src/main/scala/cloudflow/streamlets/StreamletPort.scala
RayRoestenburg: The one returns the boolean, the other sets it, but yes, I'll add it.

PR #71, opened 2019-12-09
Fine-tune action triggering on PR and `master` merge

maasg:
This PR makes that our test suite gets triggered for each external PR as well as for each `merge` action to master.


agemooij:
IT seems to have already been picked up since this PR itself got built!

PR #78, opened 2019-12-10
override getContext in SparkStreamletLogic

hepin1989:
follow up the previous one ,sorry for missing out an override


core/cloudflow-spark/src/main/scala/cloudflow/spark/SparkStreamlet.scala
```diff
@@ -230,6 +230,8 @@ trait SparkStreamlet extends Streamlet with Serializable {
  */
 abstract class SparkStreamletLogic(implicit val context: SparkStreamletContext) extends StreamletLogic[SparkStreamletContext] {
 
+  override def getContext(): SparkStreamletContext = super.getContext()
```
hepin1989: missed out

RayRoestenburg: Cool. Probably never found out cause of missing spark java api..

PR #92, opened 2019-12-20
Removes the deprecated environment variable `cluster type`.

olofwalker:
This variable has been deprecated for a long time but not removed.


maasg:
Looks good!

PR #104, opened 2020-01-14
Find the namespace where cloudflow is installed

debasishg:

* Adds a function `GetDefaultNamespace()` that fetches the namespace where cloudflow is installed. Instead of hardcoding `cloudflow` as the namespace, this can be used by the CLI. 
* Used this function to set the namespace for kafka topic creation CLI.
* Adds a label `namespace -> cloudflow-namespace` to the `protocol-version` configmap in Operator

* The namespace where cloudflow is installed was hardcoded to `cloudflow`. This needs to be changed.

* The user can now install cloudflow in any namespace and the CLI should be able to use it.

* Manually installing cloudflow on GKE with the specific operator version, we get the following which shows the label attached to the configmap:

```bash
$ kubectl describe cm cloudflow-protocol-version -n cloudflow
Name:         cloudflow-protocol-version
Namespace:    cloudflow
Labels:       namespace=cloudflow-namespace
Annotations:  <none>
```


kubectl-cloudflow/kafka/kafkatopic.go
```diff
@@ -62,7 +63,7 @@ func (in *kafakTopicList) DeepCopyObject() runtime.Object {
 }
 
 // GetAllKafkaTopics returns a list of all Kafka topics managed by Piplines for a specific application
-func GetAllKafkaTopics(applicationId string) []KafkaTopic {
+func GetAllKafkaTopics(applicationID string) []KafkaTopic {
```
olofwalker: This fixes a warning right?

kubectl-cloudflow/version/version.go
```diff
@@ -30,21 +30,47 @@ const ProtocolVersionKey = "protocol-version"
...
+// GetProtocolVersionConfigMap Get the protocol version config map set by the operator
+func GetProtocolVersionConfigMap() (*corev1.ConfigMap, error) {
 	k8sClient, k8sErr := k8s.GetClient()
 	if k8sErr != nil {
 		util.LogAndExit("Failed to create new kubernetes client, %s", k8sErr.Error())
 	}
 
 	var cm *corev1.ConfigMap
-	if configMaps, err := k8sClient.CoreV1().ConfigMaps("").List(metav1.ListOptions{}); err == nil {
+	configMaps, err := k8sClient.CoreV1().ConfigMaps("").List(metav1.ListOptions{})
```
olofwalker: We could add a label to the configmap when creating it in the Cloudflow operator and then use the ListOption to select based on label.

This shows how to select on label:
https://stackoverflow.com/questions/51106923/labelselector-for-secrets-list-in-k8s

kubectl-cloudflow/version/version.go
```diff
@@ -30,21 +30,47 @@ const ProtocolVersionKey = "protocol-version"
...
 				cm = &configMaps.Items[i]
 			}
 		}
-	} else {
+	} 
+	return cm, err
+}
+
+// GetDefaultNamespace Get the default namespace set in the protocol version config map
+func GetDefaultNamespace() (string, error) {
```
olofwalker: Maybe `GetCloudflowNamespace`? In the future, this could return a list to select from until you specified a default.

kubectl-cloudflow/kafka/kafkatopic.go
debasishg: yes! a warning in VSCode ..

kubectl-cloudflow/version/version.go
debasishg: cool .. will do .. Just curious - this implementation will work but the label will make the search efficient - right ?

kubectl-cloudflow/version/version.go
debasishg: will change ..

kubectl-cloudflow/version/version.go
olofwalker: Yes, I was thinking that the cluster may have, well many configmaps so ...

skonto: It can be fast enough with `--watch-cache` big enough from what I read https://github.com/kubernetes/community/pull/2203

kubectl-cloudflow/version/version.go
skonto: An easier approach could be to let the user pass the namespace anyway at the cli even if we currently support one Cloudflow installation.

debasishg: The problem with passing the namespace is that that affects a number of programs where u need to handle an additional argument to the cli - this approach has less impact

agemooij: I'd prefer the label-scanning version proposed by @olofwalker . There is already another CLI change proposal in our backlog that will use the native `kubectl` namespace handling, including the `-n/--namespace` flag for selecting the namespace to deploy an app to. I don't think another namespace flag would make a lot of sense at this stage.

I do agree that in the far distant future, when K8s multi-tenancy support is further along, we might be able to support multiple cloudflow instances and only then does it get an issue of possibly finding two instances. It's a bit too early to worry about that right now though IMHO.

debasishg:
@olofwalker Have incorporated the changes that u suggested in cli and operator. Pls take a look ..

kubectl-cloudflow/version/version.go
olofwalker: @agemooij @debasishg Should the CLI at this point even work if it finds more than one installation of Cloudflow then? There will be two operators there fighting for the CR being created. I would vote for an error being printed in the CLI in that case.

agemooij: @olofwalker  Agreed

kubectl-cloudflow/version/version.go
```diff
@@ -30,21 +33,58 @@ const ProtocolVersionKey = "protocol-version"
...
+		return nil, fmt.Errorf("Invalid protocol version config %s detected", cm.GetName())
+	}
+	return cm, err
+}
+
+// GetCloudflowNamespace Get the default namespace set in the protocol version config map
+func GetCloudflowNamespace() (string, error) {
+	cm, err := GetProtocolVersionConfigMap()
+	if err != nil {
+		util.LogAndExit("Could not get defaiult namespace. Kubernetes API returned an error: %s", err)
```
agemooij:
```suggestion
		util.LogAndExit("Could not get default namespace. Kubernetes API returned an error: %s", err)
```

kubectl-cloudflow/version/version.go
```diff
@@ -30,21 +33,58 @@ const ProtocolVersionKey = "protocol-version"
...
+	}
+
+	if cm == nil {
+		util.LogAndExit("Cannot find the '%s' configmap and the default namespace, please make sure that the Cloudflow operator is installed", ProtocolVersionConfigMapName)
```
agemooij:
```suggestion
		util.LogAndExit("Cannot find the '%s' ConfigMap and/or the Cloudflow namespace. Please make sure that the Cloudflow operator is installed", ProtocolVersionConfigMapName)
```

core/cloudflow-operator/src/main/scala/cloudflow/operator/CloudflowLabels.scala
```diff
@@ -55,4 +55,8 @@ object CloudflowLabels {
 
   // Managed by
   val ManagedByCloudflow = "cloudflow"
+
+  // Cloudflow installation namespace
+  val Namespace = "namespace"
+  val CloudflowNamespace = "cloudflow-namespace"
```
agemooij: These label key/values don't make much sense to me. what are they actually saying?

I think the purpose of this one specific label is to make the created ConfigMap more easily findable using a label query, e.g. the label key name should be based on the purpose of the `ConfigMap` and the value is essentially just the name of the config map itself as well.

So I suggest we remove the above two values and make both the key and value of the label to be `ProtocolVersionConfigMapName`

WDYT?

debasishg: yeah .. makes sense .. +1 .. will change

kubectl-cloudflow/kafka/kafkatopic.go
```diff
@@ -79,11 +80,16 @@ func GetAllKafkaTopics(applicationId string) []KafkaTopic {
 		util.LogAndExit("Failed to connect to cluster, %s", err.Error())
 	}
 
+	namespace, err := version.GetCloudflowNamespace()
+	if err != nil {
+		util.LogAndExit("Failed to get default namespace, %s", err.Error())
```
agemooij:
```suggestion
		util.LogAndExit("Failed to find the Cloudflow namespace, %s", err.Error())
```

kubectl-cloudflow/version/version.go
```diff
@@ -30,21 +33,58 @@ const ProtocolVersionKey = "protocol-version"
...
+			return cm, nil
 		}
-	} else {
+		return nil, fmt.Errorf("Invalid protocol version config %s detected", cm.GetName())
+	}
+	return cm, err
+}
+
+// GetCloudflowNamespace Get the default namespace set in the protocol version config map
+func GetCloudflowNamespace() (string, error) {
```
agemooij: I'm not totally up to scratch on Golang naming conventions but I suggest we rename this to `FindCloudflowNamespace ` since it is not guaranteed that there actually is one. to me it would better describe the purpose and the context at the call site.

kubectl-cloudflow/version/version.go
```diff
@@ -30,21 +33,58 @@ const ProtocolVersionKey = "protocol-version"
...
-	if configMaps, err := k8sClient.CoreV1().ConfigMaps("").List(metav1.ListOptions{}); err == nil {
-		for i, v := range configMaps.Items {
-			if v.Name == ProtocolVersionConfigMapName {
-				cm = &configMaps.Items[i]
-			}
+	configMaps, err := k8sClient.CoreV1().ConfigMaps("").List(
+		metav1.ListOptions{LabelSelector: labels.Set(labelSelector.MatchLabels).String()})
+	if err == nil {
+		if len(configMaps.Items) > 1 {
+			return nil, errors.New("Multiple cloudflow operators detected in namespace .. exiting")
```
agemooij:
```suggestion
			return nil, errors.New("Multiple Cloudflow operators detected in the cluster. This is not supported. Exiting...")
```

kubectl-cloudflow/version/version.go
```diff
@@ -30,21 +33,58 @@ const ProtocolVersionKey = "protocol-version"
...
+		}
+		cm = &configMaps.Items[0]
+		if cm.GetName() == ProtocolVersionConfigMapName {
```
agemooij: Now that we check for label and we check that there is only one. why do e need to check for the name as well? I do not think this is very useful and it introduces a corner case and an extra error message that we just don't need.

I suggest we remove this.

kubectl-cloudflow/version/version.go
debasishg: @agemooij 2 observations .. This `return` is from the control flow where we have detected one instance of the operator, but somehow the name doesn't match with the protocol-version configmap name. Possibly this case will not arise unless we inadvertently change the name in one place (operator) and not the other (cli). Secondly `go lint` complains on the formatting of the message as it has capitalization .. 
Screen Shot 2020-01-15 at 8 30 02 PM

kubectl-cloudflow/version/version.go
debasishg: This is related to the earlier comment that u made. I replied there as well. This is indeed a corner case that can be introduced if we change the name in cli and forget to change in the operator code (or vice versa) .. Should I remove this check ?

kubectl-cloudflow/version/version.go
debasishg: ok .. will change ..

kubectl-cloudflow/version/version.go
agemooij: This return AFAIk is for the case when we have found multiple ConfigMaps in the cluster with the exact same label key and value, e.g. somehow the cluster has multiple instances of the operator running in different namespaces. I think that is the situation we are then reporting an error about since we do not support such clusters.

Cloudflow is a product name so it should be capitalized. Perhaps it is complaining about the trailing dots? Whenever some message to the console says "...ing" I read it as something that is continuing after I have read it, hence the elipsis at the end. I can live without that one if it causes such issues.

kubectl-cloudflow/version/version.go
agemooij: I think now that we find the configMap by label the check on name has no value anymore.

See above for my reply to your other comment, e.g. it is still possible for the CLI and the operator to have a mismatch in the hardcoded label key/value used to identify the configMap by label. I that is the case we will not find _any_ ConfigMap from the CLI, which is a corner case we should indeed cover and report on.

kubectl-cloudflow/version/version.go
debasishg: yeah .. I think we will never get into this `return` statement since all matches in the label selector and comparisons are done using `ProtocolVersionConfigMapName`. Will remove that `return` ..

kubectl-cloudflow/version/version.go
debasishg: see my earlier reply - will remove that `return` statement ..

agemooij: @debasishg I think we can remove this conditional and just always return here instead of only when the name of the config map is the one we expect. We don't care about the name since we have already found it through its label.

kubectl-cloudflow/version/version.go
```diff
@@ -30,40 +32,70 @@ const ProtocolVersionKey = "protocol-version"
...
+	}
+	return cm.GetObjectMeta().GetNamespace(), err
+}
+
+// FailOnProtocolVersionMismatch fails and exits if the protocol version of kubectl-cloudflow does not match with the cloudflow operator protocol version.
+func FailOnProtocolVersionMismatch() {
+	_, k8sErr := k8s.GetClient()
+	if k8sErr != nil {
+		util.LogAndExit("Failed to create new kubernetes client, %s", k8sErr.Error())
+	}
```
agemooij: This is already done at the top of the `GetProtocolVersionConfigMap ` function so we are doing it twice.

kubectl-cloudflow/version/version.go
```diff
@@ -30,40 +32,70 @@ const ProtocolVersionKey = "protocol-version"
...
+
+	if cm == nil {
+		util.LogAndExit("Cannot find the '%s' ConfigMap and/or the Cloudflow namespace. Please make sure that the Cloudflow operator is installed", ProtocolVersionConfigMapName)
+	}
+	return cm.GetObjectMeta().GetNamespace(), err
+}
+
+// FailOnProtocolVersionMismatch fails and exits if the protocol version of kubectl-cloudflow does not match with the cloudflow operator protocol version.
+func FailOnProtocolVersionMismatch() {
```
agemooij: It looks like we are calling `GetProtocolVersionConfigMap` twice, once to get the namespace and once to verify the protocol version. Would it be possible to refactor this to only scan for the config map once?

debasishg: @agemooij `GetProtocolVersionConfigMap` is called twice in _different_ flow of control. If we want to avoid it we need to get the value and have it cached. Are you suggesting that ? I think that would be a micro-optimization as the call itself is not very expensive.

kubectl-cloudflow/version/version.go
debasishg: will fix

kubectl-cloudflow/version/version.go
debasishg: done ..

kubectl-cloudflow/version/version.go
olofwalker: I do not think Age is suggesting that we should cache the value, I think it's fine.

PR #109, opened 2020-01-21
Resurrect internal release - use bintray for internal release

debasishg:

* Resurrect `sbt internalRelease`. It was broken since we started releasing on to Sonatype. Internal release releases artifacts to bintray

* Internal release is needed to test examples based on released artifacts. Every time it's not feasible to release on to sonatype - hence this releases stuff to lightbend's public bintray repository

* Yes, adds an option to do an internal release of cloudflow libraries

* The patch has been tested manually by creating internal release

### DIFF
--- a/core/build.sbt
+++ b/core/build.sbt
@@ -443,7 +443,24 @@ def cloudflowModule(moduleID: String): Project = {
     .enablePlugins(AutomateHeaderPlugin)
 }
 
-lazy val commonSettings = Seq(
+// These settings are made active only when we use bintray for internal release
+// It is important that when we do final releases we need to invoke sbt as 
+// `sbt -Dsbt.sbtbintray=false`
+lazy val bintraySettings =
+  if (BintrayPlugin.isEnabledViaProp) {
+    Seq(bintrayOrganization := Some("lightbend"),
+      bintrayRepository := "cloudflow",
+      bintrayOmitLicense := true,
+      publishMavenStyle := false,
+      resolvers ++= Seq(
+        "Akka Snapshots" at "https://repo.akka.io/snapshots/",
+        "com-mvn" at "https://repo.lightbend.com/cloudflow" , Resolver.url("com-ivy",
+        url("https://repo.lightbend.com/cloudflow"))(Resolver.ivyStylePatterns)
+      )
+    )
+  } else Seq.empty
+
+lazy val commonSettings = bintraySettings ++ Seq(
   organization := "com.lightbend.cloudflow",
   headerLicense := Some(HeaderLicense.ALv2("(C) 2016-2020", "Lightbend Inc. <https://www.lightbend.com>")),
   scalaVersion := Version.Scala,
@@ -484,6 +501,8 @@ lazy val commonSettings = Seq(
   scalacOptions in (Compile, console) := (scalacOptions in (Global)).value.filter(_ == "-Ywarn-unused-import"),
   scalacOptions in (Test, console) := (scalacOptions in (Compile, console)).value,
 )
+
+releaseIgnoreUntrackedFiles := true
 
 lazy val formattingSettings = Seq(
   scalariformPreferences := scalariformPreferences.value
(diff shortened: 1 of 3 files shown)


core/build.sbt
```diff
@@ -443,7 +443,24 @@ def cloudflowModule(moduleID: String): Project = {
     .enablePlugins(AutomateHeaderPlugin)
 }
 
-lazy val commonSettings = Seq(
+// These settings are made active only when we use bintray for internal release
+// It is important that when we do final releases we need to invoke sbt as 
+// `sbt -Dsbt.sbtbintray=false`
```
skonto: We could have this set to false by default. Testing should be optionally switched on, its like a dry-run flag.

core/sbt-cloudflow/src/main/scala/cloudflow/sbt/CommonSettingsAndTasksPlugin.scala
```diff
@@ -42,11 +42,15 @@ object CommonSettingsAndTasksPlugin extends AutoPlugin {
   // common definitions
   final val CloudflowLocalConfigFile = ".lightbend/cloudflow/pipectl.json"
   final val CloudflowDockerBaseImage = "lightbend/cloudflow-base:1.3.0-M1-spark-2.4.4-flink-1.9.1-scala-2.12"
+  // used for internal release
+  final val CloudflowBintrayReleasesRepoUrl = "https://lightbend.bintray.com/cloudflow"
```
skonto: something like `CloudflowBintrayTestReleasesRepoUrl` could be more precise:

skonto: Is this going to be shipped with the plugin? I think this should be only enabled for internal use somehow.

core/build.sbt
debasishg: @skonto The logic is we will use `internalRelease` much more frequently than the actual one. Hence the plugin is on by default and we set it to false _only during an actual release_. See the other PR on `cloudflow-release`

core/sbt-cloudflow/src/main/scala/cloudflow/sbt/CommonSettingsAndTasksPlugin.scala
debasishg: @skonto Strictly speaking it's not for test, it's for an internal release which, once done, all team members can use - hence the name.

Regarding shipping, why not ? The repository where we publish is a public one - hence people can use it from outside lightbend as well. And besides, how can we keep it only for internal use ?

core/build.sbt
```diff
@@ -443,7 +443,24 @@ def cloudflowModule(moduleID: String): Project = {
...
+lazy val bintraySettings =
+  if (BintrayPlugin.isEnabledViaProp) {
+    Seq(bintrayOrganization := Some("lightbend"),
+      bintrayRepository := altBintrayRepository.value.getOrElse("cloudflow"),
+      bintrayOmitLicense := true,
+      publishMavenStyle := false,
+      resolvers += "Akka Snapshots" at "https://repo.akka.io/snapshots/",
```
olofwalker: Could combine this line and the line below.

debasishg: @olofwalker On a second look, I am thinking if we really need this akka snapshot repo added to the resolver ? Do we use snapshots ? /cc  @RayRoestenburg

core/build.sbt
```diff
@@ -485,6 +502,11 @@ lazy val commonSettings = Seq(
   scalacOptions in (Test, console) := (scalacOptions in (Compile, console)).value,
 )
 
+val altBintrayRepository = settingKey[Option[String]]("Overrides the default bintray repository if defined")
+ThisBuild / altBintrayRepository := None
```
skyluc: Do we still use the `altBintrayRepository` setting ? It was used to push nightly builds to a different bintray repo, but we don't do this anymore.
I don't see a reference to it in the release scripts.

debasishg: yeah @skyluc  I think we can remove it and set the repo name straight to `cloudflow`

core/build.sbt
debasishg: done ..

core/build.sbt
debasishg: removed `altBintrayRepository` ..

PR #113, opened 2020-01-21
Config file support, env var support, application and streamlet level support.

RayRoestenburg:

- [x] Support passing through config files, which are automatically merged in order specified (last takes most precedence)
- [x] Support application level configuration (pass through config for all streamlets)
- [x] Support streamlet specific configuration (using the application-level section, it's content is copied to the root of the conf)
- [x] Support env var replacements (env vars in the env where kubectl cloudflow is running) 
- [x] Allow config args to pass through subsystem config

Improvements to configuration system. This prepares for other planned changes, to be able eto easily configure any aspect of a cloudflow application.

Yes, users will be able to pass config files in HOCON format to kubectl-cloudflow. configuration parameters used in the file are validated, just like the arguments are validated. It is possible to pass through config that should not be validated with configuration parameters.

Tested on a Cloudflow cluster, as wel as unit testing.
- [x] Add more tests
- [ ] Docs
- [ ] Docs: Describe some limitations of config file support  (looks like only a = ${b.c.d} does not work, and does not read beyond root braces, which I think is maybe correct: https://github.com/lightbend/config/blob/master/HOCON.md#omit-root-braces)

### DIFF
--- a/kubectl-cloudflow/go.mod
+++ b/kubectl-cloudflow/go.mod
@@ -14,6 +14,7 @@ require (
 	github.com/docker/docker v0.0.0-20190327063223-da823cf3a5a3
 	github.com/docker/go-connections v0.4.0 // indirect
 	github.com/docker/go-units v0.3.3 // indirect
+	github.com/go-akka/configuration v0.0.0-20200115015912-550403a6bd87 // indirect
 	github.com/gogo/protobuf v1.2.1 // indirect
 	github.com/google/gofuzz v0.0.0-20170612174753-24818f796faf // indirect
 	github.com/googleapis/gnostic v0.2.0 // indirect
@@ -33,6 +34,7 @@ require (
 	github.com/opencontainers/go-digest v1.0.0-rc1 // indirect
 	github.com/opencontainers/image-spec v1.0.1 // indirect
 	github.com/pkg/errors v0.8.1 // indirect
+	github.com/rayroestenburg/configuration v0.0.0-20200115015912-550403a6bd87
 	github.com/sirupsen/logrus v1.4.0 // indirect
 	github.com/spf13/cobra v0.0.3
 	github.com/spf13/pflag v1.0.3 // indirect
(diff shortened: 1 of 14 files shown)


kubectl-cloudflow/go.mod
```diff
@@ -33,6 +34,7 @@ require (
 	github.com/opencontainers/go-digest v1.0.0-rc1 // indirect
 	github.com/opencontainers/image-spec v1.0.1 // indirect
 	github.com/pkg/errors v0.8.1 // indirect
+	github.com/rayroestenburg/configuration v0.0.0-20200115015912-550403a6bd87
```
skonto: Do we allow our own repo deps here?

kubectl-cloudflow/go.sum
```diff
@@ -151,6 +153,8 @@ github.com/prometheus/common v0.2.0/go.mod h1:TNfzLD0ON7rHzMJeJkieUDPYmFC7Snx/y8
 github.com/prometheus/procfs v0.0.0-20180725123919-05ee40e3a273/go.mod h1:c3At6R/oaqEKCNdg8wHV1ftS6bRYblBhIjjI8uT2IGk=
 github.com/prometheus/procfs v0.0.0-20181005140218-185b4288413d/go.mod h1:c3At6R/oaqEKCNdg8wHV1ftS6bRYblBhIjjI8uT2IGk=
 github.com/prometheus/procfs v0.0.0-20190117184657-bf6a532e95b1/go.mod h1:c3At6R/oaqEKCNdg8wHV1ftS6bRYblBhIjjI8uT2IGk=
+github.com/rayroestenburg/configuration v0.0.0-20200115015912-550403a6bd87 h1:JTGDqHa23LTUjEsJQorToCYedKxI6SM9SkhooYyKlTw=
```
skonto: pls rm your repo references.

kubectl-cloudflow/go.mod
RayRoestenburg: I wanted to make sure we can add fixes directly.

kubectl-cloudflow/go.sum
RayRoestenburg: This is auto-generated

skonto: I dont have it on my setup and I can use that lib as is.

agemooij:
Discussed offline. Due to some design discussions we have updated the spec for the new config system and that makes this Pr outdated. We will submit a new one based on the latest spec.

PR #117, opened 2020-01-23
Upgrade Spark and Flink Operators and Strimzi

yuchaoran2011:
* Upgrade Spark Operator Helm chart to the one hosted at [helm/charts](https://github.com/helm/charts/tree/master/incubator/sparkoperator).
* Upgraded Strimzi to 0.1.6.2.
* Upgraded Flink to 1.9.2 and Flink operator to 0.4.0.
* Fixed some bugs in the uninstall script.


installer/common/detect.sh
```diff
@@ -26,7 +26,7 @@ detect_strimzi_crds() {
 
 # Returns the namespace the sparkoperator is installed in or an empty string
 detect_sparkoperator() {
-    kubectl get deployment -l app.kubernetes.io/name=fdp-sparkoperator --all-namespaces --ignore-not-found=true --no-headers | awk -v i=1 -v j=1 'FNR == i {print $j}'
+    kubectl get deployment -l app.kubernetes.io/name=sparkoperator --all-namespaces --ignore-not-found=true --no-headers | awk -v i=1 -v j=1 'FNR == i {print $j}'
```
skonto: I think it would be good to have `cloudflow-sparkoperator` if it does not create any issues.

installer/install.sh
```diff
@@ -60,6 +60,7 @@ init_helm
 # helm repos.
 helm repo add lightbend-helm-charts https://repo.lightbend.com/helm-charts/ > /dev/null 2>&1
 helm repo add strimzi http://strimzi.io/charts/ > /dev/null 2>&1
+helm repo add incubator http://storage.googleapis.com/kubernetes-charts-incubator > /dev/null 2>&1
```
skonto: Should I close https://github.com/lightbend/cloudflow/pull/8? I would rather aim for updating, moving https://github.com/lightbend/fdp-sparkoperator to https://hub.helm.sh/, as central helm charts are being deprecated (https://github.com/helm/charts#deprecation-timeline).

installer/uninstall-cloudflow.sh
```diff
@@ -67,15 +67,47 @@ if [[ $REPLY =~ ^[Yy]$ ]]; then
...
+    kubectl delete clusterrolebinding cloudflow-nfs-fdp-nfs \
+      cloudflow-flink-flink-operator \
+      cloudflow-sparkoperator-crb \
+      cloudflow-operator-bindings \
+      strimzi-cluster-operator \
+      strimzi-cluster-operator-kafka-broker-delegation \
+      --ignore-not-found=true
+
+    echo "Removing StorageClasses..."
+    kubectl delete sc nfs-client \
```
skonto: This should be removed only if nsf is used, what if efs is used?

yuchaoran2011: If NFS is not used (i.e. storage class `nfs-client` is not present), this command won't do anything. It won't stop the script with an error.

installer/install.sh
yuchaoran2011: Right. I do remember that the current Helm repo is being deprecated and we need to add the chart to the hub. That can be a separate task that is implemented in parallel to this PR.

installer/common/detect.sh
yuchaoran2011: At the moment, this [label](https://github.com/helm/charts/blob/master/incubator/sparkoperator/templates/spark-operator-deployment.yaml#L13) is linked to the [name of the chart](https://github.com/helm/charts/blob/master/incubator/sparkoperator/templates/_helpers.tpl#L6). So not configurable yet. But we can make it so.

installer/install.sh
skonto: The question is whether we want to do this now as other people have done so we dont do double work.

yuchaoran2011: Agreed. We'll need @agemooij  to green light it ;)

installer/uninstall-cloudflow.sh
skonto: In general it is a good practice not to make unnecessary calls. Also we need to cleanup all storage classes.

skonto:
Could you also update the description of the task since this targets the broader community?

installer/uninstall-cloudflow.sh
yuchaoran2011: This is necessary because when NFS is used, uninstalling cloudflow without deleting the storage class will create an error next time cloudflow is installed.

installer/common/shared.sh
```diff
@@ -135,9 +136,6 @@ elif [ "${KAFKA}" = "ExternalStrimzi" ]; then
         print_error_message "Strimzi topic operator configuration is invalid."
         print_error_message "You must define environment variable 'KAFKA_STRIMZI_TOPIC_OPERATOR_NAMESPACE' and set it to a namespace that contains a running strimzi-topic-operator pod."
         print_error_message ""
-        ##TODO FIX_PIPELINES_REF
-        print_error_message "More information can be found here: https://developer.lightbend.com/docs/pipelines/current/index.html#_existing_strimzi_kafka"
-        print_error_message ""
```
yuchaoran2011: @lightbend/cloudflow-team I'm removing these URLs for now because they are no longer reachable. Let me know if you think there are alternative good links that I can put here instead.

installer/common/shared.sh
```diff
@@ -63,8 +63,9 @@ export zookeeperHosts=""
 
 # Spark Operator
 export sparkOperatorReleaseName="cloudflow-sparkoperator"
-export sparkOperatorChartVersion="0.4.0"
-export sparkOperatorImageVersion="1.3.0-M1-OpenJDK-2.4.4-0.8.2-cloudflow-2.12"
+export sparkOperatorChartVersion="0.6.2"
+export sparkOperatorImageName="lightbend/sparkoperator"
+export sparkOperatorImageVersion="1.3.0-M1-OpenJDK-2.4.4-1.0.1-cloudflow-2.12"
```
yuchaoran2011: @skonto Stavros, I think you built the current image tagged `1.3.0-M1-OpenJDK-2.4.4-0.8.2-cloudflow-2.12`. Could you also build a new image with tag `1.3.0-M1-OpenJDK-2.4.4-1.0.1-cloudflow-2.12` using the latest operator release? Or I can do it if you have a document that you can point me to for reference.

installer/install.sh
yuchaoran2011: @skonto @agemooij I think we should start the process of migrating the chart to https://hub.helm.sh/, because as Stavros said, the deprecation deadline is approaching. In my opinion, we should simply discard https://github.com/lightbend/fdp-sparkoperator. We just put https://github.com/helm/charts/tree/master/incubator/sparkoperator to Helm hub. If there's anything useful in the former, we take it and put it in the latter as we migrate. 

The same thing should be done for our flink-operator chart. Someone already [asked](https://github.com/lightbend/flink-operator/issues/2#issuecomment-577572034) to help.

skonto: @yuchaoran2011 if we put  https://github.com/helm/charts and then that repo is depracated how we are going to do modifications? I think we need to migrate to our own repo which means rename and keep what we have, as people did here: https://github.com/helm/charts/issues/18781

installer/common/shared.sh
skonto: @yuchaoran2011 sure here are the steps: https://github.com/lightbend/cloudflow/blob/master/external/spark/releaseSparkDist.sh#L65-L74, @debasishg since tags are related to a Cloudflow release should we somehow integrate this with the release scripts you have created (if not done yet)?

installer/install.sh
yuchaoran2011: What I meant was copy https://github.com/helm/charts/tree/master/incubator/sparkoperator, host it ourselves at Lightbend and then put it to Helm hub.

installer/common/shared.sh
yuchaoran2011: Got it. I'll create and push the image and test it with this PR.

yuchaoran2011: I have created and pushed a new image with tag `1.3.0-M1-OpenJDK-2.4.4-1.0.1-cloudflow-2.12`. Now install (including dependent charts) and uninstall both work.

installer/install.sh
skonto: cool

yuchaoran2011:
I'll further upgrade the chart of Spark operator used by Cloudflow to [0.6.3](https://github.com/helm/charts/pull/20570/) that fixed a CRD validation error.

core/cloudflow-operator/src/test/scala/cloudflow/operator/TestDeploymentContext.scala
```diff
@@ -42,14 +42,14 @@ trait TestDeploymentContext {
       sparkRunnerSettings =
         SparkRunnerSettings(
           driverSettings = SparkPodSettings(
-            cores = Some("0.2"),
+            cores = Some(1),
```
olofwalker: Quite a jump!

yuchaoran2011: Yeah. The new spark operator CRD requires this number to be an integer that’s >= 1.

core/cloudflow-operator/src/main/scala/cloudflow/operator/runner/SparkRunner.scala
```diff
@@ -117,8 +118,12 @@ object SparkRunner extends Runner[CR] with PatchProvider[SpecPatch] {
       Map(Operator.StreamletNameLabel -> deployment.streamletName, Operator.AppIdLabel -> appId)
 
     import ctx.sparkRunnerSettings._
+    val cores = driverSettings.cores.map(_.amount.intValue).getOrElse(1)
     val driver = Driver(
-      cores = driverSettings.cores.map(_.amount.floatValue),
+      cores = if (cores >= 1)
+        Some(cores)
+      else
+        Some(1),
```
agemooij: I suggest you use the original code and add `orElse(1) to the end.

core/cloudflow-operator/src/main/scala/cloudflow/operator/runner/SparkRunner.scala
```diff
@@ -130,7 +135,7 @@ object SparkRunner extends Runner[CR] with PatchProvider[SpecPatch] {
       securityContext = securityContext
     )
     val executor = Executor(
-      cores = executorSettings.cores.map(_.amount.floatValue),
+      cores = executorSettings.cores.map(_.amount.intValue),
```
agemooij: Number of cores is definitely a float, not an int!

core/cloudflow-operator/src/main/scala/cloudflow/operator/runner/SparkRunner.scala
```diff
@@ -212,7 +217,7 @@ object SparkResource {
   )
 
   final case class Driver(
-      cores: Option[Float] = None,
+      cores: Option[Int] = Some(1),
```
agemooij: Are you sure this is not a float value, like `cpu` in K8s resource requests/limits?

core/cloudflow-operator/src/test/scala/cloudflow/operator/TestDeploymentContext.scala
agemooij: Ah, ok. Do you have a link to background info on this? why did they do this and what's the potential impact on our minimum resource usage?

examples/spark-sensors/project/build.properties
```diff
@@ -1 +1 @@
-sbt.version = 1.2.8
+sbt.version=1.3.7
```
agemooij: So far we have kept everything at sbt 1.2.x levels due to some bad experiences with early versions of sbt 1.3.x. Those might have been solved in the mean time but perhaps we should tackle the migration to sbt 1.3.x as a separate PR and refresh our memories on what went wrong the first time we tried that (can't currently remember I'm afraid).

installer/common/shared.sh
agemooij: We have a backlog item to port those old documentation pages to the new docs. We are also planning some major installer updates so for now I think this is acceptable

installer/create-cluster-gke.sh
```diff
@@ -71,7 +71,7 @@ gcloud beta container node-pools create kafka-pool-0 \
 ## Wait for clusters to come up
 echo "Waiting for cluster to become stable before continuing with the installation....."
 gcloud compute instance-groups managed list | grep gke-$CLUSTER_NAME | awk '/'$my_name'/ {print $1}' | while read -r line ; do
-  gcloud compute instance-groups managed wait-until-stable $line
+  gcloud compute instance-groups managed wait-until --stable $line
```
agemooij: Reviewers note: the old command was deprecated

core/cloudflow-operator/src/test/scala/cloudflow/operator/TestDeploymentContext.scala
yuchaoran2011: I can't seem to find the rationale behind this change. FYI, this is where the restriction is imposed: https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/blob/master/manifest/crds/sparkoperator.k8s.io_sparkapplications.yaml#L345

core/cloudflow-operator/src/main/scala/cloudflow/operator/runner/SparkRunner.scala
yuchaoran2011: Yes it has to be an Int, as indicated in the URL in my comment above. Using a float will make our Spark app (e.g. spark-sensors) fail with a CR validation error.

examples/spark-sensors/project/build.properties
yuchaoran2011: Makes sense. I'll revert to 1.2.8.

Issue #74, opened 2019-12-09
Use scalafmt in all examples


RayRoestenburg:
**Describe the bug**
Not really a bug, but would be good for consistency, to remove use of scalariformat, and use scalafmt instead, everywhere.

agemooij:
Here's a suggested `.scalafmt.conf` file that might fit our current scalariform setup without too many huge changes:

```
version = "2.0.1"

style = defaultWithAlign

align                               = more
align.openParenCallSite             = true
align.openParenDefnSite             = true
rewrite.rules                       = [AvoidInfix,RedundantBraces,RedundantParens,SortImports] 
danglingParentheses                 = true
docstrings                          = JavaDoc
maxColumn                           = 140
unindentTopLevelOperators           = true
rewrite.neverInfix.excludeFilters   = [
  and
  min
  max
  until
  to
  by
  eq
  ne
  "should.*"
  "contain.*"
  "must.*"
  in
  ignore
  be
  taggedAs
  thrownBy
  synchronized
  have
  when
  size
  only
  noneOf
  oneElementOf
  noElementsOf
  atLeastOneElementOf
  atMostOneElementOf
  allElementsOf
  inOrderElementsOf
  theSameElementsAs
]
```

agemooij:
If we can implement the conversion with as little forced code changes as possible that would be preferable of course, just to avoid a huge commit that screws with all other open PRs.

RayRoestenburg:
Sounds good to me!

osleonard:
Taking a peek at this

olofwalker:
ScalaFmt has now been implemented in core and all examples.

PR #124, opened 2020-02-06
This address #74

osleonard:
- removed references to scalariform & replaced with scalafmt


olofwalker:
@osleonard Thank you very much for this, a much needed PR. Can you add a fmt. file for the `core` project also?

olofwalker:
@osleonard We decided to go ahead and merge this `as-is` could you consider opening a new project, adding the same scalafmt to `core`? Thank you!!

osleonard:
will make the change for core

PR #125, opened 2020-02-07
Cloudflow Akka Plugin

debasishg:

* Implements cloudflow Akka Plugin
* Does some refactoring for better code reuse

* This is part of roadmap for multi-image cloudflow application. We need separate plugins for the runtimes to generate separate images.

* Yes

* Tested manually by integrating the plugin with example applications and checking that the generated image has the label from the streamlet descriptors

### DIFF
--- a/core/sbt-cloudflow/src/main/scala/cloudflow/sbt/CloudflowAkkaPlugin.scala
+++ b/core/sbt-cloudflow/src/main/scala/cloudflow/sbt/CloudflowAkkaPlugin.scala
@@ -1,0 +1,125 @@
+/*
+ * Copyright (C) 2016-2020 Lightbend Inc. <https://www.lightbend.com>
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package cloudflow.sbt
+
+import sbt._
+import sbt.Keys._
+import sbtdocker._
+import sbtdocker.DockerKeys._
+import com.typesafe.sbt.packager.Keys._
+import spray.json._
+import cloudflow.sbt.CloudflowKeys._
+import cloudflow.blueprint.StreamletDescriptorFormat._
+
+object CloudflowAkkaPlugin extends CloudflowBasePlugin {
+
+  override def projectSettings = Seq(
+    libraryDependencies ++= Vector(
+      "com.lightbend.cloudflow" %% "cloudflow-akka-util" % BuildInfo.version,
+      "com.lightbend.cloudflow" %% "cloudflow-akka" % BuildInfo.version,
+      "com.lightbend.cloudflow" %% "cloudflow-runner" % BuildInfo.version,
+      "com.lightbend.cloudflow" %% "cloudflow-akka-testkit" % BuildInfo.version % "test"
+    ),
+
+    cloudflowAkkaDockerImageName := Def.task {
+      Some(DockerImageName((ThisProject / name).value.toLowerCase, (ThisProject / cloudflowBuildNumber).value.buildNumber))
+    }.value,
+
+    streamletDescriptorsInProject := Def.taskDyn {
+      val detectedStreamlets = cloudflowStreamletDescriptors.value
+      buildStreamletDescriptors(detectedStreamlets)
+    }.value,
+
+    buildOptions in docker := BuildOptions(
+      cache = true,
+      removeIntermediateContainers = BuildOptions.Remove.OnSuccess,
+      pullBaseImage = BuildOptions.Pull.IfMissing
+    ),
+
+    cloudflowStageAppJars := Def.taskDyn {
+      Def.task {
+        val stagingDir = stage.value
+        val projectJars = (Runtime / internalDependencyAsJars).value.map(_.data)
+        val depJars = (Runtime / externalDependencyClasspath).value.map(_.data)
+
+        val appJarDir = new File(stagingDir, AppJarsDir)
+        val depJarDir = new File(stagingDir, DepJarsDir)
+        projectJars.foreach { jar ⇒
+          IO.copyFile(jar, new File(appJarDir, jar.getName))
+        }
+        depJars.foreach { jar ⇒
+          IO.copyFile(jar, new File(depJarDir, jar.getName))
+        }
+      }
+    }.value,
+
+    imageNames in docker := {
+      val registry = cloudflowDockerRegistry.value
+      val namespace = cloudflowDockerRepository.value
+
+      cloudflowAkkaDockerImageName.value
+        .map { imageName ⇒
+          ImageName(
+            registry = registry,
+            namespace = namespace,
+            repository = imageName.name,
+            tag = Some(imageName.tag)
+          )
+        }
+        .toSeq
+    },
+
+    dockerfile in docker := {
+      // this triggers side-effects, e.g. files being created in the staging area
+      cloudflowStageAppJars.value
+
+      val appDir: File = stage.value
+      val appJarsDir: File = new File(appDir, AppJarsDir)
+      val depJarsDir: File = new File(appDir, DepJarsDir)
+
+      // pack all streamlet-descriptors into a Json array
+      val streamletDescriptorsJson = streamletDescriptorsInProject.value.toJson
+
+      val streamletDescriptorsLabelValue = makeStreamletDescriptorsLabelValue(streamletDescriptorsJson)
+
+      new Dockerfile {
+        from("adoptopenjdk/openjdk8")
+        runRaw("groupadd -r cloudflow -g 185 && useradd -u 185 -r -g root -G cloudflow -m -d /home/cloudflow -s /sbin/nologin -c CloudflowUser cloudflow")
+        user(UserInImage)
+
+        copy(depJarsDir, OptAppDir, chown = userAsOwner(UserInImage))
+        copy(appJarsDir, OptAppDir, chown = userAsOwner(UserInImage))
+        label(StreamletDescriptorsLabelName, streamletDescriptorsLabelValue)
+      }
+    },
+
+    build := showResultOfBuild.dependsOn(
+      docker.dependsOn(
+        checkUncommittedChanges
+      )
+    ).value,
+
+    buildAndPublish := showResultOfBuildAndPublish.dependsOn(
+      dockerBuildAndPush.dependsOn(
+        checkUncommittedChanges,
+        verifyDockerRegistry
+      )
+    ).value,
+
+    fork in Compile := true
+  )
+}
--- a/core/sbt-cloudflow/src/main/scala/cloudflow/sbt/CloudflowBasePlugin.scala
+++ b/core/sbt-cloudflow/src/main/scala/cloudflow/sbt/CloudflowBasePlugin.scala
@@ -1,0 +1,128 @@
+/*
+ * Copyright (C) 2016-2020 Lightbend Inc. <https://www.lightbend.com>
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package cloudflow.sbt
+
+import java.nio.charset.StandardCharsets._
+import java.util.Base64
+import java.util.zip.Deflater
+import java.io.ByteArrayOutputStream
+
+import com.typesafe.config._
+
+import sbt._
+import sbt.Keys._
+import sbtdocker.DockerKeys._
+import com.typesafe.sbt.packager.archetypes._
+import spray.json._
+
+import cloudflow.sbt.CloudflowKeys._
+import cloudflow.blueprint.StreamletDescriptorFormat._
+import cloudflow.blueprint.StreamletDescriptor
+
+/**
+ * Base class for all Cloudflow runtime plugins for multi-image use case. Contains some
+ * methods which are reused across the runtime plugins, `CloudflowFlinkPlugin`,
+ * `CloudflowAkkaPlugin` and `CloudflowSparkPlugin`.
+ */
+abstract class CloudflowBasePlugin extends AutoPlugin {
+  final val AppHome = "${app_home}"
+  final val AppTargetDir: String = "/app"
+  final val appTargetSubdir: String ⇒ String = dir ⇒ s"$AppTargetDir/$dir"
+  final val AppJarsDir: String = "app-jars"
+  final val DepJarsDir: String = "dep-jars"
+  final val OptAppDir = "/opt/cloudflow/"
+  final val scalaVersion = "2.12"
+
+  // NOTE !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
+  // The UID and GID of the `jboss` user is used in different parts of Cloudflow
+  // If you change this, you have to make sure that all references to this value are changed
+  // - fsGroups on streamlet pods uses the GID to make volumes readable
+  val UserInImage = "185" // default non-root user in the spark image
+  val userAsOwner: String ⇒ String = usr ⇒ s"$usr:cloudflow"
+  val StreamletDescriptorsLabelName = "com.lightbend.cloudflow.streamlet-descriptors"
+
+  override def requires = CommonSettingsAndTasksPlugin && StreamletScannerPlugin &&
+    JavaAppPackaging && sbtdocker.DockerPlugin
+
+  private[sbt] val verifyDockerRegistry = Def.task {
+    cloudflowDockerRegistry.value.getOrElse(throw DockerRegistryNotSet)
+  }
+
+  private[sbt] val checkUncommittedChanges = Def.task {
+    val log = streams.value.log
+    if (cloudflowBuildNumber.value.hasUncommittedChanges) {
+      log.warn(s"You have uncommitted changes in ${thisProjectRef.value.project}. Please commit all changes before publishing to guarantee a repeatable and traceable build.")
+    }
+  }
+
+  private[sbt] def makeStreamletDescriptorsLabelValue(streamletDescriptorsJson: JsValue) = {
+    // create a root object with the array
+    val streamletDescriptorsJsonStr =
+      JsObject("streamlet-descriptors" -> streamletDescriptorsJson)
+        .compactPrint
+
+    val compressed = zlibCompression(streamletDescriptorsJsonStr.getBytes(UTF_8))
+    Base64.getEncoder.encodeToString(compressed)
+  }
+
+  private[sbt] val showResultOfBuild = Def.task {
+    val log = streams.value.log
+    val imagePushed = (imageNames in docker).value.head // assuming we only build a single image!
+
+    log.info(" ") // if you remove the space, the empty line will be auto-removed by SBT somehow...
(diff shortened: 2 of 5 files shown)


core/sbt-cloudflow/src/main/scala/cloudflow/sbt/CloudflowAkkaPlugin.scala
```diff
@@ -0,0 +1,129 @@
...
+      // pack all streamlet-descriptors into a Json array
+      val streamletDescriptorsJson =
+        streamletDescriptorsInProject
+          .value
+          .toJson
+
+      val streamletDescriptorsLabelValue = makeStreamletDescriptorsLabelValue(streamletDescriptorsJson)
+
+      new Dockerfile {
+        from("adoptopenjdk/openjdk8")
```
olofwalker: Should this be an individual decision for each plugin, or should we standardize this? My thinking here is that we should always provide the same JVM for all Streamlet types, Akka, Spark, Flink etc. Wonder if the creation of the docker file could be the same for all these plugins?

core/sbt-cloudflow/src/main/scala/cloudflow/sbt/CloudflowBasePlugin.scala
```diff
@@ -0,0 +1,126 @@
...
+import cloudflow.blueprint.StreamletDescriptorFormat._
+import cloudflow.blueprint.StreamletDescriptor
+
+abstract class CloudflowBasePlugin extends AutoPlugin {
+  val AppHome = "${app_home}"
+  val AppTargetDir: String = "/app"
+  val appTargetSubdir: String ⇒ String = dir ⇒ s"$AppTargetDir/$dir"
+  val AppJarsDir: String = "app-jars"
+  val DepJarsDir: String = "dep-jars"
+  val optAppDir = "/opt/cloudflow/"
```
olofwalker:
```suggestion
  val OptAppDir = "/opt/cloudflow/"
``` 
?

olofwalker: Are any of these vals final?

core/sbt-cloudflow/src/main/scala/cloudflow/sbt/CloudflowBasePlugin.scala
```diff
@@ -0,0 +1,126 @@
...
+  final val scalaVersion = "2.12"
+
+  // NOTE !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
+  // The UID and GID of the `jboss` user is used in different parts of Cloudflow
+  // If you change this, you have to make sure that all references to this value are changed
+  // - fsGroups on streamlet pods uses the GID to make volumes readable
+  val userInImage = "185" // default non-root user in the spark image
```
olofwalker: This is also a constant right?

core/sbt-cloudflow/src/main/scala/cloudflow/sbt/CloudflowAkkaPlugin.scala
agemooij: I agree that it would be good to make sure that all three images are based on the same base if that is at all possible.

debasishg: @olofwalker we are using this same jdk across all plugins. For Spark and Flink we have separate scripts that create the base images and all of them use this same jdk. For Akka, we don't need a separate step for creating a base image - hence we have it in the plugin.

core/sbt-cloudflow/src/main/scala/cloudflow/sbt/CloudflowBasePlugin.scala
debasishg: done ..

core/sbt-cloudflow/src/main/scala/cloudflow/sbt/CloudflowBasePlugin.scala
debasishg: yes .. changed the name

core/sbt-cloudflow/src/main/scala/cloudflow/sbt/CloudflowAkkaPlugin.scala
```diff
@@ -0,0 +1,129 @@
...
+        }
+        depJars.foreach { jar ⇒
+          IO.copyFile(jar, new File(depJarDir, jar.getName))
+        }
+      }
+    }.value,
+
+    imageNames in docker := {
+      val registry = cloudflowDockerRegistry.value
+      val namespace = cloudflowDockerRepository.value.orElse(registry.map(_ ⇒ "lightbend"))
```
RayRoestenburg: This does not seem right?

core/sbt-cloudflow/src/main/scala/cloudflow/sbt/CloudflowBasePlugin.scala
```diff
@@ -0,0 +1,126 @@
...
+    val log = streams.value.log
+    if (cloudflowBuildNumber.value.hasUncommittedChanges) {
+      log.warn(s"You have uncommitted changes in ${thisProjectRef.value.project}. Please commit all changes before publishing to guarantee a repeatable and traceable build.")
+    }
+  }
+
+  private[sbt] def makeStreamletDescriptorsLabelValue(streamletDescriptorsJson: JsValue) = {
+    // create a root object with the array
+    val streamletDescriptorsJsonStr =
+      JsObject("descriptors" -> streamletDescriptorsJson)
```
RayRoestenburg:
```suggestion
      JsObject("streamlet-descriptors" -> streamletDescriptorsJson)
```

core/sbt-cloudflow/src/main/scala/cloudflow/sbt/CloudflowAkkaPlugin.scala
```diff
@@ -0,0 +1,129 @@
...
+import cloudflow.blueprint.StreamletDescriptorFormat._
+import cloudflow.sbt.CloudflowKeys._
+
+object CloudflowAkkaPlugin extends CloudflowBasePlugin {
+
+  override def projectSettings = Seq(
+    libraryDependencies ++= Vector(
+      "com.lightbend.cloudflow" %% "cloudflow-akka-util" % BuildInfo.version,
+      "com.lightbend.cloudflow" %% "cloudflow-akka" % BuildInfo.version,
+      "com.lightbend.cloudflow" % "cloudflow-runner" % BuildInfo.version,
```
RayRoestenburg: Just curious why this one is not specified as a scala dep (%%)

core/sbt-cloudflow/src/main/scala/cloudflow/sbt/CloudflowAkkaPlugin.scala
```diff
@@ -0,0 +1,129 @@
...
+      "com.lightbend.cloudflow" %% "cloudflow-akka-testkit" % BuildInfo.version % "test"
+    ),
+
+    cloudflowAkkaDockerImageName := Def.task {
```
RayRoestenburg: I assume you could inline this in the imageNames in docker part and not have it be a setting?

core/sbt-cloudflow/src/main/scala/cloudflow/sbt/CloudflowBasePlugin.scala
```diff
@@ -0,0 +1,126 @@
...
+        .compactPrint
+
+    val compressed = zlibCompression(streamletDescriptorsJsonStr.getBytes(UTF_8))
+    Base64.getEncoder.encodeToString(compressed)
+  }
+
+  private[sbt] val showResultOfBuild = Def.task {
+    val log = streams.value.log
+    val imagePushed = (imageNames in docker).value.head // assuming we only build a single image!
```
RayRoestenburg: Does this work correcly with sbt multi project (sub projects?)

core/sbt-cloudflow/src/main/scala/cloudflow/sbt/CloudflowAkkaPlugin.scala
debasishg: Why ? Usually it's the organization - right ? Or should we make it `cloudflow` ?

core/sbt-cloudflow/src/main/scala/cloudflow/sbt/CloudflowAkkaPlugin.scala
debasishg: hmm .. Just wondering why this is so .. Checked up that we have this in the original version in pipelines as well. Any idea ?

core/sbt-cloudflow/src/main/scala/cloudflow/sbt/CloudflowAkkaPlugin.scala
```diff
@@ -0,0 +1,129 @@
...
+
+package cloudflow.sbt
+
+import sbt._
+import sbt.Keys._
+import sbtdocker._
+import sbtdocker.DockerKeys._
+import com.typesafe.sbt.packager.Keys._
+import spray.json._
+import cloudflow.sbt.CloudflowKeys._
```
skonto: Unused import.

core/sbt-cloudflow/src/main/scala/cloudflow/sbt/CloudflowBasePlugin.scala
debasishg: The plugin will be attached to one project (which becomes 1 sub-project in a multi-project build). So the image generation will be one sub-project at a time. If from the top level project we issue a `build` or `buildAndPublish` it will create one image per sub-project. This is what we need - right ?

core/sbt-cloudflow/src/main/scala/cloudflow/sbt/CloudflowBasePlugin.scala
```diff
@@ -0,0 +1,126 @@
...
+import sbt.Keys._
+import sbtdocker.DockerKeys._
+import com.typesafe.sbt.packager.archetypes._
+import spray.json._
+
+import cloudflow.sbt.CloudflowKeys._
+import cloudflow.blueprint.StreamletDescriptorFormat._
+import cloudflow.blueprint.StreamletDescriptor
+
+abstract class CloudflowBasePlugin extends AutoPlugin {
```
skonto: Could you add some documentation for the use of this class and what it does? It would help the readability of the code.

core/sbt-cloudflow/src/main/scala/cloudflow/sbt/CloudflowBasePlugin.scala
```diff
@@ -0,0 +1,126 @@
...
+  // If you change this, you have to make sure that all references to this value are changed
+  // - fsGroups on streamlet pods uses the GID to make volumes readable
+  val userInImage = "185" // default non-root user in the spark image
+  val userAsOwner: String ⇒ String = usr ⇒ s"$usr:cloudflow"
+  val streamletDescriptorsLabelName = "com.lightbend.cloudflow.streamlet-descriptors"
+
+  override def requires = CommonSettingsAndTasksPlugin && StreamletScannerPlugin &&
+    JavaAppPackaging && sbtdocker.DockerPlugin
+
+  // override def projectSettings = Seq ()
```
skonto: remove this?

core/sbt-cloudflow/src/main/scala/cloudflow/sbt/CloudflowAkkaPlugin.scala
debasishg: tried that .. looks like `imageNames in docker { //..` has to return a `TaskKey[Seq[ImageName]]`

core/sbt-cloudflow/src/main/scala/cloudflow/sbt/CloudflowAkkaPlugin.scala
debasishg: removed

core/sbt-cloudflow/src/main/scala/cloudflow/sbt/CloudflowBasePlugin.scala
debasishg: done ..

core/sbt-cloudflow/src/main/scala/cloudflow/sbt/CloudflowBasePlugin.scala
debasishg: removed ...

core/sbt-cloudflow/src/main/scala/cloudflow/sbt/CloudflowBasePlugin.scala
debasishg: done ..

core/sbt-cloudflow/src/main/scala/cloudflow/sbt/CloudflowAkkaPlugin.scala
skonto: Should we enforce some structure to group imports.

core/sbt-cloudflow/src/main/scala/cloudflow/sbt/CloudflowAkkaPlugin.scala
agemooij: I think this requires a small design discussion/document on whether we should really try to control the `docker push` process. Over time we have had some fundamental issues with allowing users to choose their own registry but most of our efforts have not been very intuitive for the user. 

In this particular PR, I think having a backup would be good. I think it should indeed be "cloudflow" since from the OSS point of view "lightbend" doesn't seem right.

@debasishg or @RayRoestenburg (or anyone else) perhaps someone can create a design document (or have a zoom discussion first) to see whether we should keep this functionality or just tell the users to push their images to their registries themselves?

core/sbt-cloudflow/src/main/scala/cloudflow/sbt/CloudflowAkkaPlugin.scala
agemooij: If `%%` works, make it so. If it doesn't, then that would be very strange and might need some investigation since AFAIK it is not a pure Java lib.

core/sbt-cloudflow/src/main/scala/cloudflow/sbt/CloudflowBasePlugin.scala
agemooij: Yes, that sounds correct.

core/sbt-cloudflow/src/main/scala/cloudflow/sbt/CloudflowBasePlugin.scala
```diff
@@ -0,0 +1,129 @@
...
+  private[sbt] def buildStreamletDescriptors(
+      detectedStreamlets: Map[String, Config]): Def.Initialize[Task[Iterable[StreamletDescriptor]]] = Def.task {
+    val detectedStreamletDescriptors = detectedStreamlets.map {
+      case (_, configDescriptor) ⇒
+        val jsonString = configDescriptor.root().render(ConfigRenderOptions.concise())
+        jsonString.parseJson.convertTo[cloudflow.blueprint.StreamletDescriptor]
+    }
+    detectedStreamletDescriptors
+  }
+
```
skonto: remove space?

core/sbt-cloudflow/src/main/scala/cloudflow/sbt/CloudflowAkkaPlugin.scala
skonto: This could also be part of the top abstract class or `CommonSettingsAndTasksPlugin` to underline the fact we are re-using a property, just in case we support new runtimes inthe future like kafka streams.

core/sbt-cloudflow/src/main/scala/cloudflow/sbt/CloudflowAkkaPlugin.scala
```diff
@@ -0,0 +1,128 @@
...
+      // this triggers side-effects, e.g. files being created in the staging area
+      cloudflowStageAppJars.value
+
+      val appDir: File = stage.value
+      val appJarsDir: File = new File(appDir, AppJarsDir)
+      val depJarsDir: File = new File(appDir, DepJarsDir)
+
+      // pack all streamlet-descriptors into a Json array
+      val streamletDescriptorsJson =
+        streamletDescriptorsInProject
```
skonto: move one line up?

skonto: @debasishg is there a way to avoid having this part as duplicate code across all plugins? 
We could refactor this to a helper function.

core/sbt-cloudflow/src/main/scala/cloudflow/sbt/CloudflowBasePlugin.scala
```diff
@@ -0,0 +1,129 @@
...
+  private[sbt] val checkUncommittedChanges = Def.task {
+    val log = streams.value.log
+    if (cloudflowBuildNumber.value.hasUncommittedChanges) {
+      log.warn(s"You have uncommitted changes in ${thisProjectRef.value.project}. Please commit all changes before publishing to guarantee a repeatable and traceable build.")
+    }
+  }
+
+  private[sbt] def makeStreamletDescriptorsLabelValue(streamletDescriptorsJson: JsValue) = {
```
skonto: This is more of a helper-static method to be used from a companion object. Same for the other ones that dont return a Task.

core/sbt-cloudflow/src/main/scala/cloudflow/sbt/CloudflowAkkaPlugin.scala
skonto: @debasishg this needs to be removed from the other classes too, eg. CloudflowSparkPlugin etc.

core/sbt-cloudflow/src/main/scala/cloudflow/sbt/CloudflowBasePlugin.scala
debasishg: Initially instead of an abstract base class `CloudflowBasePlugin` I made it an object `PluginUtils` and did an `import PluginUtils._` form the individual plugins. Then I figured out that we can reuse more if we make it a parent plugin e.g. `requires` etc. Hence the current shape. I think extracting out stuff into another companion object will be a bit of an over-engineering. We can view this as a base class that offers some _common_ behaviors that can be reused in derived classes. And it's not necessary that each of them will have to return a `Task` ..

core/sbt-cloudflow/src/main/scala/cloudflow/sbt/CloudflowAkkaPlugin.scala
debasishg: done ..

core/sbt-cloudflow/src/main/scala/cloudflow/sbt/CloudflowBasePlugin.scala
debasishg: done ..

core/sbt-cloudflow/src/main/scala/cloudflow/sbt/CloudflowAkkaPlugin.scala
RayRoestenburg: would also prefer `%%`, please check if that works.

core/sbt-cloudflow/src/main/scala/cloudflow/sbt/CloudflowAkkaPlugin.scala
RayRoestenburg: ok

core/sbt-cloudflow/src/main/scala/cloudflow/sbt/CloudflowAkkaPlugin.scala
RayRoestenburg: I think this is fine, like this. (assuming it is not an outstanding issue that needs resolution)

core/sbt-cloudflow/src/main/scala/cloudflow/sbt/CloudflowAkkaPlugin.scala
RayRoestenburg: This is not the organization I think. In docker hub it's the username for instance. Why do we have to default it?
This comes from the target-env setting right?

core/sbt-cloudflow/src/main/scala/cloudflow/sbt/CloudflowAkkaPlugin.scala
debasishg: We were stripping the runner jar of the version for Flink. Hence we used `%` .. Now that we have separate images, we can keep `%%` for Spark and Akka and keep `%` for Flink.

PR #127, opened 2020-02-10
Removed an unused function and fixed some warning.

olofwalker:


core/sbt-cloudflow/src/main/scala/cloudflow/sbt/StreamletScanner.scala
```diff
@@ -90,7 +90,7 @@ object StreamletScanner {
     if (!hasDefaultConstructor(streamletClass)) {
       Failure(ConstructorMissing(streamletClass))
     } else {
-      Try(streamletClass.newInstance()).recoverWith {
+      Try(streamletClass.getDeclaredConstructor().newInstance()).recoverWith {
```
debasishg: `klass.newInstance()` has been deprecated in Java 9. Is this the reason u changed it ? The only thing we need to check here is that the 2 calls don't throw same exceptions in case of failure. See https://stackoverflow.com/a/53265188. So we need to be sure that we are catching the right exceptions here for failures.

olofwalker: @debasishg Yes I noticed the warnings when I compiled core and was in the mood for some boy scouting, so `getDeclaredConstructor` throws `NoSuchMethodException` or `SecurityException` if the constructor was not found or is not accessible. Would that be sufficient ?

agemooij: @olofwalker compare it with the exceptions on the old `newInstance` method. The `NoSuchMethodException` is new and will need to be caught if we don't already catch one of its superclasses.

debasishg: @olofwalker I just now checked .. we use the `Try (..) recoverWith` idiom and wrap the exception in a `Failure` - so I guess we should be fine with this change ..

olofwalker: @debasishg Thanks!

PR #128, opened 2020-02-12
Age old advice: Composition over inheritance. All runtime plugins now compose with the base plugin

debasishg:

* Refactoring implementation : All runtime plugins now compose with `CloudflowBasePlugin` instead of inheriting from it

* Refactoring

* No

* Manually building application images

### DIFF
--- a/core/sbt-cloudflow/src/main/scala/cloudflow/sbt/CloudflowAkkaPlugin.scala
+++ b/core/sbt-cloudflow/src/main/scala/cloudflow/sbt/CloudflowAkkaPlugin.scala
@@ -24,30 +24,17 @@ import com.typesafe.sbt.packager.Keys._
 import spray.json._
 import cloudflow.sbt.CloudflowKeys._
 import cloudflow.blueprint.StreamletDescriptorFormat._
+import CloudflowBasePlugin._
 
-object CloudflowAkkaPlugin extends CloudflowBasePlugin {
+object CloudflowAkkaPlugin extends AutoPlugin {
+
+  override def requires = CloudflowBasePlugin
 
   override def projectSettings = Seq(
     libraryDependencies ++= Vector(
       "com.lightbend.cloudflow" %% "cloudflow-akka-util" % BuildInfo.version,
       "com.lightbend.cloudflow" %% "cloudflow-akka" % BuildInfo.version,
-      "com.lightbend.cloudflow" %% "cloudflow-runner" % BuildInfo.version,
       "com.lightbend.cloudflow" %% "cloudflow-akka-testkit" % BuildInfo.version % "test"
-    ),
-
-    cloudflowAkkaDockerImageName := Def.task {
-      Some(DockerImageName((ThisProject / name).value.toLowerCase, (ThisProject / cloudflowBuildNumber).value.buildNumber))
-    }.value,
-
-    streamletDescriptorsInProject := Def.taskDyn {
-      val detectedStreamlets = cloudflowStreamletDescriptors.value
-      buildStreamletDescriptors(detectedStreamlets)
-    }.value,
-
-    buildOptions in docker := BuildOptions(
-      cache = true,
-      removeIntermediateContainers = BuildOptions.Remove.OnSuccess,
-      pullBaseImage = BuildOptions.Pull.IfMissing
     ),
 
     cloudflowStageAppJars := Def.taskDyn {
@@ -66,22 +53,6 @@ object CloudflowAkkaPlugin extends CloudflowBasePlugin {
         }
       }
     }.value,
-
-    imageNames in docker := {
-      val registry = cloudflowDockerRegistry.value
-      val namespace = cloudflowDockerRepository.value
-
-      cloudflowAkkaDockerImageName.value
-        .map { imageName ⇒
-          ImageName(
-            registry = registry,
-            namespace = namespace,
-            repository = imageName.name,
-            tag = Some(imageName.tag)
-          )
-        }
-        .toSeq
-    },
 
     dockerfile in docker := {
       // this triggers side-effects, e.g. files being created in the staging area
@@ -106,20 +77,5 @@ object CloudflowAkkaPlugin extends CloudflowBasePlugin {
         label(StreamletDescriptorsLabelName, streamletDescriptorsLabelValue)
       }
     },
-
-    build := showResultOfBuild.dependsOn(
-      docker.dependsOn(
-        checkUncommittedChanges
-      )
-    ).value,
-
-    buildAndPublish := showResultOfBuildAndPublish.dependsOn(
-      dockerBuildAndPush.dependsOn(
-        checkUncommittedChanges,
-        verifyDockerRegistry
-      )
-    ).value,
-
-    fork in Compile := true
   )
 }
--- a/core/sbt-cloudflow/src/main/scala/cloudflow/sbt/CloudflowBasePlugin.scala
+++ b/core/sbt-cloudflow/src/main/scala/cloudflow/sbt/CloudflowBasePlugin.scala
@@ -25,6 +25,7 @@ import com.typesafe.config._
 
 import sbt._
 import sbt.Keys._
+import sbtdocker._
 import sbtdocker.DockerKeys._
 import com.typesafe.sbt.packager.archetypes._
 import spray.json._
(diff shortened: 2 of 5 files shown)


core/sbt-cloudflow/src/main/scala/cloudflow/sbt/CloudflowBasePlugin.scala
```diff
@@ -58,6 +59,58 @@ abstract class CloudflowBasePlugin extends AutoPlugin {
   override def requires = CommonSettingsAndTasksPlugin && StreamletScannerPlugin &&
     JavaAppPackaging && sbtdocker.DockerPlugin
 
+  override def projectSettings = Seq(
+    libraryDependencies ++= Vector(
+      "com.lightbend.cloudflow" % "cloudflow-runner"         % BuildInfo.version
```
debasishg: This has to be `%` and not `%%` since we need the runner jar without any version information for Flink. Hence we build the runner jar from `core` without the version information.

core/sbt-cloudflow/src/main/scala/cloudflow/sbt/CloudflowBasePlugin.scala
```diff
@@ -58,6 +59,58 @@ abstract class CloudflowBasePlugin extends AutoPlugin {
...
+    }.value,
+
+    buildOptions in docker := BuildOptions(
+      cache = true,
+      removeIntermediateContainers = BuildOptions.Remove.OnSuccess,
+      pullBaseImage = BuildOptions.Pull.IfMissing
+    ),
+
+    imageNames in docker := {
+      val registry = cloudflowDockerRegistry.value
```
olofwalker: are these val's used anywhere else?

core/sbt-cloudflow/src/main/scala/cloudflow/sbt/CloudflowAkkaPlugin.scala
```diff
@@ -24,32 +24,19 @@ import com.typesafe.sbt.packager.Keys._
 import spray.json._
 import cloudflow.sbt.CloudflowKeys._
 import cloudflow.blueprint.StreamletDescriptorFormat._
+import CloudflowBasePlugin._
 
-object CloudflowAkkaPlugin extends CloudflowBasePlugin {
+object CloudflowAkkaPlugin extends AutoPlugin {
+
+  override def requires = CloudflowBasePlugin
```
olofwalker: ahh, so Autoplugin can "require" the behavior and settings of a another plugin (?)

core/sbt-cloudflow/src/main/scala/cloudflow/sbt/CloudflowFlinkPlugin.scala
```diff
@@ -25,30 +25,20 @@ import spray.json._
 import cloudflow.sbt.CloudflowKeys._
 import cloudflow.blueprint.StreamletDescriptorFormat._
 import cloudflow.sbt.CloudflowKeys._
+import CloudflowBasePlugin._
 
-object CloudflowFlinkPlugin extends CloudflowBasePlugin {
+object CloudflowFlinkPlugin extends AutoPlugin {
   final val flinkVersion                  = "1.9.2"
```
olofwalker: constant = Captilize first letter ?

olofwalker: There are a few in the Flink plugin also.

core/sbt-cloudflow/src/main/scala/cloudflow/sbt/CloudflowBasePlugin.scala
debasishg: These need to be set for `sbt-docker` plugin for building the docker image

core/sbt-cloudflow/src/main/scala/cloudflow/sbt/CloudflowAkkaPlugin.scala
debasishg: yes! That's more idiomatic way of composing plugins instead of using inheritance

core/sbt-cloudflow/src/main/scala/cloudflow/sbt/CloudflowFlinkPlugin.scala
debasishg: thanks for pointing out - will change

core/sbt-cloudflow/src/main/scala/cloudflow/sbt/CloudflowBasePlugin.scala
olofwalker: So they are an override?

core/sbt-cloudflow/src/main/scala/cloudflow/sbt/CloudflowBasePlugin.scala
agemooij: Might be worth a brief inline comment to explain this to future readers of the code.

[very minor] also, the big gap between the artifact id and the version is weird since there is nothing to align with.

core/sbt-cloudflow/src/main/scala/cloudflow/sbt/CloudflowBasePlugin.scala
agemooij: They are local vals and so I think @olofwalker is right and e could just inline them since they each seem to be used only once.

core/sbt-cloudflow/src/main/scala/cloudflow/sbt/CloudflowFlinkPlugin.scala
agemooij: I'm slightly worried about the large number of places where we hardcode these version Strings but I think that is a separate issue to look into and should not block this PR. Something to think about though...

core/sbt-cloudflow/src/main/scala/cloudflow/sbt/CloudflowSparkPlugin.scala
```diff
@@ -25,20 +25,22 @@ import spray.json._
 
 import cloudflow.blueprint.StreamletDescriptorFormat._
 import cloudflow.sbt.CloudflowKeys._
+import CloudflowBasePlugin._
 
-object CloudflowSparkPlugin extends CloudflowBasePlugin {
+object CloudflowSparkPlugin extends AutoPlugin {
   final val sparkVersion                  = "2.4.4"
   final val cloudflowVersion              = "1.3.1-SNAPSHOT"
```
agemooij: @debasishg  How are we going to replace this hardcoded version number during a release?

core/sbt-cloudflow/src/main/scala/cloudflow/sbt/CloudflowBasePlugin.scala
debasishg: They cannot be inlined since you cannot have an sbt `TaskKey` evaluation within an anonymous function. Hence I have to evaluate it outside the `.map { ..` ..
Screen Shot 2020-02-12 at 2 36 55 PM

core/sbt-cloudflow/src/main/scala/cloudflow/sbt/CloudflowSparkPlugin.scala
debasishg: Yeah .. we can have it in the base plugin but that does not solve the problem though. But a bigger question which we started discussing is if we need to have the cloudflow version as part of the the base image name.

core/sbt-cloudflow/src/main/scala/cloudflow/sbt/CloudflowBasePlugin.scala
debasishg: I will add the comment. Regarding the big gap I think it's from `scalafmt` ..

core/sbt-cloudflow/src/main/scala/cloudflow/sbt/CloudflowBasePlugin.scala
olofwalker: @debasishg Thanks for that clarification, as usual my sbt-fu is lacking

agemooij:
PS: I'm not amused by your PR title "Age old advice" since I'm not **THAT** old and it's a bit "pot kettle black" 😉

core/sbt-cloudflow/src/main/scala/cloudflow/sbt/CloudflowSparkPlugin.scala
agemooij: Yes, I still had a tab open with the following comment:

> Here's my €0.02: 
> 
> I think the issue here is mostly related to release lifecycles and the question whether that is different for base images vs the other project artifacts.
> 
> I do think there is value in marking the base images that are part of an official release with the verified version number of that release, even though that might mean recreating the same image but with a different tag for releases where nothing has changed in the base image.
> 
> That said, I think that should be a special case and done as part of our release automation, which can also update existing example projects to use the

PR #131, opened 2020-02-18
Removed obsolete M1 string in image tags

yuchaoran2011:
No changes made to the images themselves. Just re-tagged Spark, Spark operator and Cloudflow base images and re-pushed them.


yuchaoran2011:
Yeah agreed. I'm going to merge it despite a weird build error that seems unrelated and caused by a networking error when fetching dependencies during build.

PR #139, opened 2020-02-21
This fixes a index out of range crash in the CLI.

olofwalker:


kubectl-cloudflow/version/version.go
```diff
@@ -43,10 +43,14 @@ func GetProtocolVersionConfigMap() (*corev1.ConfigMap, error) {
...
 		metav1.ListOptions{LabelSelector: labels.Set(labelSelector.MatchLabels).String()})
+
 	if err == nil {
 		if len(configMaps.Items) > 1 {
 			return nil, errors.New("Multiple Cloudflow operators detected in the cluster. This is not supported. Exiting")
 		}
+		if len(configMaps.Items) < 1 {
+			return nil, errors.New("No Cloudflow operators detected in the cluster. Exiting")
+		}
 		return &configMaps.Items[0], nil
```
agemooij: Looks good. I guess we could instead also just check whether there is exactly one instance and report everything else as a failure.

olofwalker: This is more informative IMO

PR #142, opened 2020-02-24
Dependency override for akka-http to take care of older version in skuber

debasishg:

* Dependency override for `akka-http` for skuber

* skuber uses an older version of `akka-http` which is not compatible with akka `2.6.3`. Hence need this dependency override

* fixes a problem of crashing operator

* manually tested on a GKE cluster

### DIFF
--- a/core/build.sbt
+++ b/core/build.sbt
@@ -359,6 +359,10 @@ lazy val operator =
 
       publishArtifact in (Compile, packageDoc) := false,
       publishArtifact in (Compile, packageSrc) := false,
+      // skuber version 2.4.0 depends on akka-http 10.1.9 : hence overriding
+      // with akka-http 10.1.11 to use akka 2.6.3
+      // remove this override once skuber is updated
+      dependencyOverrides += AkkaHttp,
 
       buildOptions in docker := BuildOptions(
         cache = true,


core/build.sbt
```diff
@@ -359,6 +359,10 @@ lazy val operator =
 
       publishArtifact in (Compile, packageDoc) := false,
       publishArtifact in (Compile, packageSrc) := false,
+      // skuber version 2.4.0 depends on akka-http 10.1.9 : hence overriding
+      // with akka-http 10.1.11 to use akka 2.6.3
+      // remove this override once skuber is updated
+      dependencyOverrides += "com.typesafe.akka" %% "akka-http" % Version.AkkaHttp,
```
agemooij: I would just add a simple dependency on the pre-defined AkkaHttp dependency in the main dependencies list of the project. It will have the same effect and will be slightly easier to maintain.

agemooij: We already do the same for the `AkkaStream` and `AkkaSlf4j` dependencies, probably for the exact same reason.

debasishg: @agemooij  do you mean to use `force()` ?

```
val AkkaHttp  = "com.typesafe.akka"   %% "akka-http"   % Version.AkkaHttp force()
```

agemooij: No, simpler than that. Just add a normal, run of the mill dependency to the right version. As long as the version we want is the latest version, the normal deendency evition mechanism will use that instead of any outdated transitives.

So just add it like this:

```
    .settings(
        AkkaSlf4j,
        AkkaStream,
        AkkaHttp,     // <-----
        Ficus,
        .....
```

skonto: It uses `10.1.9` and we go straight to 10.1.11, I guess it is safe.

skonto: Also let's add a comment somewhere to take a note for this behavior.

agemooij: FYI: we were already using 10.1.10 on other sub-projects.

@skonto Do you mean that Skuber also depends on those two artifacts?

debasishg: ok .. if it works the same way as `dependencyOverrides` it should be fine ..

debasishg: skuber does not depend on them. But skuber also depends on akka-stream .. not sure changing that to `2.6.3` will have any other impact ..

debasishg: But I chose to stick with `dependencyOverrides` as sbt doc recommends it to be the preferred way

skonto: @agemooij @debasishg we list the other too artifacts in Dependencies file for the core project. They both depend on the akka-http version are we sure that we get the right version of them?

PR #143, opened 2020-02-24
A streamlet gets a fixed port (3000) if it has an endpoint.

RayRoestenburg:

Fixes the streamlet endpoint to a single value.
- A streamlet instance can (currently) only have one or zero endpoints.
- One streamlet instance eventually runs in one pod.
- For local running, the endpoints are modified so everything can run in a single JVM.

Requested by users, it is easier if service ports and container ports do not change, when setting uo ingresses.

Yes, the container port for a streamlet (if it has an endpoint) is now always set to 3000.

- [x] unit tested.
- [x] local running examples (taxi-ride example for instance).
- [x] Running an example on a GKE cluster.
- [x] Checked all docs for examples of ports higher than 3000

### DIFF
--- a/core/cloudflow-blueprint/src/main/scala/cloudflow/blueprint/deployment/ApplicationDescriptor.scala
+++ b/core/cloudflow-blueprint/src/main/scala/cloudflow/blueprint/deployment/ApplicationDescriptor.scala
@@ -81,10 +81,9 @@ object ApplicationDescriptor {
     val namedStreamletDescriptors = blueprint.streamlets.map(streamletToNamedStreamletDescriptor)
     val connections               = blueprint.connections.map(toConnection)
     val deployments =
-      namedStreamletDescriptors.zipWithIndex
-        .map {
-          case (streamlet, index) ⇒
-            StreamletDeployment(sanitizedApplicationId, streamlet, image, index, connections)
+      namedStreamletDescriptors
+        .map { streamlet ⇒
+          StreamletDeployment(sanitizedApplicationId, streamlet, image, connections)
         }
 
     ApplicationDescriptor(sanitizedApplicationId,
@@ -152,18 +151,18 @@ final case class StreamletDeployment(
 )
 
 object StreamletDeployment {
-  val ServerAttributeName          = "server"
-  val MinimumEndpointContainerPort = 3000
+  val ServerAttributeName   = "server"
+  val EndpointContainerPort = 3000
 
   def name(appId: String, streamlet: String) = s"${appId}.${streamlet}"
 
   def apply(appId: String,
             streamlet: StreamletInstance,
             image: String,
-            index: Int,
             allConnections: Vector[Connection],
+            containerPort: Int = EndpointContainerPort,
             replicas: Option[Int] = None): StreamletDeployment = {
-    val (config, endpoint) = configAndEndpoint(appId, streamlet, index)
+    val (config, endpoint) = configAndEndpoint(appId, streamlet, containerPort)
     StreamletDeployment(
       name(appId, streamlet.name),
       streamlet.descriptor.runtime.name,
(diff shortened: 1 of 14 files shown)


core/cloudflow-blueprint/src/main/scala/cloudflow/blueprint/deployment/ApplicationDescriptor.scala
```diff
@@ -152,18 +151,17 @@ final case class StreamletDeployment(
 )
 
 object StreamletDeployment {
-  val ServerAttributeName          = "server"
-  val MinimumEndpointContainerPort = 3000
+  val ServerAttributeName   = "server"
+  val EndpointContainerPort = 3000
```
debasishg: What happens when we have multiple end points ? e.g. 2 ingresses especially for local running ..

agemooij: @debasishg see below. The `LocalRunner` is fixed to deal with this directly.

debasishg: 👍 .. missed it ..

core/cloudflow-runner/src/main/scala/cloudflow/runner/LocalRunner.scala
```diff
@@ -112,25 +112,31 @@ object LocalRunner extends StreamletLoader {
...
-          .withFallback(baseConfig)
-          .withValue("cloudflow.local", ConfigValueFactory.fromAnyRef(true))
-
-        (streamletDescriptor, patchedRunnerConfig)
+    var endpointIdx = 0
+    val streamletsWithConf = streamlets.map { streamletInstance ⇒
+      val streamletName = streamletInstance.name
+      // Make sure that we convert any backslash in the path to a forward slash since we want to store this in a JSON value
+      val localStorageDirectory =
+        Files.createTempDirectory(s"local-runner-storage-${streamletName}").toFile.getAbsolutePath.replace('\\', '/')
```
skonto: Can we do this at the json encoding side? What are the issues if we escape it?

RayRoestenburg: I think it makes sense to make sure that the path is correct before putting it in JSON, but what you say could work (escaping). This is code that will likely get changed a lot with new local running, so I'll leave it like this for now.

agemooij: @RayRoestenburg what impact does this have for local running on Windows?

cc: @olofwalker

RayRoestenburg: This code has not changed guys.. we have tested local running on Windows.

core/cloudflow-runner/src/main/scala/cloudflow/runner/LocalRunner.scala
```diff
@@ -112,25 +112,31 @@ object LocalRunner extends StreamletLoader {
...
+      log.info(s"Using local storage directory: $localStorageDirectory")
+      val deployment: StreamletDeployment =
+        StreamletDeployment(appDescriptor.appId,
+                            streamletInstance,
+                            "",
+                            appDescriptor.connections,
+                            StreamletDeployment.EndpointContainerPort + endpointIdx)
+      deployment.endpoint.foreach(_ => endpointIdx += 1)
```
maasg: I wonder whether a simple  `endpointIdx += 1` wouldn't work here (or even better, `zipWithIndex.map(=>... ) ` above)

RayRoestenburg: I do not want to count for the streamlets without endpoint

RayRoestenburg: In the example I noticed 3002, 3003 for taxi ride app instead of 3000, 3001.

maasg: ack!

PR #147, opened 2020-02-25
Adds support for scalafmt

skonto:
Enables scalafmt on compile at the core and examples sub-projects. Also adds the required logic at the github actions side to fail if format is wrong.

We need to be able to fail if format is not correct at a PR stage and also check it by default when we compile the code.

No

manually by having the wrong format and running the scripts/build-all.sh script.

### DIFF
--- a/core/build.sbt
+++ b/core/build.sbt
@@ -7,10 +7,11 @@ import sbtrelease.ReleaseStateTransformations._
 
 lazy val root =
   Project(id = "root", base = file("."))
-    .enablePlugins(ScalaUnidocPlugin, JavaUnidocPlugin)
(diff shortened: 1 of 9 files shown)


skonto:
@agemooij @RayRoestenburg created a new PR and closed the other which was messed up.

skonto:
Build fails:
"[warn] /home/runner/work/cloudflow/cloudflow/core/sbt-cloudflow/src/main/scala/cloudflow/sbt/CloudflowSparkPlugin.scala isn't formatted properly!"
Cool that means it works. This is due to the latest commits.

RayRoestenburg:
LGTM @skonto !

PR #148, opened 2020-02-25
Owner references

olofwalker:
This PR adds a chain over ownership to all K8s resources created by the Cloudflow operator when deploying an application.

This changes also the behavior when undeploying an application, it is no longer possible to retain the Kafka topics when undeploying an application.


core/build.sbt
```diff
@@ -334,161 +331,150 @@ lazy val operator =
...
-          registry = None,
-          namespace = Some("lightbend"),
-          repository = "cloudflow-operator",
-          tag = Some(cloudflowBuildNumber.value.asVersion)
-        )
-      ),
-
+            ImageName(
+              registry = None,
+              namespace = Some("olofwalker"),
```
yuchaoran2011: This isn't meant to be committed, right?

core/cloudflow-operator/src/main/scala/cloudflow/operator/CloudflowApplication.scala
```diff
@@ -107,16 +107,8 @@ object CloudflowApplication {
...
     CustomResource(
-      Spec(
-        appId,
-        appVersion,
-        streamlets,
-        connections,
-        deployments,
-        agentPaths
-      )
+      applicationSpec
```
RayRoestenburg: 👍

core/cloudflow-operator/src/main/scala/cloudflow/operator/CloudflowOwnerReferences.scala
```diff
@@ -0,0 +1,26 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package cloudflow.operator
+
+import skuber.OwnerReference
+
+case class CloudflowOwnerReferences(list: List[OwnerReference])
```
RayRoestenburg: There is only one owner right (the CR)? what would be wrong with using something like this instead?
```
case class CloudflowOwnerReference(owner: OwnerReference)
```

core/cloudflow-operator/src/main/scala/cloudflow/operator/Main.scala
```diff
@@ -45,8 +48,9 @@ object Main extends {
 
       HealthChecks.serve(settings)
 
-      val client = connectToKubernetes()
-      installProtocolVersion(client.usingNamespace(settings.podNamespace))
+      val client          = connectToKubernetes()
+      val ownerReferences = getDeploymentOwnerReference(settings, client.usingNamespace(settings.podNamespace))
```
RayRoestenburg: minor: ownerRefernces plural, getDeploymentOwnerReference singular

RayRoestenburg: REading this I wonder if it would help to make the ownerReference implicit.

core/cloudflow-operator/src/main/scala/cloudflow/operator/Main.scala
```diff
@@ -87,6 +91,14 @@ object Main extends {
       """.stripMargin)
   }
 
+  private def getDeploymentOwnerReference(settings: Settings, client: skuber.api.client.KubernetesClient)(implicit ec: ExecutionContext) =
+    CloudflowOwnerReferences(
+      Await.result(client
+                     .getInNamespace[Deployment](Name.ofCloudflowOperatorDeployment, settings.podNamespace)
+                     .map(_.metadata.ownerReferences),
```
RayRoestenburg: Ah this is why it is a list.

RayRoestenburg: Another noob question: so it does not work to get the CR itself and get the metadata.uid instead and use that as an owner reference?

core/cloudflow-operator/src/main/scala/cloudflow/operator/runner/FlinkRunner.scala
```diff
@@ -106,15 +106,17 @@ object FlinkRunner extends Runner[CR] {
...
 
     CustomResource[Spec, Status](jobSpec)
       .withMetadata(
         ObjectMeta(
           name = name,
           namespace = namespace,
           annotations = Map("prometheus.io/scrape" -> "true", "prometheus.io/port" -> PrometheusConfig.PrometheusJmxExporterPort.toString),
-          labels = labels
+          labels = labels,
+          ownerReferences = ownerReferences.list
```
RayRoestenburg: do you ever not use the list? maybe it is easier to just pass a `List[OwnerReference]` around?

core/cloudflow-operator/src/main/scala/cloudflow/operator/Main.scala
RayRoestenburg: Oh wait this is another owner reference, you want the owner that created the cloudflow operator..

core/cloudflow-operator/src/main/scala/cloudflow/operator/action/AppActions.scala
```diff
@@ -69,23 +70,25 @@ object AppActions {
...
 
-  private def akkaRole(namespace: String, labels: CloudflowLabels): Role =
+  private def akkaRole(namespace: String, labels: CloudflowLabels, ownerReferences: CloudflowOwnerReferences): Role =
     Role(
       metadata = ObjectMeta(
         name = Name.ofAkkaRole(),
         namespace = namespace,
-        labels = labels(Name.ofAkkaRole)
+        labels = labels(Name.ofAkkaRole),
+        ownerReferences = ownerReferences.list
```
RayRoestenburg: yeah I think it will be easier just to pass along `List[OwnerReference]` I don't think we really need to `CloudflowOwnerReferences` type.

core/build.sbt
olofwalker: No, have to clean this PR up a bit, was added in a hurry yesterday evening.

core/cloudflow-operator/src/main/scala/cloudflow/operator/CloudflowOwnerReferences.scala
olofwalker: The k8s concept is a list, so, in the end, you add a list to the K8s object, you could argue that the conversion could happen then instead of here.

core/cloudflow-operator/src/main/scala/cloudflow/operator/action/AppActions.scala
olofwalker: Removed it.

core/cloudflow-operator/src/main/scala/cloudflow/operator/runner/FlinkRunner.scala
olofwalker: Changed it to use a `List[]` instead

core/cloudflow-operator/project/build.properties
```diff
@@ -0,0 +1 @@
+sbt.version=1.3.4
```
agemooij: This should be 1.3.8

PR #149, opened 2020-02-25
Fixed a Flink 1.10 issue

yuchaoran2011:
Updated the `config.sh` script to be the one provided in Flink 1.10. 

Without this PR, the following error will occur in the task manager when running `taxi-ride-fare` app:
```
/opt/flink/bin/taskmanager.sh: line 52: runBashJavaUtilsCmd: command not found
/opt/flink/bin/taskmanager.sh: line 53: extractExecutionParams: command not found
[ERROR] Could not get JVM parameters properly.
```
This is because the two functions mentioned in the error were only introduced in `config.sh` script in Flink 1.10.

Yes. Fixed a user-facing bug.

By running `taxi-ride-fare` and making sure the Flink job is running smoothly.

### DIFF
--- a/external/multi-base-images/flink/flink-entrypoint.sh
+++ b/external/multi-base-images/flink/flink-entrypoint.sh
@@ -42,6 +42,8 @@ fi
 echo "web.upload.dir: $FLINK_HOME" >> "$FLINK_HOME/conf/flink-conf.yaml"
 echo "jobmanager.web.upload.dir: $FLINK_HOME" >> "$FLINK_HOME/conf/flink-conf.yaml"
 
+echo "taskmanager.memory.flink.size: 1024mb" >> "$FLINK_HOME/conf/flink-conf.yaml"
+
 # Add JMX metric reporter to config
 echo "metrics.reporters: jmx" >> "$FLINK_HOME/conf/flink-conf.yaml"
 
--- a/external/base-image/config.sh
+++ b/external/base-image/config.sh
@@ -1,5 +1,5 @@
 #!/usr/bin/env bash
-#
+################################################################################
 #  Licensed to the Apache Software Foundation (ASF) under one
 #  or more contributor license agreements.  See the NOTICE file
 #  distributed with this work for additional information
@@ -15,14 +15,14 @@
 #  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 #  See the License for the specific language governing permissions and
 # limitations under the License.
-#
+################################################################################
 
 constructFlinkClassPath() {
     local FLINK_DIST
     local FLINK_CLASSPATH
 
     while read -d '' -r jarfile ; do
-        if [[ "$jarfile" =~ .*flink-dist.*.jar ]]; then
+        if [[ "$jarfile" =~ .*/flink-dist[^/]*.jar$ ]]; then
             FLINK_DIST="$FLINK_DIST":"$jarfile"
         elif [[ "$FLINK_CLASSPATH" == "" ]]; then
             FLINK_CLASSPATH="$jarfile";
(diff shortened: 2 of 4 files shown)


external/multi-base-images/flink/flink-entrypoint.sh
```diff
@@ -42,6 +42,8 @@ fi
 echo "web.upload.dir: $FLINK_HOME" >> "$FLINK_HOME/conf/flink-conf.yaml"
 echo "jobmanager.web.upload.dir: $FLINK_HOME" >> "$FLINK_HOME/conf/flink-conf.yaml"
 
+echo "taskmanager.memory.flink.size: 1024mb" >> "$FLINK_HOME/conf/flink-conf.yaml"
```
yuchaoran2011: A larger default needs to be provided. Without this line, the following error will occur:
```
 - 'taskmanager.memory.flink.size' is not specified, use the configured deprecated task manager heap value (512.000mb (536870912 bytes)) for it.
 - The derived from fraction network memory (51.200mb (53687092 bytes)) is less than its min value 64.000mb (67108864 bytes), min value will be used instead
Exception in thread "main" org.apache.flink.configuration.IllegalConfigurationException: Sum of configured Framework Heap Memory (128.000mb (134217728 bytes)), Framework Off-Heap Memory (128.000mb (134217728 bytes)), Task Off-Heap Memory (0 bytes), Managed Memory (204.800mb (214748368 bytes)) and Network Memory (64.000mb (67108864 bytes)) exceed configured Total Flink Memory (512.000mb (536870912 bytes)).
	at org.apache.flink.runtime.clusterframework.TaskExecutorProcessUtils.deriveInternalMemoryFromTotalFlinkMemory(TaskExecutorProcessUtils.java:320)
	at org.apache.flink.runtime.clusterframework.TaskExecutorProcessUtils.deriveProcessSpecWithTotalFlinkMemory(TaskExecutorProcessUtils.java:221)
	at org.apache.flink.runtime.clusterframework.TaskExecutorProcessUtils.processSpecFromConfig(TaskExecutorProcessUtils.java:143)
	at org.apache.flink.runtime.util.BashJavaUtils.getTmResourceJvmParams(BashJavaUtils.java:62)
	at org.apache.flink.runtime.util.BashJavaUtils.main(BashJavaUtils.java:46)
[ERROR] Could not get JVM parameters properly.
```

debasishg:
@yuchaoran2011 Previously we used to add the cloudflow jars to the classpath in `config.sh`. Don't we need to do that now ? And did u test with the `taxi-ride` example if it runs properly ?

yuchaoran2011:
@debasishg Yes we still do that. I left that part (just one line unless I'm mistaken) in config.sh untouched. I tested this PR with taxi-ride-fare and it runs smoothly.

agemooij:
@yuchaoran2011 did you align this new memory setting with the resources we are actually setting on the pods through K8s and the operator?

It's going to be tricky setting such hardcoded values since they don't automatically scale along with what the user sets at the K8s level. Any ideas on how to make this more dynamic?

yuchaoran2011:
@agemooij That's a very good point. I thought about it when testing this PR. But opted for a quick and dirty solution to fix the broken master first. For overall task manager resource requirements, we configure some of them in [application.conf](https://github.com/lightbend/cloudflow/blob/master/core/cloudflow-operator/src/main/resources/application.conf#L143). So instead of hard-coding the value, we can make it configurable there too. And Flink operator can check whether the overall resource requested is compatible with this `taskmanager.memory.flink.size` parameter. But that's going to be a feature we'll need to contribute. 
I've created [CSP-1903](https://lightbend.atlassian.net/browse/CSP-1903) to track this issue.

agemooij:
Thanks @yuchaoran2011 !

Lockdain:
We face the issue with the default setting of 1024 mb at #805

PR #150, opened 2020-02-26
use single shot execution to avoid waiting until timeout

maasg:
This PR is an attempt to improve the flaky Spark tests that often fails our CI builds.
Using the single execution mode (`Trigger.Once`), we allow the query to stop early and prevent the long 30s wait.


agemooij:
It works great locally. Execution time went from a guaranteed 30 seconds to 1-2 seconds. So if nothing else it speeds up the build. I hope it also gets rid of our unpredictable failures for this test.

PR #152, opened 2020-02-26
Improve Spark test execution time

maasg:
This PR fixes the race condition in the `SparkEgressEgressSpec`. The cause was that in the test, there are two queries being executed, but the wait exits after any of them finish, resulting in an unknown state for the other query.
- If it had finished as well => test pass
- If it was not finished yet => test fail

In this PR, we introduce a wait for all queries with a manual  timeout check to avoid hanging there forever in case of failures.
Also, the execution of all queries is changed to "Once", making the tests run much faster.

`sbt:cloudflow-spark-tests> test`

Before:
```
[success] Total time: 22 s, completed Feb 26, 2020 8:30:08 PM
```

After:
```
[success] Total time: 5 s, completed Feb 26, 2020 9:01:18 PM
```

### DIFF
--- a/core/cloudflow-spark-testkit/src/main/scala/cloudflow/spark/testkit/SparkStreamletTestkit.scala
+++ b/core/cloudflow-spark-testkit/src/main/scala/cloudflow/spark/testkit/SparkStreamletTestkit.scala
@@ -221,8 +221,12 @@ final case class SparkStreamletTestkit(session: SparkSession, config: Config = C
       sparkStreamlet: SparkStreamlet,
       duration: Duration
   ): Unit = {
+    val t0 = System.currentTimeMillis()
     val queryExecution = sparkStreamlet.setContext(ctx).run(ctx.config)
-    session.streams.awaitAnyTermination(duration.toMillis)
+    while (session.streams.active.nonEmpty && (System.currentTimeMillis()-t0)<duration.toMillis ) {
+      session.streams.awaitAnyTermination(duration.toMillis)
+      session.streams.resetTerminated()
+    }
     queryExecution.stop()
     Await.result(queryExecution.completed, duration)
   }
--- a/core/cloudflow-spark-testkit/src/main/scala/cloudflow/spark/testkit/SparkStreamletTestkit.scala
+++ b/core/cloudflow-spark-testkit/src/main/scala/cloudflow/spark/testkit/SparkStreamletTestkit.scala
@@ -221,9 +221,9 @@ final case class SparkStreamletTestkit(session: SparkSession, config: Config = C
       sparkStreamlet: SparkStreamlet,
       duration: Duration
   ): Unit = {
-    val t0 = System.currentTimeMillis()
+    val t0             = System.currentTimeMillis()
     val queryExecution = sparkStreamlet.setContext(ctx).run(ctx.config)
-    while (session.streams.active.nonEmpty && (System.currentTimeMillis()-t0)<duration.toMillis ) {
+    while (session.streams.active.nonEmpty && (System.currentTimeMillis() - t0) < duration.toMillis) {
       session.streams.awaitAnyTermination(duration.toMillis)
       session.streams.resetTerminated()
     }
--- a/core/cloudflow-spark-testkit/src/main/scala/cloudflow/spark/testkit/TestSparkStreamletContext.scala
+++ b/core/cloudflow-spark-testkit/src/main/scala/cloudflow/spark/testkit/TestSparkStreamletContext.scala
@@ -20,12 +20,11 @@ package testkit
 import java.nio.file.attribute.FileAttribute
 
 import com.typesafe.config._
+
 import scala.reflect.runtime.universe._
-
-import org.apache.spark.sql.{ Dataset, Encoder, SparkSession }
+import org.apache.spark.sql.{Dataset, Encoder, SparkSession}
 import org.apache.spark.sql.execution.streaming.MemoryStream
-import org.apache.spark.sql.streaming.{ OutputMode, StreamingQuery }
-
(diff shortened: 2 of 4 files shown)


debasishg:
@maasg I see another test failure from applications ..
Screen Shot 2020-02-27 at 11 09 59 AM

maasg:
@debasishg thanks! It's a tricky case. That test depends on having time to generate data. 
It looks like it can't run under the `Trigger.Once` trigger semantic.

core/cloudflow-spark-testkit/src/main/scala/cloudflow/spark/testkit/TestSparkStreamletContext.scala
```diff
@@ -60,6 +59,7 @@ private[testkit] class TestSparkStreamletContext(override val streamletRef: Stri
         stream.writeStream
           .outputMode(outputMode)
           .format("memory")
+          .trigger(Trigger.Once)
```
skonto: Is it way slower if we omit this?

core/cloudflow-spark-testkit/src/main/scala/cloudflow/spark/testkit/SparkStreamletTestkit.scala
```diff
@@ -221,8 +221,12 @@ final case class SparkStreamletTestkit(session: SparkSession, config: Config = C
       sparkStreamlet: SparkStreamlet,
       duration: Duration
   ): Unit = {
+    val t0 = System.currentTimeMillis()
     val queryExecution = sparkStreamlet.setContext(ctx).run(ctx.config)
-    session.streams.awaitAnyTermination(duration.toMillis)
+    while (session.streams.active.nonEmpty && (System.currentTimeMillis()-t0)<duration.toMillis ) {
+      session.streams.awaitAnyTermination(duration.toMillis)
```
skonto: Another options would be to call `processAllAvailable` and also be more strict: `session.streams.active.foreach(x => x.awaitTermination())` if any query fails then we should fail immediately.

maasg: Interesting idea. I guess this is a tricky situation.  `streams.active` will be evaluated only once and it's possible that after we wait for the first query to terminate, other queries, previously considered active, are already terminated.
Also, total runtime will potentially be `#queries x timeout` in the worst case.

Check for active queries each time seems a better option to me. What do you think?

core/cloudflow-spark-testkit/src/main/scala/cloudflow/spark/testkit/TestSparkStreamletContext.scala
maasg: The issue I'm observing with the failing test is that there's no data produced for the first query. Looks like the `rate` producer generates the first batch with zero records :-(

maasg:
@debasishg @skonto @RayRoestenburg I found a way to make this work. Could you have a look?

core/cloudflow-spark-testkit/src/main/scala/cloudflow/spark/testkit/TestSparkStreamletContext.scala
skonto: processAllAvailable waits until you get all the data.

core/cloudflow-spark-testkit/src/main/scala/cloudflow/spark/testkit/SparkStreamletTestkit.scala
skonto: With two streams both should finish fat, if not the test is not tuned anyway. It depends.

core/cloudflow-spark-testkit/src/main/scala/cloudflow/spark/testkit/TestSparkStreamletContext.scala
```diff
@@ -44,7 +45,7 @@ private[testkit] class TestSparkStreamletContext(override val streamletRef: Stri
                                                  override val config: Config = ConfigFactory.empty)
     extends SparkStreamletContext(StreamletDefinition("appId", "appVersion", streamletRef, "streamletClass", List(), List(), config),
                                   session) {
-
+  val ProcessingTimeInterval = 1.second
```
skonto: This is hardcoded that guesses that the data will be processed in 1 sec? I suspect that this is a bit fragile.

maasg: No. This is the `Trigger` interval, not total processing time. I have observed that fixing the micro-batch period provides improved reliability in observing data being processed. Some tests had like a 40 seconds 'wait time' that can now be reduced to 2 seconds. (given that the fixed 1-second 'micro-batch' ensures that the data would have been processed already)

RayRoestenburg: I agree with @stavros, we should make it possible to change this.

maasg: @RayRoestenburg I was typing the above msg before your comment. Does that answer your question as well?

RayRoestenburg: Thanks, yes.

maasg: @skonto do you agree that this can stay as-is?

PR #158, opened 2020-02-28
Fix for regexp check in ConfigParameters of type String

olofwalker:


olofwalker:
@debasishg 

```
[info] Successfully built the following Cloudflow application image:
[info]  
[info]   docker.io/olofwalker/sensor-data-scala:140-9775a75
[info]  ```

PR #162, opened 2020-03-03
Support scalafmt based formatting at compile time

skonto:
When user runs `sbt compile` code should be formatted.
Some properties might be inherited due to parent-child relationships defined but here we make all explicit. Also note that at the root project we only do an aggregation no project depends on it.
Contributors should be able to get their code formatted automatically at compile time.

No.

Manually by checking out the code modifying a file at random project and testing if formatting will change it back to its correct form.

### DIFF
--- a/core/build.sbt
+++ b/core/build.sbt
@@ -45,8 +45,9 @@ lazy val root =
 
 lazy val streamlets =
   cloudflowModule("cloudflow-streamlets")
-    .enablePlugins(GenJavadocPlugin)
+    .enablePlugins(GenJavadocPlugin, ScalafmtPlugin)
     .settings(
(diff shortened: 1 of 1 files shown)


olofwalker:
@skonto ScalaFmt discourages the use of this since it will interfere with editor undo buffers and add to the overall compile time. 

`⚠️ This option is discouraged since it messes up undo buffers in the editor and it slows down compilation. It is recommended to use "format on save" in the editor instead.`

RayRoestenburg:
@olofwalker AFAIK we have all agreed that we should turn this on, as @skonto has done.

skonto:
merging thank you both!

maasg:
+1

Issue #151, opened 2020-02-26
Flink consumer group id is not unique enough


agemooij:
**Describe the bug**
Quoted from a Gitter conversation:

> V.Zaytsev @unit7
> Hi everyone! Can anyone suggest me why for Flink-streamlets kafka consumer-group-id is specified as the name of the input topic? In fact, this means that only one Flink-streamlet can read from a particular topic, that is all Flink-streamlets will be combined in one consumer group. So what was the design idea? For akka-streamlets, consumer-group-id is formed taking into account the streamlet and inlet from which the reading occurs.
> Thanks in advance.
> The related sources:
> - https://github.com/lightbend/cloudflow/blob/master/core/cloudflow-flink/src/main/scala/cloudflow/flink/FlinkStreamletContextImpl.scala#L49
> - https://github.com/lightbend/cloudflow/blob/master/core/cloudflow-akka/src/main/scala/cloudflow/akkastream/AkkaStreamletContextImpl.scala#L77
>
>
>Debasish Ghosh @debasishg 11
>@unit7-0 We can surely align the consumer group id to make it similar to akka. But Flink Kafka connectors don’t use the consumer group management functionality because they use lower-level APIs on each parallel instance for more control on individual partition consumption. So, essentially, the “group.id” setting in the Flink Kafka connector is only used for committing offsets back to ZK / Kafka brokers.
>
>
> V.Zaytsev @unit7
> @debasishg, Did I understand correctly that since each flink-streaming-job tracks the offsets individually and saves it by the internal checkpoint mechanism, is there no need to specify a different groupd.id for each streamlet? Accordingly in the case when there are two streamlets reading the same topic but has different business logic will work correctly although the consumers will be in the same consumer-group?

I think they are right and we should adjust our consumer group id to be the same as the one we use for the Akka streamlet runner.

debasishg:
Fixed with https://github.com/lightbend/cloudflow/pull/174

PR #174, opened 2020-03-06
Kafka consumer group id in Flink Streamlet now aligned with Akka Streamlet - includes appId, streamletRef and inlet name

debasishg:
Github Issue Id: https://github.com/lightbend/cloudflow/issues/151

* Kafka consumer group id in flink streamlet processing now includes `appId`, `streamletRef` and `inletName` which is aligned with what Akka streamlet does

* With the existing logic only 1 streamlet will be able to read from a Kafka topic since it has the topic name as the consumer group

* Not really

* Tested manually on a GKE cluster with the taxi-ride application

### DIFF
--- a/core/cloudflow-flink/src/main/scala/cloudflow/flink/FlinkStreamletContextImpl.scala
+++ b/core/cloudflow-flink/src/main/scala/cloudflow/flink/FlinkStreamletContextImpl.scala
@@ -24,7 +24,6 @@ import org.apache.flink.streaming.connectors.kafka._
 import com.typesafe.config._
 import cloudflow.streamlets._
 import java.{ util ⇒ ju }
-import java.util.UUID.randomUUID
 
 /**
  * An implementation of `FlinkStreamletContext`
@@ -43,11 +42,13 @@ class FlinkStreamletContextImpl(
    * @return the data read as `DataStream[In]`
    */
   override def readStream[In: TypeInformation](inlet: CodecInlet[In]): DataStream[In] = {
-    val srcTopic = resolvePort(inlet.name)
+    val savepointPath = findSavepointPathForPort(inlet)
+    val srcTopic      = savepointPath.value
+    val groupId       = savepointPath.groupId(inlet)
 
     val properties = new ju.Properties
     properties.setProperty("bootstrap.servers", config.getString("cloudflow.kafka.bootstrap-servers"))
-    properties.setProperty("group.id", groupId(inlet))
+    properties.setProperty("group.id", groupId)
 
     val consumer = new FlinkKafkaConsumer[In](
       srcTopic,
@@ -74,7 +75,8 @@ class FlinkStreamletContextImpl(
    */
   override def writeStream[Out: TypeInformation](outlet: CodecOutlet[Out], stream: DataStream[Out]): DataStreamSink[Out] = {
 
-    val destTopic = resolvePort(outlet.name)
+    val savepointPath = findSavepointPathForPort(outlet)
+    val destTopic     = savepointPath.value
 
     val properties = new ju.Properties
     properties.setProperty("bootstrap.servers", config.getString("cloudflow.kafka.bootstrap-servers"))
@@ -88,11 +90,5 @@ class FlinkStreamletContextImpl(
         FlinkKafkaProducer.Semantic.AT_LEAST_ONCE
       )
     )
-  }
-
-  private def groupId[T](inlet: CodecInlet[T]) = {
-    val base = s"${streamletDefinition.appId}.${streamletDefinition.streamletRef}.${inlet.name}"
-    if (inlet.hasUniqueGroupId) base + randomUUID.toString
-    else base
   }
 }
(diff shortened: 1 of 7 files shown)


core/cloudflow-flink/src/main/scala/cloudflow/flink/FlinkStreamletContextImpl.scala
```diff
@@ -46,7 +47,7 @@ class FlinkStreamletContextImpl(
 
     val properties = new ju.Properties
     properties.setProperty("bootstrap.servers", config.getString("cloudflow.kafka.bootstrap-servers"))
-    properties.setProperty("group.id", srcTopic)
+    properties.setProperty("group.id", groupId(inlet))
```
RayRoestenburg: @debasishg srcTopic, I think resolves to `SavepointPath.value`, which is  s"${appId}.${streamletRef}.${portName}".
`SavepointPath` encodes this. Maybe we can reuse that? and also fix this in the `AkkaStreamletContextImpl`? (I think the groupId function there takes a streamletRef and inlet that is not necessary)
So what I mean is move the code to create groupId to SavepointPath, since this must encode what is needed for connecting to the savepoint (the topic).

RayRoestenburg: maybe resolvePort can also be moved to StreamletContext so it works for all streamlets.

examples/taxi-ride/project/cloudflow-plugins.sbt
```diff
@@ -3,4 +3,4 @@
 resolvers += "Akka Snapshots" at "https://repo.akka.io/snapshots/"
 resolvers += Resolver.url("cloudflow", url("https://lightbend.bintray.com/cloudflow"))(Resolver.ivyStylePatterns)
 
-addSbtPlugin("com.lightbend.cloudflow" % "sbt-cloudflow" % "1.3.2-pre-153-49e280a")
+addSbtPlugin("com.lightbend.cloudflow" % "sbt-cloudflow" % "1.3.2-SNAPSHOT")
```
debasishg: will change this to an internal release when merging into master

core/cloudflow-flink/src/main/scala/cloudflow/flink/FlinkStreamletContextImpl.scala
```diff
@@ -88,4 +89,10 @@ class FlinkStreamletContextImpl(
       )
     )
   }
+
+  private def groupId[T](inlet: CodecInlet[T]) = {
```
RayRoestenburg: This can likely also move to `StreamletContext`. (and use `SavepointPath` to get the right value)

core/cloudflow-flink/src/main/scala/cloudflow/flink/FlinkStreamletContextImpl.scala
debasishg: @ray I like this that u did in Akka ..

```scala
private def findSavepointPathForPort(port: StreamletPort): SavepointPath =
  streamletDefinition
    .resolveSavepoint(port)
    .getOrElse(throw PortNotFoundException(port, this))
```

better than `resolvePort` returning an opaque string. And the following is much more meaningful ..

```scala
val savepointPath = findSavepointPathForPort(inlet)
val topic         = savepointPath.value
```

I will remove `resolvePort` and move `findSavepointPathForPort` to `StreamletContext` so that it can be used as above from all runtimes.

Agree?

RayRoestenburg: yes

debasishg:
@RayRoestenburg Done with all the changes .. Please take a look ..

core/cloudflow-streamlets/src/main/scala/cloudflow/streamlets/StreamletContext.scala
```diff
@@ -36,6 +36,18 @@ trait StreamletContext {
    */
   def streamletRef: String = streamletDefinition.streamletRef
 
+  /**
+   * Get the savepoint path (topic name) from the port
+   *
+   * @param port the StreamletPort
+   * @return the savepoint path
+   * @throws PortNotFoundException if there is not mapping
```
RayRoestenburg:
```suggestion
   * @throws PortNotFoundException if there is no mapping found
```

core/cloudflow-streamlets/src/main/scala/cloudflow/streamlets/StreamletContext.scala
```diff
@@ -80,3 +92,6 @@ trait StreamletContext {
       extends Exception(s"Mount path for Volume Mount named [${volumeMount.name}] is unavailable.")
 
 }
+
+case class PortNotFoundException(port: StreamletPort, context: StreamletContext)
```
RayRoestenburg:
```suggestion
case class PortNotFoundException(port: StreamletPort, streamletDefinition: StreamletDefinition)
```
Might be better.

core/cloudflow-streamlets/src/main/scala/cloudflow/streamlets/StreamletDefinition.scala
```diff
@@ -49,6 +50,11 @@ case class StreamletDefinition(appId: String,
  */
 final case class SavepointPath(appId: String, streamletRef: String, portName: String) {
   def value: String = s"${appId}.${streamletRef}.${portName}"
+  def groupId[T](inlet: CodecInlet[T]) = {
```
RayRoestenburg: 👍

debasishg:
@RayRoestenburg done ..

PR #176, opened 2020-03-09
Enable back ownerReferences on secrets

skyluc:
But only create the secrets once, after the CR has been created.

On deployment, secrets are created once, after the CR creation, with the full `ownerReferences` data.

Making sure to have `ownerReference` data on all resources we create.

No

Building new version of `kubectl-cloudflow`, deploying taxi-ride example with it.

### DIFF
--- a/kubectl-cloudflow/cmd/deploy.go
+++ b/kubectl-cloudflow/cmd/deploy.go
@@ -162,8 +162,9 @@ func (opts *deployOptions) deployImpl(cmd *cobra.Command, args []string) {
 		util.LogAndExit("%s", err)
 	}
 
+	// Delay the creation of the secret for after the ownerReferences has been added.
+	// Creating then updating the secret generates problem for the Flink streamlets deployment.
 	streamletNameSecretMap := deploy.CreateSecretsData(&applicationSpec, configurationKeyValues)
-	createOrUpdateStreamletSecrets(k8sClient, namespace, streamletNameSecretMap)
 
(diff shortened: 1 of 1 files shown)


kubectl-cloudflow/cmd/deploy.go
```diff
@@ -180,12 +181,8 @@ func (opts *deployOptions) deployImpl(cmd *cobra.Command, args []string) {
...
 	ownerReference = storedCR.GenerateOwnerReference()
 
-  // TODO: Temporarily disabled updating ownerReferences on secret
-  // It currently breaks Flink streamlets with config parameters.
-  // Disableling it so Cloudflow works while investigating of the problem.
-
-	//streamletNameSecretMap = deploy.UpdateSecretsWithOwnerReference(ownerReference, streamletNameSecretMap)
-	//createOrUpdateStreamletSecrets(k8sClient, namespace, streamletNameSecretMap)
+	streamletNameSecretMap = deploy.UpdateSecretsWithOwnerReference(ownerReference, streamletNameSecretMap)
+	createOrUpdateStreamletSecrets(k8sClient, namespace, streamletNameSecretMap)
```
olofwalker: @skyluc Have you tested what happens if you re-configure the application? That part updates the ownerreferences also

skyluc: Hmm. No. Didn't check that.

skyluc: Tested reconfiguring a Flink streamlet. It worked fine, data was processed after restart of the pods.
Looking at the code, we only create/update the secret once. So if the problem at deploy is linked to timing, it does affect this.

PR #179, opened 2020-03-10
Put the snippets used in doc in their own folder

skyluc:
Extract the code snippets used in the documentation to the `example/snippets` folder.
Some of the code snippets were already taken from the projects in `examples`, other were hardcoded in the documentation.
The snippets are now in projects in the `snippets` folder. They are compiled and tested with the other example projects.

This only part of the code snippets to extract from the doc, but the structure change in the documentation is needed for other reasons.

The code snippets in the documentation need to be compiled and check, to make sure that they are always compatible with the core project.

No

Running the documentation build, with equivalent PR, checking the result.


skyluc:
related cloudflow-docs PR lightbend/cloudflow-docs#40

examples/examples.yaml
```diff
@@ -21,3 +21,35 @@ examples:
   - id: "tensorflow-akka"
     path: "tensorflow-akka"
     description: This app shows how to serve a Tensorflow model using Akka
+  - id: "sensor-data-scala-snippets"
```
agemooij: @skyluc we might have to re-assess the usage of this file. It had a long-forgotten purpose at some point but it would be interesting to see whether we still need a manually updated list of example projects.

examples/examples.yaml
```diff
@@ -21,3 +21,35 @@ examples:
...
+    description: |
+      Set of files containing the Java Akka streams code snippets used in the documentation
+  - id: "spark-scala-snippets"
+    path: "snippets/modules/ROOT/examples/spark-scala"
+    description: |
+      Set of files containing the Scala Spark code snippets used in the documentation
+  - id: "build-akka-streamlets-java-snippets"
+    path: "snippets/modules/ROOT/examples/build-akka-streamlets-java"
+    description: |
+      Set of files containing the Scala code snippets used in the build-akka-streamlets documentation
```
agemooij: Is there a good reason to keep this separate from `akkastreams-java` (and the same for the scala side)? Whot not have less top-level containers and (if/when necessary) just use folders to structure the contents of such projects.

I had expected one single snippets project or perhaps, if that made more sense, one per runtime. I'm not sure this additional directory structure serves a good purpose.

examples/snippets/antora.yml
```diff
@@ -0,0 +1,3 @@
+name: "docsnippets"
+version: 1.3.1
```
agemooij: Do we need the version hardcoded here?
Every additional place that hardcodes our version number is one more problem to solve for automated release management.

examples/examples.yaml
skyluc: It is used in `scripts/build-all.sh`, which is used in the PR check to build and test all example projects.
It is really practical, I only had to add the new projects to the list, and they became part of the CI.

examples/snippets/antora.yml
skyluc: Yes, we will need a specific version number, to support multi-version documentation
On master, it can be `latest` on something like that. But in the branches, it should be the actual version.

agemooij: How can we keep this maintainable? 
Should this be auto-updated by the release automation?

examples/examples.yaml
agemooij: Wouldn't it work just as well to just list the number of sub-dirs or `find` all sbt projects in the structure?

examples/examples.yaml
agemooij: Every extra sbt project we add here requires us to updates its dependencies, etc. I'd prefer the smallest possible number of snippet projects and then use internal directories to add structure where needed. If the Java vs Scala difference is tricky then that could be a necessary reason for having more than 1 snippets (e.g. 2) projects.

skyluc: One of the reasons to have different project is that most of them define a blueprint
And I remember that we had problems with sbt projects defining multiple Cloudflow app subproject, and we include blueprint in the documentation, which have to be verified.
So I went the easy way, one project per blueprint related bloc of snippets.

Also, having one, or a couple of, big sbt project will only be moving the complexity.
The `build-*-streamlets-*` projects already contain each 4 library projects, one for each steps of the explanation, and one app project for the blueprint.

But I should be able to group some of the simple projects.

agemooij: ok, thanks. Just trying to keep a handle on the complexity 😉

examples/snippets/modules/ROOT/examples/akkastreams-java/project/cloudflow-plugins.sbt
```diff
@@ -0,0 +1,5 @@
+// Resolver for the cloudflow-sbt plugin
+//
+resolvers += Resolver.url("cloudflow", url("https://lightbend.bintray.com/cloudflow"))(Resolver.ivyStylePatterns)
+
+addSbtPlugin("com.lightbend.cloudflow" % "sbt-cloudflow" % "1.3.1-RC4")
```
RayRoestenburg: This should be 1.3.1?

examples/snippets/modules/ROOT/examples/akkastreams-java/src/main/java/cloudflow/akkastreamsdoc/FilterStreamlet.java
```diff
@@ -0,0 +1,129 @@
...
+import akka.stream.javadsl.*;
+import akka.util.ByteString;
+import com.typesafe.config.Config;
+import cloudflow.akkastream.*;
+import cloudflow.akkastream.javadsl.RunnableGraphStreamletLogic;
+import cloudflow.streamlets.*;
+import cloudflow.streamlets.avro.AvroInlet;
+import cloudflow.streamlets.avro.AvroOutlet;
+
+import java.nio.file.Path;
```
RayRoestenburg: [minor] we usually put the java imports first, then scala, then ...

examples/snippets/modules/ROOT/examples/akkastreams-scala/build.sbt
```diff
@@ -0,0 +1,42 @@
...
+
+lazy val sensorData =  (project in file("."))
+    .enablePlugins(CloudflowAkkaStreamsApplicationPlugin)
+    .settings(
+//end::docs-projectSetup-example[]
+      libraryDependencies ++= Seq(
+        "com.lightbend.akka"     %% "akka-stream-alpakka-file"  % "1.1.2",
+        "com.typesafe.akka"      %% "akka-http-spray-json"      % "10.1.10",
+        "ch.qos.logback"         %  "logback-classic"           % "1.2.3",
+        "com.typesafe.akka"      %% "akka-http-testkit"         % "10.1.10" % "test",
```
RayRoestenburg: Have you checked, for instance by running these examples, that it does not start with many versions of Akka libraries on classpath? We depend on `val AkkaHttp     = "10.1.9"` in cloudflow itself. Probably best to keep that the same.

examples/snippets/modules/ROOT/examples/akkastreams-scala/build.sbt
```diff
@@ -0,0 +1,42 @@
+//tag::docs-projectSetup-example[]
+import sbt._
+import sbt.Keys._
+
+lazy val sensorData =  (project in file("."))
+    .enablePlugins(CloudflowAkkaStreamsApplicationPlugin)
+    .settings(
+//end::docs-projectSetup-example[]
+      libraryDependencies ++= Seq(
+        "com.lightbend.akka"     %% "akka-stream-alpakka-file"  % "1.1.2",
```
RayRoestenburg: This might depend on a different Akka version than `2.5.29`

examples/snippets/modules/ROOT/examples/akkastreams-scala/project/cloudflow-plugins.sbt
```diff
@@ -0,0 +1,5 @@
+// Resolver for the cloudflow-sbt plugin
+//
+resolvers += Resolver.url("cloudflow", url("https://lightbend.bintray.com/cloudflow"))(Resolver.ivyStylePatterns)
+
+addSbtPlugin("com.lightbend.cloudflow" % "sbt-cloudflow" % "1.3.1-RC4")
```
RayRoestenburg:
```suggestion
addSbtPlugin("com.lightbend.cloudflow" % "sbt-cloudflow" % "1.3.1")
```

examples/snippets/modules/ROOT/examples/akkastreams-scala/src/main/scala/cloudflow/akkastreamsdoc/DataHttpIngress.scala
```diff
@@ -0,0 +1,18 @@
...
+import cloudflow.akkastream.util.scaladsl._
+
+import cloudflow.streamlets.avro._
+import cloudflow.streamlets._
+
+import akka.http.scaladsl.marshallers.sprayjson.SprayJsonSupport._
+import JsonSupport._
+
+class DataHttpIngress extends AkkaServerStreamlet {
+  val out         = AvroOutlet[Data]("out").withPartitioner(RoundRobinPartitioner)
```
RayRoestenburg: very minor, you can leave the `.withPartitioner(RoundRobinPartitioner)` out cause it is default. (maybe for didactic purposes it's better to be explicit so not sure)

examples/snippets/modules/ROOT/examples/build-akka-streamlets-java/step0/src/main/java/com/example/ReportPrinter.java
```diff
@@ -0,0 +1,18 @@
...
+import cloudflow.akkastream.*;
+import cloudflow.akkastream.javadsl.*;
+
+public class ReportPrinter extends AkkaStreamlet {
+  // 1. TODO Create inlets and outlets
+  // 2. TODO Define the shape of the streamlet
+  public StreamletShape shape() { throw new UnsupportedOperationException("Not Implemented"); }
+  // 3. TODO Override createLogic to provide StreamletLogic
+  public RunnableGraphStreamletLogic createLogic() { throw new UnsupportedOperationException("Not Implemented"); }
+}
```
RayRoestenburg: !WHY LUC WHY!
https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap03.html#tag_03_206
lol I'm just kidding of course.

examples/snippets/modules/ROOT/examples/sensor-data-scala/src/main/scala/sensordata/ConfigParams.scala
```diff
@@ -0,0 +1,36 @@
...
+trait ConfigParams {
+  val LogLevel = RegExpConfigParameter(
+    "log-level",
+    "Provide one of the following log levels, debug, info, warning or error",
+    "^debug|info|warning|error$",
+    Some("debug")
+  )
+
+  /*
+  val InfluxDBActive = BooleanConfigParameter(
```
RayRoestenburg: Remove commented code please.

examples/snippets/modules/ROOT/examples/sensor-data-scala/src/main/scala/sensordata/ConfigParams.scala
```diff
@@ -0,0 +1,36 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package sensordata
+
+import cloudflow.streamlets._
+
+trait ConfigParams {
```
RayRoestenburg: Its not clear whose configparams these are.. I would prefer to define these where they are used.

RayRoestenburg: Oh it also looks like (unless there is a 'too big diff' in here) that it is not used?

examples/snippets/modules/ROOT/examples/sensor-data-scala/.gitignore
```diff
@@ -0,0 +1,30 @@
+etc/bootstrap-local.conf
```
RayRoestenburg: Why is the entire sensor-data-scala copied in here @skyluc ? Can we avoid it?

examples/snippets/modules/ROOT/examples/build-akka-streamlets-scala/build.sbt
```diff
@@ -0,0 +1,75 @@
+lazy val root =
```
RayRoestenburg: The build akka streamlets example looks really nice, good to have it added!

examples/snippets/modules/ROOT/examples/sensor-data-scala/.gitignore
agemooij: As discussed offline, the idea is that eventually we get rid of the standalone sensor-data-xxx examples in favor of these snippets. The sensor data examples were always meant as getting started info more than full examples.

agemooij:
> I'm not sure what exactly was discussed but would be great not to duplicate sensor-data-scala code. What is the plan?

As discussed offline, the idea is that eventually we get rid of the standalone sensor-data-xxx examples in favor of these snippets. The sensor data examples were always meant as getting started info more than full examples.

skyluc:
@RayRoestenburg the goal of this PR is to start transferring the base files used for code snippets in the documentation, to a location where Antora can easily access then, and they can also be shared with the enterprise documentation.
I didn't not create any new documentation, and did not fix any of the code snippets, except to make sure that they actually compile.

As for the copy of `sensor-data-scala`, I would have loved to do without it, but it is use as the base of the `develop-example-streamlets` documentation page.

@agemooij if `sensor-data-scala` is not meant to stay in the 'normal' examples, I think we should move it in the snippets section now. It would be a cleaner transition.

agemooij:
> @agemooij if `sensor-data-scala` is not meant to stay in the 'normal' examples, I think we should move it in the snippets section now. It would be a cleaner transition.

Agreed. Feel free to do that as part of this PR. Both existing `sensor-data-...` example projects can be removed AFAIC

skyluc:
Moved `sensor-data-scala`, and remove `sensor-data-java`.
Updated the version numbers for the cloudflow plugin.
Set a generic version for the snippets (until we correctly document how version will work).

I will likely merge this PR tomorrow morning, so we can also merge the `cloudflow-docs` one.

skyluc:
Merging this PR, to be able to merge the `cloudflow-docs` one.
I'll go through the last comments in the next PR.

PR #182, opened 2020-03-11
1.3.1 Release Notes

agemooij:


release-notes/cloudflow-1.3.1.md
```diff
@@ -0,0 +1,42 @@
+# Cloudflow 1.3.1 Release Notes
+
+Cloudflow 1.3.1 is a minor patch release and contains the following changes:
```
olofwalker:
```suggestion
Cloudflow 1.3.1 is a patch release and contains the following changes:
```

olofwalker: This is not a minor release IMO, we upgrade a lot of components, as a customer I would be very interested in why and what this means.

release-notes/cloudflow-1.3.1.md
```diff
@@ -0,0 +1,42 @@
...
+
+**Cloudflow 1.3.1 was tested on the following Kubernetes distributions/versions:**
+
+- Google Kubernetes Engine (GKE) using Kubernetes 1.13, 1.15, and 1.16
+- Amazon Elastic Kubernetes Service (EKS) using Kubernetes 1.14
+- Azure Kubernetes Service (AKS) using Kubernetes 1.13 (Enterprise edition only)
+- Openshift 3.11 using Kubernetes 1.11 (Enterprise edition only)
+- Openshift 4.3 using Kubernetes 1.16 (Enterprise edition only)
+
+**NOTE**: We are planning on removing the dual-installer architecture in Cloudflow 1.4/1.5, which will enable OSS installation and upgrade support for all platforms currently supported only by the Enterprise installer.
```
olofwalker: Maybe move this down to the roadmap?

agemooij: @olofwalker I had it there before but I moved it here because the above text makes very clear to the reader that there are two installers and that the Enterprise installer supports a lot more distros. I thought it would be good to put the note there to hook on that realization instead of as a more anonymous item on the roadmap. And my second reason for moving it here is that it is not entirely clear at the moment whether this will fit into our 1.4 timeline and I didn't want to add a "1.5 and beyond" section since that is still quite up in the air.

release-notes/cloudflow-1.3.1.md
```diff
@@ -0,0 +1,42 @@
+# Cloudflow 1.3.1 Release Notes
+
+Cloudflow 1.3.1 is a patch release and contains the following changes:
+
```
debasishg: should we call it a patch release with so many version upgrades ?

release-notes/cloudflow-1.3.1.md
```diff
@@ -0,0 +1,42 @@
...
+- Upgraded to include Alpakka Kafka 2.0.2, which includes significant speed ups. Also updated the Akka streams Streamlet APIs to match.
+- Streamlet HTTP endpoints now get predictable port numbers instead of randomized ones
+- Support for fully uninstalling Cloudflow including all associated resources, dependencies and deployed applications
+- The CLI now supports Cloudflow installations in non-default namespaces by auto-detecting the installation namespace
+- The OSS installer now supports Amazon Elastic Kubernetes Service (EKS)
+
+**Known issues with this release:**
+
+- The Enterprise edition cannot be installed on K8s version 1.16 or higher due to a compatibility issue with the Lightbend Console.
+- The `kubectl cloudflow status` command does not report a correct rollup status for the entire application when it includes Flink streamlets; e.g. it reports `Pending` even though all individual streamlets are `Running`.
```
debasishg: May be we can mention that this does not affect the running of the application in any way

release-notes/cloudflow-1.3.1.md
```diff
@@ -0,0 +1,42 @@
...
+Cloudflow 1.3.1 can be installed on a Kubernetes cluster in the usual way using either the OSS or Enterprise installers. Enterprise installations can be upgraded in-place by running the 1.3.1 installation bootstrap script (and choosing the same storage classes).
+
+**The Cloudflow 1.3.1 CLI can be downloaded using one of the following links:**
+
+- Linux:
+- MacOS:
+- Windows:
+
+**The roadmap for the Cloudflow 1.4 release, scheduled for April/May 2020, currently looks like this:**
+
```
debasishg: verify and deploy blueprint through CLI ?

release-notes/cloudflow-1.3.1.md
agemooij: It's not a major release and it increases the patch level. 
What other term would you suggest?

To me these are just dependency updates instead of major changes. But perhaps I'm underestimating the impact?

release-notes/cloudflow-1.3.1.md
agemooij: Good point

release-notes/cloudflow-1.3.1.md
agemooij: @debasishg That is part of the multiple image support epic. I didn't want to dive deeper than the epic level in this brief roadmap list.

PR #188, opened 2020-03-13
Csp 1919 all in one app

maasg:
This is a test application that exercises all the current runtimes of cloudflow and produces output for each system independently. 
At the blueprint level it uses multiple instances of the same streamlet, and multicast to several streamlets. 
This will serve as basis to add more features we would like to test, like configuration parameters, scaling defaults, etc..

For the moment, the intention is to have a baseline that we can use to automate a test. 
Is there something missing to achieve that first goal? 
When answering that question, please refrain from feature creep. 
I'd like to have a minimal end-to-end test working before we add all bells and whistles.

I'm thinking that we can get fair initial coverage with some `kubectl cloudflow` scripting.
An example scenario would be:
```
kctl cloudflow deploy  swiss-knife
pods <- kctl cloudflow status // check that deployment succeeded, capture pods
foreach (pod <- pods.filter(is_egress)) 
   log <- kubectl log pod
   check_is_working(log)
// profit!
```


RayRoestenburg:
This should not be in the examples direcory IMHO. can you add it under something related to testing?

maasg:
@RayRoestenburg this is in `/examples/testing` as discussed with @skyluc to follow the `/examples/snippets` pattern. But we can add it to `/ites` or `/intergration-test`
Not sure if any of them is better than the others. I don't see a clear winner there, so let's use democracy :-)

examples/testing/swiss-knife/src/main/avro/agg.avsc
```diff
@@ -0,0 +1,20 @@
+{
+    "namespace": "sensors",
```
RayRoestenburg: should this be swissknife?

examples/testing/swiss-knife/src/main/scala/swissknife/AkkaTransformation.scala
```diff
@@ -0,0 +1,38 @@
...
+  val in    = AvroInlet[Data]("in")
+  val out   = AvroOutlet[Data]("out").withPartitioner(RoundRobinPartitioner)
+  val shape = StreamletShape(in).withOutlets(out)
+
+  override def createLogic = new RunnableGraphStreamletLogic() {
+    def runnableGraph = sourceWithOffsetContext(in).via(flow).to(committableSink(out))
+    def flow =
+      FlowWithCommittableContext[Data]
+        .map { data ⇒
+          data.copy(src = data.src + "-akka")
```
RayRoestenburg:
```suggestion
          data.copy(src = s"${data.src}-akka")
```
Super minor of course.

examples/testing/swiss-knife/src/main/scala/swissknife/FlinkCounter.scala
```diff
@@ -0,0 +1,27 @@
...
+class FlinkCounter extends FlinkStreamlet {
+
+  @transient val in  = AvroInlet[Data]("in")
+  @transient val out = AvroOutlet[Data]("out", _.src)
+
+  @transient val shape = StreamletShape.withInlets(in).withOutlets(out)
+
+  override def createLogic() = new FlinkStreamletLogic {
+    override def buildExecutionGraph = {
+      val stream: DataStream[Data] =
```
RayRoestenburg: I remember that @skyluc had a very simple flink streamlet, that did not catch the issue that we saw, "because it was too simple". This might be the case here? (I'm not sure but might be interesting to chat about it)

examples/testing/swiss-knife/src/main/scala/swissknife/SparkRandomGenDataIngress.scala
```diff
@@ -0,0 +1,55 @@
...
+import cloudflow.spark.{ SparkStreamlet, SparkStreamletLogic }
+import org.apache.spark.sql.Dataset
+import org.apache.spark.sql.streaming.OutputMode
+import org.apache.spark.sql.functions._
+
+import cloudflow.spark.sql.SQLImplicits._
+
+case class Rate(timestamp: Timestamp, value: Long)
+
+class SparkRandomGenDataIngress extends SparkStreamlet {
```
RayRoestenburg: Should we not try to generate deterministically?

RayRoestenburg:
@maasg yes I think a root `integration-test` folder in the project would be better. This is not an example, and we're not going to use it for snippets, or document it in that way.

RayRoestenburg:
Do you have an idea how the `check_is_working(log)` will work? 
What do you think of adding http endpoints to (some of the) streamlets and just fire off requests to these, to check internal state?

maasg:
@RayRoestenburg 
Using the @viktorklang's conjecture, I have been checking the logs manually using `kubectl logs ...`. The output has an easy pattern that we can check. I think it will be easy to automate.

If we add networked endpoints, we will have to do some `kubectl port-forward` tricks to access them. It might be harder than the logs.

examples/testing/swiss-knife/README.md
```diff
@@ -0,0 +1,3 @@
+# Swiss Army App
```
RayRoestenburg: Swiss Army Knife App?

;-)

maasg:
@RayRoestenburg thanks for finding my copy/paste left-overs. I have cleaned them now :-)

maasg:
@RayRoestenburg I thought some more about it and I think you are right. This is not "a sample app that we (may) use for testing"  but a "specific app that's part of our integration testing suite". 
Moved.

RayRoestenburg:
Thanks @gerard! BTW the app looks good as a starting point!

examples/testing/swiss-knife/src/main/scala/swissknife/SparkRandomGenDataIngress.scala
agemooij: I would indeed prefer deterministic behavior because it is much easier to verify the actual output with the expected output

integration-tests/swiss-knife/project/cloudflow-plugins.sbt
```diff
@@ -0,0 +1,7 @@
+// Resolver for the cloudflow-sbt plugin
+//
+resolvers += "Akka Snapshots" at "https://repo.akka.io/snapshots/"
+resolvers += Resolver.url("cloudflow", url("https://lightbend.bintray.com/cloudflow"))(Resolver.ivyStylePatterns)
+
+addSbtPlugin("com.lightbend.cloudflow" % "sbt-cloudflow" % "1.3.2-SNAPSHOT")
```
agemooij: We will need to make sure this gets updated as part of the release automation tooling. This happens automatically for (known) examples but this one breaks that pattern and will have to be added manually.

integration-tests/swiss-knife/src/main/blueprint/blueprint.conf
```diff
@@ -0,0 +1,18 @@
+blueprint {
+  streamlets {
+    ingress = swissknife.SparkDataGenerator
```
agemooij: To limit the deployment size of this app a bit we cold change this to an Akka-based generator. that would safe at least 3 pods.

integration-tests/swiss-knife/src/main/scala/README.md
```diff
@@ -0,0 +1 @@
+# This is an example project that illustrates the use of the Spark Support in Cloudflow
```
agemooij: This file can be removed since it is just a leftover from a template. It also contains incorrect data.

integration-tests/swiss-knife/src/main/scala/swissknife/FlinkCounter.scala
```diff
@@ -0,0 +1,42 @@
...
+ */
+package swissknife
+
+import cloudflow.flink.FlinkStreamlet
+import org.apache.flink.streaming.api.scala._
+import cloudflow.streamlets.StreamletShape
+import cloudflow.streamlets.avro._
+import cloudflow.flink._
+
+class FlinkCounter extends FlinkStreamlet {
```
agemooij: It is called `FlinkCounter` but it performs no counting but instead does the same transformation as the Akka "transformation" streamlet.

I suggest we use consistent naming for all three transformation streamlets

integration-tests/swiss-knife/src/main/scala/swissknife/SparkCounter.scala
```diff
@@ -0,0 +1,54 @@
...
+      writeStream(outStream, out, OutputMode.Append).toQueryExecution
+    }
+
+    private def process(inDataset: Dataset[Data]): Dataset[Data] = {
+      val query = inDataset
+        .withColumn("ts", $"timestamp".cast(TimestampType))
+        .withColumn("updated_src", concat($"src", lit("-spark")))
+        .withWatermark("ts", "0 seconds")
+        .groupBy(window($"ts", "5 seconds"), $"updated_src")
+        .agg(count($"src").as("count"))
```
agemooij: Is there any real reason for the spark streamlet to do more custom work then the Akka and Flink ones?
I suggest we stick with al streamlets doing the same thing.

maasg: We probably need to implement something similar with Flink rather than simplifying this. 
In the case of Spark, having a stateful operator exercises the state store and PVC access, which have been a source of problems.

agemooij: OK, makes sense. But let's make sure we don't add non-essential details just to keep the signal to noise level high.

examples/testing/swiss-knife/src/main/scala/swissknife/FlinkCounter.scala
maasg: @skyluc what was the issue that Ray refers to?

examples/testing/swiss-knife/src/main/scala/swissknife/SparkRandomGenDataIngress.scala
maasg: It produces a counter now.

PR #190, opened 2020-03-17
CLI based deployment of applications through blueprint

debasishg:

This PR does the following:
* implements deployment of cloudflow applications through blueprint
* supports both blueprint and image based deployment using flags `--blueprint` and `--image` respectively
* supports passing of initial scale per streamlet using flag `--scale` e.g. `--scale cdr-aggregator=3,cdr-generator1=3`
* migrates unit tests of `deploy` and part of `cmd` to Ginkgo and Gomega
* removes `AppVersion` from go and scala code (as per earlier discussion)
* changes `Deployment.Replicas` to `*int` to support dynamic setting and unsetting as required (as per discussion)

* To support deployment of blueprint (required for multi-image support)

* Lots of changes (see above)

* Tested manually on a GKE cluster


core/sbt-cloudflow/src/main/scala/cloudflow/sbt/CloudflowBasePlugin.scala
```diff
@@ -126,7 +127,10 @@ object CloudflowBasePlugin extends AutoPlugin {
   private[sbt] def makeStreamletDescriptorsLabelValue(streamletDescriptorsJson: JsValue) = {
     // create a root object with the array
     val streamletDescriptorsJsonStr =
-      JsObject("streamlet-descriptors" -> streamletDescriptorsJson).compactPrint
+      JsObject(
+        "streamlet-descriptors" -> streamletDescriptorsJson,
+        "api-version"           -> JsString(ApplicationDescriptor.Version)
```
RayRoestenburg: Rename `ApplicationDescriptor.Version` to `ApplicationDescriptor.APIVersion`?

core/cloudflow-spark-testkit/src/main/scala/cloudflow/spark/testkit/TestSparkStreamletContext.scala
```diff
@@ -43,9 +43,8 @@ private[testkit] class TestSparkStreamletContext(override val streamletRef: Stri
                                                  inletTaps: Seq[SparkInletTap[_]],
                                                  outletTaps: Seq[SparkOutletTap[_]],
                                                  override val config: Config = ConfigFactory.empty)
-    extends SparkStreamletContext(StreamletDefinition("appId", "appVersion", streamletRef, "streamletClass", List(), List(), config),
-                                  session) {
-  val ProcessingTimeInterval = 1500.milliseconds
+    extends SparkStreamletContext(StreamletDefinition("appId", streamletRef, "streamletClass", List(), List(), config), session) {
+  val ProcessingTimeInterval = 1.second
```
RayRoestenburg: Why was this changed? (I'm fine with it if that works ok, just asking)

core/sbt-cloudflow/src/main/scala/cloudflow/sbt/CloudflowAkkaPlugin.scala
```diff
@@ -25,10 +25,29 @@ import spray.json._
 import cloudflow.sbt.CloudflowKeys._
 import cloudflow.blueprint.StreamletDescriptorFormat._
 import CloudflowBasePlugin._
+import com.lightbend.sbt.javaagent.JavaAgent
+import com.lightbend.sbt.javaagent.JavaAgent.JavaAgentKeys.resolvedJavaAgents
+import java.io.File
 
 object CloudflowAkkaPlugin extends AutoPlugin {
+  val AppRunner: String = "akka-entrypoint.sh"
+  val AppHome           = "${app_home}"
```
RayRoestenburg: How is  this used?

kubectl-cloudflow/cmd/deploy.go
```diff
@@ -42,9 +46,28 @@ func init() {
 		Use:   "deploy",
 		Short: "Deploys a Cloudflow application to the cluster.",
 		Long: `Deploys a Cloudflow application to the cluster.
-The arguments to the command consists of a docker image path and optionally one
-or more '[streamlet-name].[configuration-parameter]=[value]' pairs, separated by
-a space.
+The behavior of the command depends upon the flags passed in to it. The command supports deployment of
+a cloudflow application either through the blueprint file or through an image.
```
RayRoestenburg:
```suggestion
a cloudflow application either through a blueprint file or through an image.
```

kubectl-cloudflow/cmd/deploy.go
```diff
@@ -42,9 +46,28 @@ func init() {
...
+
+Deployment through a blueprint file can be configured using the --blueprint flag followed by the
+blueprint file name.
+
+	kubectl cloudflow deploy --blueprint ./call-record-aggregator-bluprint.conf
```
RayRoestenburg:
```suggestion
	kubectl cloudflow deploy --blueprint ./call-record-aggregator-blueprint.conf
```

kubectl-cloudflow/cmd/deploy.go
```diff
@@ -42,9 +46,28 @@ func init() {
 		Use:   "deploy",
 		Short: "Deploys a Cloudflow application to the cluster.",
 		Long: `Deploys a Cloudflow application to the cluster.
-The arguments to the command consists of a docker image path and optionally one
-or more '[streamlet-name].[configuration-parameter]=[value]' pairs, separated by
-a space.
+The behavior of the command depends upon the flags passed in to it. The command supports deployment of
+a cloudflow application either through the blueprint file or through an image.
+
+Deployment through a blueprint file can be configured using the --blueprint flag followed by the
```
RayRoestenburg:
```suggestion
Deployment through a blueprint file can be done using the --blueprint flag followed by the
```

kubectl-cloudflow/cmd/deploy.go
```diff
@@ -42,9 +46,28 @@ func init() {
...
+blueprint file name.
+
+	kubectl cloudflow deploy --blueprint ./call-record-aggregator-bluprint.conf
+
+Deployment through a docker image can be configured using the --image flag followed by the image
```
RayRoestenburg:
```suggestion
Deployment through a docker image can be done using the --image flag followed by the image
```

kubectl-cloudflow/cmd/deploy.go
```diff
@@ -42,9 +46,28 @@ func init() {
...
+url.
+
+	kubectl cloudflow deploy --image docker.registry.com/my-company/sensor-data-scala:292-c183d80 
+
+The command optionally takes as arguments configuration parameters specified as key=value pairs in the form of
```
RayRoestenburg:
```suggestion
The command optionally takes configuration parameters as arguments, specified as key=value pairs in the form of
```

kubectl-cloudflow/cmd/deploy.go
```diff
@@ -81,32 +105,198 @@ the stored credentials.
 
 You can update the credentials with the "update-docker-credentials" command.
 `,
-		Example: `kubectl cloudflow deploy registry.test-cluster.io/cloudflow/sensor-data-scala:292-c183d80 valid-logger.log-level=info valid-logger.msg-prefix=valid`,
-		Args:    validateDeployCmdArgs,
-		Run:     deployOpts.deployImpl,
+		Example: `kubectl cloudflow deploy -i registry.test-cluster.io/cloudflow/sensor-data-scala:292-c183d80 valid-logger.log-level=info valid-logger.msg-prefix=valid`,
+		PreRun: func(cmd *cobra.Command, args []string) {
```
RayRoestenburg: 👍

kubectl-cloudflow/cmd/deploy.go
```diff
@@ -81,32 +105,198 @@ the stored credentials.
...
+	deployOpts.cmd.Flags().StringToIntVar(&deployOpts.replicas, "scale", map[string]int{}, "Accepts key/value pairs for replicas per streamlet")
 
 	rootCmd.AddCommand(deployOpts.cmd)
 }
 
 func (opts *deployOptions) deployImpl(cmd *cobra.Command, args []string) {
 	version.FailOnProtocolVersionMismatch()
 
-	imageRef := args[0]
+	if (cmd.Flag("image").Changed == true) && (cmd.Flag("blueprint").Changed == true) {
```
RayRoestenburg: Is this necessary since you already have these checks in PreRun?

kubectl-cloudflow/cmd/deploy.go
```diff
@@ -81,32 +105,198 @@ the stored credentials.
...
+		util.LogAndExit("%s", "Cannot specify both blueprint and image in command")
+	}
+	if (cmd.Flag("image").Changed == false) && (cmd.Flag("blueprint").Changed == false) {
+		util.LogAndExit("%s", "Need to specify either blueprint or image in command")
+	}
+	// if (cmd.Flag("image").Changed == true) && (cmd.Flag("scale").Changed == true) {
```
RayRoestenburg: Remove commented code please.

kubectl-cloudflow/cmd/deploy.go
```diff
@@ -81,32 +105,198 @@ the stored credentials.
...
+		deployBlueprint(cmd, opts, args)
+	}
+}
+
+func deployBlueprint(cmd *cobra.Command, opts *deployOptions, args []string) {
+	// read blueprint file and load its content
+	blueprintFile := cmd.Flag("blueprint").Value.String()
+	content, err := ioutil.ReadFile(blueprintFile)
+	if err != nil {
+		util.LogAndExit("%s", err.Error())
```
RayRoestenburg: I think this needs some more user friendly message.

kubectl-cloudflow/cmd/deploy.go
```diff
@@ -81,32 +105,198 @@ the stored credentials.
...
+	}
+	config, err := loadBlueprintConfigContent(string(content[:]))
+	if err != nil {
+		util.LogAndExit("%s", err.Error())
```
RayRoestenburg: Same here.

kubectl-cloudflow/cmd/deploy.go
```diff
@@ -81,32 +105,198 @@ the stored credentials.
...
+	}
+	fmt.Printf("%s\n", config)
```
RayRoestenburg: Is this debugging output maybe?

kubectl-cloudflow/cmd/deploy.go
```diff
@@ -81,32 +105,198 @@ the stored credentials.
...
+
+	// get the images from the blueprint
+	images := config.GetNode("blueprint.images").GetObject().Items()
```
RayRoestenburg: What if it has no "images" (what error will the user see)?

kubectl-cloudflow/cmd/deploy.go
```diff
@@ -81,32 +105,198 @@ the stored credentials.
...
+
+	var imageRefs []deploy.ImageReference
+	var dockerRegistryURL string
+
+	for _, imageRef := range images {
+		ref, err := parseImageReference(imageRef.GetString())
```
RayRoestenburg: this is a "foo=docker.io/bla/foo:1234" string right? that's not really a Image Reference?

kubectl-cloudflow/cmd/deploy.go
```diff
@@ -81,32 +105,198 @@ the stored credentials.
...
+		if err != nil {
+			util.LogAndExit("%s", err.Error())
+		}
+		imageRefs = append(imageRefs, *ref)
+		dockerRegistryURL = ref.Registry
+	}
+
+	applicationSpec, pulledImages := deploy.GetCloudflowApplicationSpecFromDockerImages(imageRefs, config, opts.replicas)
```
RayRoestenburg: The name of this method is a bit confusing, and I guess the config passed in here is the blueprint?

kubectl-cloudflow/cmd/deploy.go
```diff
@@ -81,32 +105,198 @@ the stored credentials.
...
+	jsonb, err := json.Marshal(applicationSpec)
+	fmt.Println(string(jsonb))
```
RayRoestenburg: debugging right

kubectl-cloudflow/cmd/deploy.go
RayRoestenburg: I guess this creates an cloudflow application CR from the blueprint and the images

RayRoestenburg: pulledImages, do we need to pull them all? or does it inspect?

kubectl-cloudflow/cmd/deploy.go
```diff
@@ -81,32 +105,198 @@ the stored credentials.
...
+
+	namespace := applicationSpec.AppID
+	k8sClient, k8sErr := k8s.GetClient()
+	if k8sErr != nil {
+		util.LogAndExit("Failed to create new kubernetes client for Cloudflow application `%s`, %s", namespace, k8sErr.Error())
+	}
+
+	cloudflowApplicationClient, err := k8s.GetCloudflowApplicationClient(namespace)
```
RayRoestenburg: Thinking about it the idea of a CloudflowApplicationClient is a bit weird, it is just a k8s client right?

kubectl-cloudflow/cmd/deploy.go
```diff
@@ -81,32 +105,198 @@ the stored credentials.
...
+
+	serviceAccount.ObjectMeta.OwnerReferences = []metav1.OwnerReference{ownerReference}
+	if _, err := createOrUpdateServiceAccount(k8sClient, namespace, serviceAccount); err != nil {
+		util.LogAndExit("%s", err)
+	}
+
+	util.PrintSuccess("Deployment of application `%s` has started.\n", namespace)
+}
+
+func loadBlueprintConfigContent(contents string) (*configuration.Config, error) {
```
RayRoestenburg: Would be great if Blueprint had it's own struct that would get loaded into?

RayRoestenburg: Ah right but that will come with @skonto 's work... ok makes sense. (Sorry I'm just reading top down)

kubectl-cloudflow/deploy/deploy.go
```diff
@@ -92,6 +95,311 @@ func GetCloudflowApplicationDescriptorFromDockerImage(dockerRegistryURL string,
 	return spec, *pulledImage
 }
 
+// GetCloudflowApplicationSpecFromDockerImages pulls a image and extracts the Cloudflow Application descriptor from a docker label
```
RayRoestenburg: This comment is incorrect

kubectl-cloudflow/deploy/deploy.go
```diff
@@ -92,6 +95,311 @@ func GetCloudflowApplicationDescriptorFromDockerImage(dockerRegistryURL string,
...
+func GetCloudflowApplicationSpecFromDockerImages(imageRefs []ImageReference,
```
RayRoestenburg: the name can be improved, you don't really get the appspec from docker images. `CreateApplicationSpecFromBlueprintAndImages`.. ehhh ugh... can we maybe do this in two steps?
`GetStreamletDescriptors(imageRefs)`
`CreateCloudflowApplication(blueprint, streamletdescriptors)`
Something like that.

kubectl-cloudflow/deploy/deploy.go
```diff
@@ -92,6 +95,311 @@ func GetCloudflowApplicationDescriptorFromDockerImage(dockerRegistryURL string,
...
+	}
+
+	// get all streamlet descriptors, image digests and pulled images in arrays
+	var streamletDescriptors []domain.Descriptor
+	var imageDigests []string
+	var apiVersion string
+	var pulledImages []*docker.PulledImage
+	var deployImages = make(map[string]string)
+	for _, imageRef := range imageRefs {
+		streamletsDescriptorsDigestPair, version := docker.GetCloudflowStreamletDescriptorsForImage(client, imageRef.FullURI)
```
RayRoestenburg: I think in this naming we can drop the 'Cloudflow' bit.

kubectl-cloudflow/docker/client.go
```diff
@@ -136,3 +137,50 @@ func imageMatchesNameAndTag(image types.ImageSummary, imageNameAndTag string) bo
 	}
 	return false
 }
+
+// GetCloudflowStreamletDescriptorsForImage extracts the configuration of a Cloudflow label from a docker image
+// and the image with digest in a struct
+func GetCloudflowStreamletDescriptorsForImage(cli *client.Client, imageName string) (domain.CloudflowStreamletDescriptorsDigestPair, string) {
```
RayRoestenburg: it would be great if we could get the label from the image without actually pulling the image. (That might be an improvement we can do later, after this PR, but if it is easy, rather do it right away?)

kubectl-cloudflow/domain/cloudflowapplication.go
```diff
@@ -124,6 +129,12 @@ type CloudflowApplicationSpec struct {
 	LibraryVersion string            `json:"library_version,omitempty"`
 }
 
+// Version of the application descriptor format
+const Version = "1"
+
+// LibraryVersion specifies cloudflow release version
+const LibraryVersion = "1.3.1-SNAPSHOT"
```
RayRoestenburg: How will we update this?

kubectl-cloudflow/domain/cloudflowapplication.go
```diff
@@ -173,6 +183,12 @@ type CloudflowApplicationDescriptorDigestPair struct {
 	ImageDigest   string `json:"imageDigest"`
 }
 
+// CloudflowStreamletDescriptorsDigestPair is a pair of streamlet descriptors and image digest
+type CloudflowStreamletDescriptorsDigestPair struct {
+	StreamletDescriptors []Descriptor `json:"streamletDescriptors"`
+	ImageDigest          string       `json:"imageDigest"`
```
RayRoestenburg: What is the digest?

RayRoestenburg: ah this is the 'sha'  for the image, right?

core/cloudflow-spark-testkit/src/main/scala/cloudflow/spark/testkit/TestSparkStreamletContext.scala
debasishg: Will take a look.

core/sbt-cloudflow/src/main/scala/cloudflow/sbt/CloudflowAkkaPlugin.scala
debasishg: `akka-entrypoint.sh` uses `$app_home` - hence it needs to be set which is done here. This is what was earlier as well. But I guess this will change when we have a separate base image for akka.

kubectl-cloudflow/cmd/deploy.go
debasishg: done

kubectl-cloudflow/cmd/deploy.go
debasishg: done ..

kubectl-cloudflow/cmd/deploy.go
debasishg: done ..

kubectl-cloudflow/cmd/deploy.go
debasishg: done ..

kubectl-cloudflow/cmd/deploy.go
debasishg: done ..

kubectl-cloudflow/cmd/deploy.go
debasishg: removed it .. thanks for pointing it out ..

kubectl-cloudflow/cmd/deploy.go
debasishg: done ..

kubectl-cloudflow/cmd/deploy.go
debasishg: changed ..

kubectl-cloudflow/cmd/deploy.go
debasishg: changed ..

kubectl-cloudflow/cmd/deploy.go
debasishg: removed ..

kubectl-cloudflow/cmd/deploy.go
debasishg: @RayRoestenburg  This is the section of code for deploying using a blueprint. And I am assuming that blueprint verification will be done before deploy. This section `images` is a mandatory part of the blueprint configuration. Can we assume that it _will be_ there ?

kubectl-cloudflow/cmd/deploy.go
debasishg: yes, this is a single record from the `images` section of the blueprint. e.g.

```
images {
    aggr = "eu.gcr.io/bubbly-observer-178213/spark-aggregation:184-a48ddfe-dirty"
    ings = "eu.gcr.io/bubbly-observer-178213/akka-cdr-ingestor:184-a48ddfe-dirty"
    outp = "eu.gcr.io/bubbly-observer-178213/akka-java-aggregation-output:184-a48ddfe-dirty"
  }
```

Each of these refers to an image. Suggestion for a better name ?

kubectl-cloudflow/cmd/deploy.go
debasishg: * Yes, the function creates an application CR from images and blueprint. Will `GetCloudflowApplicationSpecFromDockerImagesAndBlueprint` be a better name ? I tried to make it a bit terser :-)

* Yes, the `config` refers to that of the blueprint. I can change the name of the variable to make that explicit.

debasishg: * regarding `pulledImages` I will check if I can optimize it. It's basically the same stuff that was done before with images.

kubectl-cloudflow/cmd/deploy.go
debasishg: removed

kubectl-cloudflow/cmd/deploy.go
debasishg: yeah, k8s client for the cloudflow application - I kept the name that was used before.

kubectl-cloudflow/cmd/deploy.go
debasishg: yes, this will move out once we have @skonto 's code checked in. I tried to make my piece self containing - will do these refactorings before check-in. This will impose a serializable merge with Stavros' work, which I think, is logical.

kubectl-cloudflow/deploy/deploy.go
debasishg: fixed

kubectl-cloudflow/deploy/deploy.go
debasishg: ok .. will refactor

kubectl-cloudflow/deploy/deploy.go
debasishg: 👍

kubectl-cloudflow/docker/client.go
debasishg: will check ..

kubectl-cloudflow/domain/cloudflowapplication.go
debasishg: I was thinking of pulling all such mutable constants (huh! oxymoron) in a single file

kubectl-cloudflow/domain/cloudflowapplication.go
debasishg: yes ..

kubectl-cloudflow/deploy/deploy.go
debasishg: refactored

kubectl-cloudflow/docker/client.go
```diff
@@ -58,6 +60,24 @@ func GetClient(version string) (*client.Client, error) {
...
 
+// GetVersionedClient returns a client with the proepr version
+func GetVersionedClient() (*client.Client, error) {
+	apiversion, apierr := exec.Command("docker", "version", "--format", "'{{.Server.APIVersion}}'").Output()
+	if apierr != nil {
+		return nil, fmt.Errorf("Could not get docker API version, is the docker daemon running? API error: %s", apierr.Error())
+	}
+
+	trimmedapiversion := strings.Trim(string(apiversion), "\t \n\r'")
+	client, error := GetClient(trimmedapiversion)
```
skonto: Go code style requires not to use a reserved keyword name here as It collides with the builtin error interface. The convention is to use `err` or something similar.

skonto: Btw there is a nice code style guide here: https://github.com/golang/go/wiki/CodeReviewComments.

kubectl-cloudflow/cmd/deploy.go
skonto: @debasishg @RayRoestenburg I think we should run the verification process before deployment anyway. The reason you want to have `verify` as a standalone command is because you want the user to be able to verify his blueprint while he constructs them or generates them.

kubectl-cloudflow/cmd/status.go
```diff
@@ -69,12 +69,12 @@ func calcAppStatus(applicationCR *domain.CloudflowApplication) string {
...
 	for _, d := range applicationCR.Spec.Deployments {
-		if d.Replicas == 0 && d.Runtime == "akka" {
+		if d.Replicas == nil && d.Runtime == "akka" {
 			expectedPods = expectedPods + 1 // TODO replica defaults
-		} else if d.Replicas == 0 && d.Runtime == "spark" {
+		} else if d.Replicas == nil && d.Runtime == "spark" {
 			expectedPods = expectedPods + 1 + 2 // TODO replica defaults 1 driver, 2 executors
 		} else {
-			expectedPods = expectedPods + d.Replicas
+			expectedPods = expectedPods + *d.Replicas
```
skonto: if `d.Replicas == nil` && runtime is `flink` this will fail here.

core/cloudflow-spark-testkit/src/main/scala/cloudflow/spark/testkit/TestSparkStreamletContext.scala
maasg: this should be back to `1500.milliseconds`

kubectl-cloudflow/deploy/deploy.go
```diff
@@ -23,19 +27,9 @@ import (
...
-	trimmedapiversion := strings.Trim(string(apiversion), "\t \n\r'")
-	client, error := docker.GetClient(trimmedapiversion)
+	client, error := docker.GetVersionedClient()
 	if error != nil {
-		client, error = docker.GetClient("1.39")
-		if error != nil {
-			fmt.Printf("No compatible version of the Docker server API found, tried version %s and 1.39", trimmedapiversion)
-			panic(error)
-		}
+		panic(error)
```
skonto: Can we not panic here (https://github.com/golang/go/wiki/CodeReviewComments#dont-panic)?

core/cloudflow-spark-testkit/src/main/scala/cloudflow/spark/testkit/TestSparkStreamletContext.scala
RayRoestenburg: @maasg should this be configurable? it seems like a 'magic' number.

kubectl-cloudflow/deploy/deploy.go
```diff
@@ -92,6 +86,328 @@ func GetCloudflowApplicationDescriptorFromDockerImage(dockerRegistryURL string,
...
+	Internal server `json:"internal,omitempty"`
+}
+type server struct {
+	Server containerPort `json:"server,omitempty"`
+}
+type containerPort struct {
+	ContainerPort int `json:"container-port,omitempty"`
+}
+
+func getServerConfigAndEndpoint(appID string, streamlet domain.Streamlet) (json.RawMessage, domain.Endpoint) {
```
skonto: Let's add proper documentation to all our high level functions. It improves code readability and its part of Go's way of doing things (https://blog.golang.org/godoc).

core/sbt-cloudflow/src/main/scala/cloudflow/sbt/CloudflowAkkaPlugin.scala
RayRoestenburg: What sets this `$app_home` variable in the `akka-entrypoint.sh` script? (just curious)

kubectl-cloudflow/deploy/deploy.go
```diff
@@ -379,3 +695,67 @@ func prefixWithStreamletName(streamletName string, key string) string {
...
+// ImageReference is a destructured version of the image path
+type ImageReference struct {
+	Registry   string
+	Repository string
+	Image      string
+	Tag        string
+	FullURI    string
+}
+
+type blueprint struct {
```
skonto: @debasishg are you planning to allign this structure with the one I have in #129 ?
There is overlapping here.

kubectl-cloudflow/deploy/deploy.go
```diff
@@ -379,3 +695,67 @@ func prefixWithStreamletName(streamletName string, key string) string {
 func formatStreamletConfigKeyFq(streamletName string, key string) string {
 	return fmt.Sprintf("cloudflow.streamlets.%s", prefixWithStreamletName(streamletName, key))
 }
+
+/*
+{
+  blueprint : {
```
skonto: This part is redundant right?

kubectl-cloudflow/deploy/deploy_test.go
```diff
@@ -1,103 +1,273 @@
...
+	"github.com/go-akka/configuration"
 	"github.com/lightbend/cloudflow/kubectl-cloudflow/domain"
-	"github.com/stretchr/testify/assert"
 )
 
-func TestMain(m *testing.M) {
-	code := m.Run()
-	os.Exit(code)
+func createBlueprint() string {
+	return `blueprint {
```
RayRoestenburg: very minor, start with newline so blueprint is a little nicer to read?

kubectl-cloudflow/deploy/deploy_test.go
```diff
@@ -1,103 +1,273 @@
...
-		`valid-logger.msg-prefix-5=true`,
-		`valid-logger.msg-prefix-6="20 minutes"`,
-		`valid-logger.msg-prefix-7="20M"`}
+func createImageDigests() []string {
+	digests := make([]string, 3)
+
+	digests[0] = "sha256:98bded501c9e7e9101cbc0971cc529e9d2240459924b30b44b2ea1976c328a46"
+	digests[1] = "sha256:b5eb3e5d5bb245ef322ac9ac1698c7ee7543707c20184a3e94824afc429567e9"
+	digests[2] = "sha256:567e1fb63145a5f267894d70ca49d38573655c6b6d97ca64fb41402d2aede208"
+	return digests
```
RayRoestenburg:
```suggestion
	return []string { 
	  "sha256:98bded501c9e7e9101cbc0971cc529e9d2240459924b30b44b2ea1976c328a46",
	  "sha256:b5eb3e5d5bb245ef322ac9ac1698c7ee7543707c20184a3e94824afc429567e9",
	  "sha256:567e1fb63145a5f267894d70ca49d38573655c6b6d97ca64fb41402d2aede208"
	}
```

kubectl-cloudflow/deploy/deploy_test.go
```diff
@@ -1,103 +1,273 @@
...
+		bp             blueprint
+		streamletIDs   []string
+		streamletIDMap map[string]bool
+	)
 
-	var spec domain.CloudflowApplicationSpec
-	json.Unmarshal([]byte(applicationConfiguration), &spec)
+	BeforeEach(func() {
+		config, err := getBlueprintConfig(createBlueprint())
+		Ω(err).ShouldNot(HaveOccurred())
```
RayRoestenburg: what is the `Ω` for?

RayRoestenburg: oh that must be gomega syntax? would not have expected that in golang...

kubectl-cloudflow/docker/client.go
```diff
@@ -58,6 +60,24 @@ func GetClient(version string) (*client.Client, error) {
 	return client.NewClientWithOpts(client.WithVersion(version))
 }
 
+// GetVersionedClient returns a client with the proepr version
```
RayRoestenburg:
```suggestion
// GetVersionedClient returns a client with the proper version
```

kubectl-cloudflow/docker/client.go
```diff
@@ -136,3 +156,50 @@ func imageMatchesNameAndTag(image types.ImageSummary, imageNameAndTag string) bo
...
+
+// GetStreamletDescriptorsForImage extracts the configuration of a Cloudflow label from a docker image
+// and the image with digest in a struct
+func GetStreamletDescriptorsForImage(cli *client.Client, imageName string) (domain.CloudflowStreamletDescriptorsDigestPair, string) {
+	images, err := cli.ImageList(context.Background(), types.ImageListOptions{})
+
+	streamletDescriptorsLabelName := "com.lightbend.cloudflow.streamlet-descriptors"
+
+	if err != nil {
+		util.LogAndExit("Failed to list local docker images, %s", err.Error())
```
RayRoestenburg: I think it makes more sense for private methods not to exit, but return an error.

kubectl-cloudflow/domain/cloudflowapplication.go
RayRoestenburg: hehe

kubectl-cloudflow/docker/client.go
```diff
@@ -136,3 +156,50 @@ func imageMatchesNameAndTag(image types.ImageSummary, imageNameAndTag string) bo
 	}
 	return false
 }
+
+// GetStreamletDescriptorsForImage extracts the configuration of a Cloudflow label from a docker image
+// and the image with digest in a struct
+func GetStreamletDescriptorsForImage(cli *client.Client, imageName string) (domain.CloudflowStreamletDescriptorsDigestPair, string) {
```
RayRoestenburg: I know this has been done before but I don't really think that `GetStreamletDescriptorsForImage` should be part of the `docker` package. The same goes for `GetCloudflowApplicationDescriptor` like methods. not sure when we should clean up some of these structures. Since the cli will get more code, it's probably good to add a ticket for this.

kubectl-cloudflow/docker/client.go
RayRoestenburg: We should only LogAndExit in handling the commands on top level.

kubectl-cloudflow/cmd/deploy.go
RayRoestenburg: correct deploy must always verify. That's not here yet cause debasish needs to merge with your code, I guess.

kubectl-cloudflow/deploy/deploy.go
skonto: Same for the other structs.

kubectl-cloudflow/docker/client.go
debasishg: fixed here and in some other places too ..

kubectl-cloudflow/cmd/status.go
debasishg: fixed

kubectl-cloudflow/deploy/deploy.go
debasishg: Replaced this with `LogAndExit` but did not change in all places since I am not sure about the rationale of the original author. May take it on a case by case basis

kubectl-cloudflow/deploy/deploy.go
debasishg: I looked at the `blueprint` structure in your PR. My requirement is a bit different and the struct I need is much smaller. Hence I decided to use a leaner version of the struct privately in the `deploy` package. But yes, some other structs will be reused.

kubectl-cloudflow/deploy/deploy.go
debasishg: removed

kubectl-cloudflow/deploy/deploy_test.go
debasishg: done

kubectl-cloudflow/deploy/deploy_test.go
debasishg: done ..

kubectl-cloudflow/deploy/deploy_test.go
debasishg: yeah .. replaced with sanity .. (my Haskell instincts took over momentarily - sorry)

kubectl-cloudflow/docker/client.go
debasishg: changed

kubectl-cloudflow/docker/client.go
debasishg: fixed for this instance .. there may be others (not part of this PR) which we can clean up globally in a separate PR

kubectl-cloudflow/docker/client.go
debasishg: 👍

kubectl-cloudflow/deploy/deploy.go
RayRoestenburg: I think this is a great one for code style / rules. I think we should almost never panic in  the CLI. We need to always provide the user with simple to understand feedback.

kubectl-cloudflow/deploy/deploy.go
RayRoestenburg: This is private though? let's make a decision on code improvement doc.

kubectl-cloudflow/deploy/deploy.go
RayRoestenburg: since verify will always happen as part of deploy, can't you use the same struct?

kubectl-cloudflow/deploy/deploy_test.go
RayRoestenburg: lol

kubectl-cloudflow/cmd/deploy.go
```diff
@@ -42,9 +46,28 @@ func init() {
 		Use:   "deploy",
 		Short: "Deploys a Cloudflow application to the cluster.",
 		Long: `Deploys a Cloudflow application to the cluster.
-The arguments to the command consists of a docker image path and optionally one
-or more '[streamlet-name].[configuration-parameter]=[value]' pairs, separated by
-a space.
+The behavior of the command depends upon the flags passed in to it. The command supports deployment of
+a cloudflow application either through a blueprint file or through an image.
```
RayRoestenburg:
```suggestion
a cloudflow application either through a blueprint file or through a docker image.
```

kubectl-cloudflow/cmd/deploy.go
```diff
@@ -81,32 +105,195 @@ the stored credentials.
...
+		util.LogAndExit("Cannot read blueprint file: Error : %s", err.Error())
+	}
+	// this needs to be replaced with calls from blueprint package once verify blueprint is in
+	blueprintConfig, err := loadBlueprintConfigContent(string(content[:]))
+	if err != nil {
+		util.LogAndExit("Cannot load blueprint file information: Error : %s", err.Error())
+	}
+
+	// get the images from the blueprint
+	imagesByName := blueprintConfig.GetNode("blueprint.images").GetObject().Items()
```
RayRoestenburg: just make sure this can't crash, hopefully it just returns an empty slice

kubectl-cloudflow/cmd/deploy.go
RayRoestenburg: This is fine, you only use what is on the right side of equals sign, which are image references.

kubectl-cloudflow/cmd/deploy.go
```diff
@@ -81,32 +105,195 @@ the stored credentials.
...
+
+	var imageRefs []deploy.ImageReference
+	var dockerRegistryURL string
+
+	for _, imageRef := range imagesByName {
+		ref, err := parseImageReference(imageRef.GetString())
+		if err != nil {
+			util.LogAndExit("Cannot parse image reference from blueprint: Error : %s", err.Error())
```
RayRoestenburg: Minor, you could collect all errors and tell the user at once about all image references that could not be parsed, instead of finding them one by one.

kubectl-cloudflow/cmd/deploy.go
```diff
@@ -81,32 +105,195 @@ the stored credentials.
...
+		}
+		imageRefs = append(imageRefs, *ref)
+
+		// currently we assume all images in the blueprint will have the same registry
+		// hence overwriting is ok. However we need to handle this differently when we have 
+		// support for multiple registries in blueprint
+		dockerRegistryURL = ref.Registry
```
RayRoestenburg: FTR this will be checked by Stavros's code, which we discussed was the best way to do it.

kubectl-cloudflow/cmd/deploy.go
```diff
@@ -81,32 +105,195 @@ the stored credentials.
...
+	}
+
+	applicationSpec, pulledImages := deploy.CreateApplicationSpecFromBlueprintAndImages(imageRefs, blueprintConfig, opts.replicas)
```
RayRoestenburg: minor: maybe name `opts.replicas`, `opts.replicasByStreamletName` instead?

kubectl-cloudflow/cmd/deploy.go
RayRoestenburg: ok we can fix this in cleanup of code in separate PR.

kubectl-cloudflow/cmd/deploy.go
```diff
@@ -81,32 +105,195 @@ the stored credentials.
...
+
+	namespace := applicationSpec.AppID
+	k8sClient, k8sErr := k8s.GetClient()
+	if k8sErr != nil {
+		util.LogAndExit("Failed to create new kubernetes client for Cloudflow application `%s`, %s", namespace, k8sErr.Error())
+	}
+
+	cloudflowApplicationClient, err := k8s.GetCloudflowApplicationClient(namespace)
+	if err != nil {
+		util.LogAndExit("Failed to create new client for Cloudflow application `%s`, %s", namespace, err.Error())
```
RayRoestenburg: This text is a little vague for the user. what really happened here?

kubectl-cloudflow/cmd/deploy.go
```diff
@@ -81,32 +105,195 @@ the stored credentials.
...
+		// support for multiple registries in blueprint
+		dockerRegistryURL = ref.Registry
+	}
+
+	applicationSpec, pulledImages := deploy.CreateApplicationSpecFromBlueprintAndImages(imageRefs, blueprintConfig, opts.replicas)
+
+	namespace := applicationSpec.AppID
+	k8sClient, k8sErr := k8s.GetClient()
+	if k8sErr != nil {
+		util.LogAndExit("Failed to create new kubernetes client for Cloudflow application `%s`, %s", namespace, k8sErr.Error())
```
RayRoestenburg: Is this the same as could not connect? I think this text can be improved, to be more user friendly, less implementation jargon:

'Failed to connect to kubernetes cluster: %s'
I think reporting the name of the app back is not necessary (for Cloudflow application `%s`), they just typed it in, and they know this is cloudflow...

kubectl-cloudflow/cmd/deploy.go
RayRoestenburg: It's also weird that we create a k8sclient and then we need another client, which talks to k8s, but this is somehow a cloudflowApplicationClient. The code for creating those two is weird. We can delay to pick this up on cleanup. But have a look at what happens there.

kubectl-cloudflow/cmd/deploy.go
debasishg: exactly .. will be there once verify code is in place. That's why I suggest @skonto 's PR to be merged first and then mine @RayRoestenburg

kubectl-cloudflow/cmd/deploy.go
debasishg: changed the name of the variable as we discussed yesterday

kubectl-cloudflow/deploy/deploy.go
debasishg: I was thinking that we should document only the _public_ functions. Should we also document private ones ? @skonto @RayRoestenburg

kubectl-cloudflow/deploy/deploy.go
debasishg: will revisit this part once the verify code goes in ..

kubectl-cloudflow/cmd/deploy.go
debasishg: added this as a comment ..

kubectl-cloudflow/cmd/deploy.go
```diff
@@ -81,32 +105,195 @@ the stored credentials.
...
+	// read blueprint file and load its content
+	blueprintFile := cmd.Flag("blueprint").Value.String()
+	content, err := ioutil.ReadFile(blueprintFile)
+	if err != nil {
+		util.LogAndExit("Cannot read blueprint file: Error : %s", err.Error())
+	}
+	// this needs to be replaced with calls from blueprint package once verify blueprint is in
+	blueprintConfig, err := loadBlueprintConfigContent(string(content[:]))
+	if err != nil {
+		util.LogAndExit("Cannot load blueprint file information: Error : %s", err.Error())
```
RayRoestenburg:
```suggestion
		util.LogAndExit("Cannot load blueprint file information: Error : %s", err.Error())
```
```suggestion
		util.LogAndExit("Cannot load blueprint file: Error : %s", err.Error())
```

RayRoestenburg: Actually, the only thing that can go wrong here is that the format is wrong, so maybe just say that.

kubectl-cloudflow/deploy/deploy.go
```diff
@@ -23,19 +27,9 @@ import (
...
-	client, error := docker.GetClient(trimmedapiversion)
-	if error != nil {
-		client, error = docker.GetClient("1.39")
-		if error != nil {
-			fmt.Printf("No compatible version of the Docker server API found, tried version %s and 1.39", trimmedapiversion)
-			panic(error)
-		}
+	client, err := docker.GetVersionedClient()
+	if err != nil {
+		panic(err)
```
RayRoestenburg: We should not panic IMHO

RayRoestenburg: I think this function should not logAndExit. Only use LogAndExit from the command handling code.

kubectl-cloudflow/deploy/deploy.go
```diff
@@ -92,6 +86,331 @@ func GetCloudflowApplicationDescriptorFromDockerImage(dockerRegistryURL string,
...
 }
 
+// CreateApplicationSpecFromBlueprintAndImages pulls all images necessary to create a Cloudflow Application descriptor
+// from a docker label. The function returns the created application spec and the pulled in images
+func CreateApplicationSpecFromBlueprintAndImages(imageRefs []ImageReference,
+	blueprintConfig *configuration.Config, replicas map[string]int) (domain.CloudflowApplicationSpec, []*docker.PulledImage) {
+
+	client, err := docker.GetVersionedClient()
+	if err != nil {
+		util.LogAndExit(err.Error())
```
RayRoestenburg: This function should not logAndExit but return an error.

kubectl-cloudflow/deploy/deploy.go
```diff
@@ -92,6 +86,331 @@ func GetCloudflowApplicationDescriptorFromDockerImage(dockerRegistryURL string,
...
+	spec.Streamlets = streamlets
+	spec.Deployments = deployments
+	return spec
+}
+
+func getConnections(bp blueprint) []domain.Connection {
+	var conns []domain.Connection
+	for inlet, outlets := range bp.connections {
+		for _, outlet := range outlets {
+			// need to switch inlet and outlet as we have them reversed in the blueprint map of connections
```
RayRoestenburg: Not sure what this comment means? something to still fix?
Or does it say what it does? (the "need to " throws me off)

kubectl-cloudflow/cmd/deploy.go
debasishg: fixed

kubectl-cloudflow/cmd/deploy.go
debasishg: done ..

kubectl-cloudflow/cmd/deploy.go
debasishg: done ..

kubectl-cloudflow/cmd/deploy.go
debasishg: done ..

kubectl-cloudflow/cmd/deploy.go
debasishg: done ..

kubectl-cloudflow/cmd/deploy.go
debasishg: done ..

kubectl-cloudflow/cmd/deploy.go
debasishg: done ..

kubectl-cloudflow/deploy/deploy.go
debasishg: done ..

kubectl-cloudflow/deploy/deploy.go
debasishg: done ..

kubectl-cloudflow/deploy/deploy.go
debasishg: removed the comment

core/cloudflow-spark-testkit/src/main/scala/cloudflow/spark/testkit/TestSparkStreamletContext.scala
maasg: It's a magic number.

RayRoestenburg:
Since this work is now in the multi-image feature branch, this PR can be closed, correct @debasishg @skonto ?

PR #192, opened 2020-03-19
Resolving local.conf so that substitutions work.

RayRoestenburg:

local.conf is now resolved, so that substitutions work.

To make substitutions work:
```
foo = 1
bar = ${foo}
```

Substitutions will work in local.conf

local test.


maasg:
+1

PR #193, opened 2020-03-20
Csp1919 Add End to end test

maasg:
This PR implements an end-to-end test that attempts to execute the complete application deployment flow.
```
ensure clean cluster -> 
deploy -> 
check app list -> 
check app status -> 
check pods ready -> 
check output foreach (Akka, Spark, Flink) ->
undeploy
```
It should be clear from the Specs that describe the behavior expected for the different stages of the app lifecycle.


integration-tests/itest/deploy_test.go
```diff
@@ -0,0 +1,461 @@
...
+	cmd := exec.Command("kubectl", "cloudflow", "list")
+	out, er := cmd.CombinedOutput()
+	if er != nil {
+		log.Fatal("could not check app status")
+		err = er
+		return
+	}
+	str := string(out)
+	splits := strings.Split(str, "\n")
+	whitespaces := regexp.MustCompile(`\s+`)
```
olofwalker: A capital `w` away from disaster :)

integration-tests/itest/deploy_test.go
```diff
@@ -0,0 +1,461 @@
...
+func ensureAppNotDeployed(app app) error {
+	apps, err := listApps()
+	if err != nil {
+		return err
+	}
+	found := false
+	for _, entry := range apps {
+		if entry.name == app.name {
+			fmt.Printf("This is the app you are looking for: [%s]", entry.name)
+			found = true
```
olofwalker: insert `break` ?

integration-tests/itest/deploy_test.go
```diff
@@ -0,0 +1,461 @@
...
+		}
+	}
+	if found {
+		err := undeploy(app)
+		if err != nil {
```
olofwalker: You could make this a tiny amount less annoying by replacing these two lines with ``if err:= undeploy(app); err != nil {``

integration-tests/itest/deploy_test.go
```diff
@@ -0,0 +1,461 @@
...
+
+func undeploy(app app) error {
+	fmt.Printf("Issuing Undeploy of app [%s]\n", app.name)
+	cmd := exec.Command("kubectl", "cloudflow", "undeploy", app.name)
+	_, err := cmd.CombinedOutput()
+	return err
+}
+
+func ensureNoPods(app app) error {
+	fmt.Printf("Ensuring no pods for app [%s]\n", app.name)
```
olofwalker: I think you could use a `Ticker` here to clean this up a bit.

https://gobyexample.com/tickers

integration-tests/itest/deploy_test.go
```diff
@@ -0,0 +1,461 @@
...
+		res = append(res, entry.streamlet)
+	}
+	return res
+}
+
+func getStatus(app app) (status appStatus, err error) {
+	cmd := exec.Command("kubectl", "cloudflow", "status", app.name)
+	out, er := cmd.CombinedOutput()
+	if er != nil {
+		err = er
```
olofwalker: Not sure if this is a pattern (with named return variables) commonly used in Go, it adds one line for every error also.

integration-tests/itest/deploy_test.go
```diff
@@ -0,0 +1,461 @@
...
+			streamletPod.ready = ready
+			streamletPods = append(streamletPods, streamletPod)
+		}
+	}
+	status.streamletPods = streamletPods
+	return status, nil
+}
+
+func parseLineInN(str string, segments int) (parsed []string, err error) {
+	parts := Whitespaces.Split(str, AllSubstrings)
```
olofwalker: I would just move the regexp down here instead of having an exported global variable.

integration-tests/itest/deploy_test.go
maasg: how do you return only an error in this case? 
I can't:
`return err // doesn't match signature`
`return nil, err` // nil is not of type appStatus`
What's the way to _GO_ ?

olofwalker: You should initialize an empty instance of the class and the error, some people also chose to return a pointer, then you can `return nil, err` , but I recommend the former.

integration-tests/itest/cli/cli.go
```diff
@@ -0,0 +1,13 @@
+package cli
+
+import (
+	"fmt"
+	"os/exec"
+)
+
+func Undeploy(appName string) error {
```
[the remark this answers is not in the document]
maasg: I'm experiencing that right now ;-)

integration-tests/itest/deploy_test.go
maasg: That's quite some scaffolding to set it up. I'm not sure that there's a cost/benefit advantage.
I reduced the complexity of the code in question. Maybe good enough?

integration-tests/itest/deploy_test.go
maasg: Yikes. Is there some form of type ascription possible?

olofwalker: Stop fighting it :)

maasg: :smile:

integration-tests/README.md
```diff
@@ -0,0 +1,19 @@
...
+
+
+## Running the integration test suite
+### Prerequisites
+You must have a working installation of Go v1.14 or later.
+You must have `kubectl` installed and configured to talk to the target cluster.
+You must have an installed Cloudflow cluster available for the integration tests.
+
+### Manual Execution
+From the `/itest` directory, run `ginkgo -v` or `go test`
```
agemooij:
```suggestion
From the `/integration-tests` directory, run `ginkgo -v` or `go test`
```

agemooij: Do you need to have `ginkgo` installed or does that happen automatically?

agemooij:
```suggestion
From the `itest` sub-directory, run `ginkgo -v` or `go test`
```

integration-tests/itest/cli/cli.go
```diff
@@ -0,0 +1,191 @@
+package cli
```
agemooij: Is this part of the test or part of our testing framework?
Could we make the distinction more strict?

agemooij: Are these utils for testing the cli? Tests for the cli?

integration-tests/itest/deploy_test.go
```diff
@@ -0,0 +1,306 @@
...
+
+// Assumptions:
+// In this test, we are going to deploy the test application to an existing cluster.
+// For this first step, we assume that the cluster is correctly deployed and that kubectl is configured to talk to it.
+// The application is already built
+// We also assume that the test is going to be executed manually.
+// Automation will be the next step
+
+const (
+	JsonTokenSrc    = "/keybase/team/assassins/gcloud/pipelines-serviceaccount-key-container-registry-read-write.json"
```
agemooij: What ere these credentials required for?
Is there a way we can decouple this so people outside of the core Cloudflow team, including the CI server, can also run them? Can we inject credentials through an environment var?

integration-tests/itest/deploy_test.go
```diff
@@ -0,0 +1,306 @@
...
+	ShortTimeout    = 60  // seconds
+	LongTimeout     = 240 // seconds
+	InitialWaitTime = "30s"
+)
+
+var swissKnifeApp = cli.App{
+	Image: "eu.gcr.io/bubbly-observer-178213/swiss-knife:189-277e424",
```
agemooij: I think we should publish this in a public docker registry (e.g. the lightbend dockerhub?) so no credentials are required for pulling and non-team members can also run these tests.

integration-tests/itest/deploy_test.go
```diff
@@ -0,0 +1,306 @@
...
+
+	Context("when I deploy an application", func() {
+		It("should start a deployment", func() {
+			jsonToken := getToken()
+			output, err := cli.Deploy(swissKnifeApp, jsonToken)
+			Expect(err).NotTo(HaveOccurred())
+			expected := "Deployment of application `" + swissKnifeApp.Name + "` has started."
+			Expect(output).To(ContainSubstring(expected))
+			waitTime, _ := time.ParseDuration(InitialWaitTime)
+			time.Sleep(waitTime) // this wait is needed to let the application deploy
```
agemooij: I agree with the idea discussed yesterday that watching the cluster for the relevant events instead of waiting a set time will be a better way to determine when the deployment has succeeded or failed.

If this is hard, for instance if we can't just monitor status updated on the application CR, that should lead to feature requests. E.g. we should not be polling all expected pods sine that is an implementation detail of the Cloudflow operator. We should watch for the operator to mark the application (or all its pods) as running in the Status section of the CR.

integration-tests/itest/deploy_test.go
maasg: This is the JSON Token for the Docker repo. I think that we should externalize this in the scope of "running test on CI" - This should work for the team in the current "run manual" mode.

integration-tests/itest/deploy_test.go
```diff
@@ -0,0 +1,306 @@
...
+			close(done)
+		}, LongTimeout)
+	})
+
+	Context("A deployed test application that uses akka, spark, and flink", func() {
+		It("should contain a spark process", func() {
+			status, err := cli.Status(swissKnifeApp)
+			Expect(err).NotTo(HaveOccurred())
+			streamlets := cli.GetStreamlets(&status)
+			Expect(streamlets).To(ContainElement("spark-process"))
```
agemooij: would it be a good idea to make the test less brittle by checking for the runtime instead of the hardcoded streamlet name?

integration-tests/itest/deploy_test.go
maasg: good idea!

integration-tests/itest/deploy_test.go
```diff
@@ -0,0 +1,306 @@
...
+
+		It("should contain an akka process", func() {
+			status, err := cli.Status(swissKnifeApp)
+			Expect(err).NotTo(HaveOccurred())
+			streamlets := cli.GetStreamlets(&status)
+			Expect(streamlets).To(ContainElement("akka-process"))
+		})
+
+		It("should produce a counter in the raw output log", func(done Done) {
+			pod, err := cli.GetSinglePodForStreamlet(swissKnifeApp, "raw-egress")
```
agemooij: Does this mean "get one of the (possibly many) pods for this streamlet"? If so, I would rename it to `getOneOfThePodsForStreamlet` or something similar. The word "single" here can read as an expectation that there should only be a single pod, which is probably not true for this app/all apps.

integration-tests/itest/deploy_test.go
```diff
@@ -0,0 +1,306 @@
...
+			close(done)
+		}, LongTimeout)
+
+	})
+
+	Context("A deployed application can be undeployed", func() {
+		It("should undeploy the test app", func(done Done) {
+			err := cli.Undeploy(swissKnifeApp)
+			Expect(err).NotTo(HaveOccurred())
+			err = expectFindApp(swissKnifeApp, false)
```
agemooij: Is there no explicit wait time involved here? Or is it hidden somewhere?

integration-tests/itest/deploy_test.go
```diff
@@ -0,0 +1,306 @@
...
+			Expect(err).NotTo(HaveOccurred())
+			close(done)
+		}, ShortTimeout)
+	})
+})
+
+func expectFindApp(app cli.App, expected bool) error {
```
agemooij: I suggest we move all of these utils to a Golang package since they all look like general test utils instead of being specific for this test.

integration-tests/itest/deploy_test.go
maasg: Would a user subscribe to k8s events to know that the deployment succeeded? 
The intention here is to use the tools we offer to the users to exercise the cloudflow application lifecycle. 
In this case, we have:
```
kubectl cloudflow deploy <app>
```
which is followed later by:
```
kubectl cloudflow status <app>
-> has status my expected streamlets?
```

integration-tests/itest/deploy_test.go
maasg: How could we do that with `kubectl cloudflow` ? 
Maybe the answer is that we should have a column "backend" in the status report. 
This is how it looks like right now:
```
$ kubectl cloudflow status swiss-knife
Name:             swiss-knife
Namespace:        swiss-knife
Version:          189-277e424
Created:          2020-03-24 12:53:12 +0100 CET
Status:           Pending

STREAMLET         POD                              STATUS            RESTARTS          READY             
spark-process     swiss-knife-spark-process-driver Unknown           0                 
ingress           swiss-knife-ingress-driver       Unknown           0   
```

integration-tests/itest/deploy_test.go
maasg: Timeout management is provided by the framework. Notice the `func(done Done)` which provides a `channel` to communicate that the test is done. If not, the test will interrupt when the time is over and fail with a timeout. 
The timeout period is specified as a parameter to the call, which you find below :arrow_down: 
`, ShortTimeout)`

integration-tests/itest/deploy_test.go
maasg: I'd argue that `ListApps()` belongs to the package and this method is a helper method to implement the specific test logic, in this case, assert that the app is in the list. (in scala: `ListApps().exist(e => e.name == app.name)`)
I think that moving it into the pkg would pollute it as each test will contribute to such noise. But I'm a Go noob.
What do other team members think? @olofwalker @RayRoestenburg @skonto

integration-tests/itest/cli/cli.go
maasg: I think that becomes clear with the usage
```
cli.Deploy(swissKnifeApp, jsonToken)
...
cli.ListApps()
...
cli.Status(swissKnifeApp)
```

integration-tests/itest/deploy_test.go
maasg: done

integration-tests/itest/deploy_test.go
olofwalker: I would prefer to have tests in one package and support for tests in one.

maasg: @olofwalker currently it's structured like this:
```
test suite
|-- cli (supporting lib. Maybe add kctl next to it)

```
Do you suggest to reorganize it as:
```
main (noop)
 |- test suite
 |- cli
 |- kctl (tbc)
```
?
I particularly struggle with supporting functions, such as map-like transformations (e.g. `GetStreamletNames(StreamletPods)`. It seems silly to put them in a separate package.
Or maybe it's the way to _GO_ ?

integration-tests/itest/deploy_test.go
agemooij: So you are saying this is a first attempt and you agree that we need to externalize this in subsequent steps, right?
I can agree with that but what I want to prevent is us writing dozens of these tests with hardcoded, team-only credentials. So I suggest we pick up this particular issue as part of the next step, which is "Ci-ification" of this test

I would also urge us to use public repos instead of protected repos t sidestep the entire issue, unless we explicitly want to test registry authentication aspects of the CLI.

integration-tests/itest/deploy_test.go
agemooij: Exactly. I was reacting to the explicit wait, which you explained is only there due to a bug in the current implementation of application status. I wold appreciate a comment above this line and a link to the relevant issue so any reader will known why we are doing this and can see when it will be fixed.

integration-tests/itest/deploy_test.go
agemooij: I agree part of our end to end testing is about testing the user interaction. IMO that shouldn't mean we cannot use the lower level aspects of kubernetes, all accessible through kubectl, to make our tests more powerful.

that said, re-reading your test I agree that this usage of hardcoded streamlet names is fairly acceptable. The biggest problem for me is the high decoupling between the test application and the tests, which makes it likely that changes in the test application will frequently break these tests. I was looking for a way to make that less likely for innocent changes (as in: they would not have a breaking effect in the real world) such as changing the name of a streamlet.

integration-tests/itest/deploy_test.go
agemooij: It is fairly unintuitive to me to know that behind the scenes this is actively a repeating poll that only gets short-circuited by either the expected answer being found or the timeout firing.

It's a minor point but IMO for the reader to understand what is going on here it would be useful to get a hint (from the name of the function) that "polling until" is involved.

integration-tests/itest/deploy_test.go
agemooij: But `expectFindApp` and `checkLastLogsContains` are clearly general purpose utils that will be reused in other tests, right? There is nothing specific to this particular tests in the logic for those functions or others.

So unless you think that this one spec will be the only spec we will need to perform those checks, I think they should be more separate.

agemooij: I'm fine with waiting a little bit until we write  second test but what I want to prevent is dozens of minor variations of these util functions or one tests needing to import another test to gain access to its utils.

integration-tests/itest/deploy_test.go
maasg: I Agree. I'll change to something like "PollingFor..."

integration-tests/itest/deploy_test.go
maasg: done.

integration-tests/itest/deploy_test.go
maasg: created a public ticket and added a link to it:
https://github.com/lightbend/cloudflow/issues/201

integration-tests/itest/deploy_test.go
maasg: I changed the test structure to make clear that this test is checking the `status` output and not the runtime. It would be good to have a "spark" and "flink" suites that check specific characteristics and behavior for each backend.

integration-tests/itest/deploy_test.go
maasg: I reorganized the test and helper classes to reduce noise as much as possible.

maasg:
@agemooij I think I addressed all your comments. Would you like to give it another look?

integration-tests/itest/deploy_test.go
maasg: done!

integration-tests/itest/cli/cli.go
agemooij: [very minor, personal opinion only] I do indeed understand it from its call-site usage. It is clear there.
From a code browsing and file/folder structure perspective I would personally have changed it from `cli/cli.go` to `testkit/cli.go`, `testkit/cli/cli.go` or some such similar indication of purpose.

integration-tests/itest/deploy_suite_test.go
```diff
@@ -0,0 +1,13 @@
...
+	"testing"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+func TestDeploy(t *testing.T) {
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "Deploy Suite")
+}
```
agemooij: For me, not being a Gingko expert, it is not clear how this line somehow runs the tests in `deploy_test.go`. How are those two linked? By name? By convention?

agemooij: It might be useful for us non-Golang experts to add a comment explaining how to work with this in the future.

integration-tests/itest/deploy_test.go
```diff
@@ -0,0 +1,227 @@
...
+	if found {
+		if err := cli.Undeploy(app); err != nil {
+			return err
+		}
+		return ensureNoPods(app)
+	}
+	return nil
+}
+
+func ensureNoPods(app cli.App) error {
```
agemooij: This seems like a generally useful function to be reused on other tests. I suggest we extract it to `kctl.go`

integration-tests/itest/deploy_test.go
```diff
@@ -0,0 +1,227 @@
...
+			return err
+		}
+		fmt.Printf("...%d", len(pods))
+	}
+	return nil
+}
+
+// pollUntilPodsStatusIs polls the status of each pod of the application to be the expected status
+// returns when all pods have the expected status.
+func pollUntilPodsStatusIs(app cli.App, expected string) (res string, err error) {
```
agemooij: This seems like a generally useful function to be reused on other tests. I suggest we extract it to `kctl.go`

integration-tests/itest/kctl/kctl.go
```diff
@@ -0,0 +1,84 @@
+package kctl
```
agemooij: [minor] I do not think it is very useful to shorten `kubectl` to `kctl`. It introduces a cognitive load, forcing people to get used to a new alias for a well-known tool. I suggest you rename this to `kubectl`.

integration-tests/itest/deploy_suite_test.go
maasg: I hope a Go expert can explain. This is auto-generated by the framework. 
(Intuitively, I think that this line hooks into the built-in test support by registering one test called "TestDeploy" and uses that to bootstrap its own runtime.)

maasg: @agemooij the [ginkgo home](https://onsi.github.io/ginkgo/) page has a breakdown of this structure under the 'getting started' head. 
>Let’s break this down:

Go allows us to specify the books_test package alongside the books package. Using books_test instead of books allows us to respect the encapsulation of the books package: your tests will need to import books and access it from the outside, like any other package. This is preferred to reaching into the package and testing its internals and leads to more behavioral tests. You can, of course, opt out of this – just change package books_test to package books
We import the ginkgo and gomega packages into the test’s top-level namespace by performing a dot-import. If you’d rather not do this, check out the Avoiding Dot Imports section below.
TestBooks is a testing test. The Go test runner will run this function when you run go test or ginkgo.
RegisterFailHandler(Fail): A Ginkgo test signals failure by calling Ginkgo’s Fail(description string) function. We pass this function to Gomega using RegisterFailHandler. This is the sole connection point between Ginkgo and Gomega.
RunSpecs(t *testing.T, suiteDescription string) tells Ginkgo to start the test suite. Ginkgo will automatically fail the testing.T if any of your specs fail.

maasg: I don't see a need to replicate any of that on our code.

maasg:
Added a fail-fast short-circuit in case the cluster is not available or cloudflow is not installed.
There's no point in testing in that case.

PR #197, opened 2020-03-23
Second and final set of documentation snippets

skyluc:
Extracted the last of the code snippets (Scala, Java, blueprint) out of `cloudflow-docs`.
Integrated the snippets in snippet projects, so the code included in the documentation can be compiled, tested and checked.

As the Cloudflow project evolves, any code snippets in the documentation can become out-dated.
Integrating the snippets in the project and the CI allows to keep the code and the documentation aligned.
A large portion of the extracted snippets were already broken, and required fixes to have them pass the checks.

The snippets in the documentation will be correct again.

Running the CI build. Building the documentation against this branch. Checking the generated documentation manually.


skyluc:
Cloudflow docs PR: lightbend/cloudflow-docs#44

examples/call-record-aggregator/project/cloudflow-plugins.sbt
```diff
@@ -2,5 +2,5 @@
 //
 resolvers += "Akka Snapshots" at "https://repo.akka.io/snapshots/"
 resolvers += Resolver.url("cloudflow", url("https://lightbend.bintray.com/cloudflow"))(Resolver.ivyStylePatterns)
-addSbtPlugin("com.lightbend.cloudflow" % "sbt-cloudflow" % "1.3.1")
+addSbtPlugin("com.lightbend.cloudflow" % "sbt-cloudflow" % "1.3.2-SNAPSHOT")
```
agemooij: This is not a published build, right. AFAIk you need to create an internal release to make this work.

agemooij: Yet the CI build is not broken for reasons I do not understand unless someone has pushed a SNAPSHOT jar to the repo.

skyluc: ah right. It is because I run the CI tests. I'll revert those changes.

skyluc: The CI build a `-SNAPSHOP` version of the core locally, and update all the examples project to use this built version. So it doesn't really care of this version string.
It is made to ensure that all the examples are up-to-date with the core code.

agemooij: True, but it would break the build for anyone else checking out this branch, or master after merging this, since they wouldn't have the SNAPSHOT build (or they have an older one)

examples/snippets/modules/ROOT/examples/build-akka-streamlets-scala/step3/src/test/scala/com/example/TestProcessorSpec.scala
```diff
@@ -0,0 +1,52 @@
...
+  private implicit val system = ActorSystem("AkkaStreamletSpec")
+  private implicit val mat = ActorMaterializer()
+
+  //tag::afterAll[]
+  override def afterAll: Unit = {
+    TestKit.shutdownActorSystem(system)
+  }
+  //end::afterAll[]
+
+  "An TestProcessor" should {
```
agemooij: [super-duper minor]: 
```suggestion
  "A TestProcessor" should {
```

examples/snippets/modules/ROOT/examples/build-spark-streamlets-scala/build.sbt
```diff
@@ -0,0 +1,79 @@
+lazy val root =
```
agemooij: Was this new project added to the examples.yaml file so we can update versions in it and run targets against it with the examples scripts?

skyluc: Good catch. Maybe we should just list the folders ... :grin:

PR #200, opened 2020-03-25
Csp 1962 scale coverage

maasg:
Adds coverage for the scale feature in `kubectl cloudflow`
Tests that each type of streamlet (Akka, spark, flink) can be scaled up and down on a deployed application running on an actual cluster.

### DIFF
--- a/integration-tests/itest/cli/cli.go
+++ b/integration-tests/itest/cli/cli.go
@@ -7,6 +7,7 @@ import (
 	"regexp"
 	"strconv"
 	"strings"
+	"time"
 )
 
 // App represents an application name and docker image
@@ -42,6 +43,8 @@ type StreamletPod struct {
 	Ready     bool
 }
 
+var pollSleepInterval, _ = time.ParseDuration("5s")
+
 // Deploy initiates the deployment of an application to the k8s cluster
 func Deploy(app App, pwd string) (deployRes string, deployErr error) {
 	cmd := exec.Command("kubectl", "cloudflow", "deploy", app.Image, "--username", "_json_key", "--password", pwd)
@@ -52,6 +55,13 @@ func Deploy(app App, pwd string) (deployRes string, deployErr error) {
 // Undeploy initiates an application undeployment on the active cluster
 func Undeploy(app App) error {
 	cmd := exec.Command("kubectl", "cloudflow", "undeploy", app.Name)
+	_, err := cmd.CombinedOutput()
+	return err
+}
+
+// Scale changes the scale factor of a streamlet in an app
+func Scale(app App, streamlet string, scale int) error {
+	cmd := exec.Command("kubectl", "cloudflow", "scale", app.Name, streamlet, strconv.Itoa(scale))
 	_, err := cmd.CombinedOutput()
 	return err
 }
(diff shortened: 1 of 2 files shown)


integration-tests/itest/cli/cli.go
```diff
@@ -176,16 +190,51 @@ func GetStreamletPod(status *AppStatus, streamlet string) *StreamletPod {
...
+func GetPodsForStreamlet(app App, streamlet string) (pods []string, err error) {
+	status, err := Status(app)
+	if err != nil {
+		return nil, err
+	}
+	return GetPods(&status, streamlet), nil
+}
+
+// PollUntilExpectedPodsForStreamlet retrieves the pods that correspond to the given streamlet when
+// the pod count is or becomes equal to the expected count.
```
RayRoestenburg:
```suggestion
// PollUntilExpectedPodsForStreamlet polls the pods that correspond to the given streamlet until 
// the pod count becomes equal to the expected count.
```

integration-tests/itest/deploy_test.go
```diff
@@ -125,6 +126,53 @@ var _ = Describe("Application deployment", func() {
 
 	})
 
+	Context("A deployed streamlet can be scaled", func() {
+		// A function that calculates the streamlet scale based on pod count
+		type scalePodCorrection func(scale int) int
+		var noCorrection scalePodCorrection = func(scale int) int { return scale }
+		var coordinatorCorrection scalePodCorrection = func(scale int) int { return scale - 1 }
+		scaleCheck := func(streamlet string, scaleCorrection scalePodCorrection) {
```
RayRoestenburg: Could have done scaleCorrection as an int instead, fill in -1 or 0, no need for func.

maasg: this is more to give it a clear name that explains why it's there. i.e. to show the intention.

integration-tests/itest/cli/cli.go
maasg: Thanks for the suggestion. I find that "retrieves" better describes the functionality of the method because it gives a return value.

RayRoestenburg: The more important part is that this method *keeps polling* until it gets to the expected count.
I found this sentence a little strange:
> retrieves the pods that correspond to the given streamlet when the pod count is or becomes equal to the expected count.
Note the `is or becomes equal to ` part.  

I think the comment would be clearer as:

> PollUntilExpectedPodsForStreamlet continues to retrieve pods that correspond to the given streamlet until the pod count is equal to the expected count.

Or something more to that effect.

integration-tests/itest/deploy_test.go
RayRoestenburg: you could just call the int variable `scalePodCorrection`? that is a clear enough name, right?

integration-tests/itest/cli/cli.go
maasg: I see where the confusion is: 
I'm using "retrieves" to convey "the value it returns" while you are reading it as the action it's doing internally. 
I see the ambiguity in this. I replaced 'retrieves' with 'returns' and added a note about polling. 
I hope it's clear now.

maasg:
@RayRoestenburg All comments have been addressed. - merging this to continue on the feature branch.

maasg:
@RayRoestenburg @olofwalker thanks for the reviews!

PR #202, opened 2020-03-25
Fix for pending status bug

RayRoestenburg:

Fix for pending status bug.

- Calculating the appStatus is moved from kubectl cloudflow to the cloudflow operator. This solves the issue of not knowing how many pods should run when replicas is not set (and defaults are used).
- An `expectedPodCount` is now kept per StreamletStatus. The CloudflowApplication.Status is updated with updateApp, updatePod and deletePod methods. In updateApp, existing PodStatus lists are kept from the previous status that are still relevant, expectedPodCounts are updated. 
- The Status is also updated on CloudflowApplication event, if it already had a status. This is because it can happen that a user scales the app, which must change expectedPodCounts, while no pod watch events occur for a while, which is where normally the updateApp is called. In this in-between time the status should be reported as Pending.

- The logic for calculating PodStatus was incorrect. I've checked in Kubernetes source code and docs, to make sure we have a good solution.
- the PodStatus now reports nrOfContainers and nrOfContainersReady, to show the same output as get pods, 1/1 for one running container, one container expected to run.
- A pod with metadata.deletionTimestamp set is terminating.
- A pod status Phase, if set, should be reported. If it is Pending, or not set, the container states might have more detailed info, which is reported.

- [x] unit tests
- [x] On GKE with taxi-ride app (Akka, Flink)
- [x] On GKE with call-record-aggregator (Akka, Spark)

Note:
New fields: `expectedPodCount` is added to `StreamletStatus`, PodStatus now has `nrOfContainersReady` and `nrOfContainers`. These are added as optional fields to not crash the operator on existing apps that do not have these fields.


core/cloudflow-operator/src/main/scala/cloudflow/operator/CloudflowApplication.scala
```diff
@@ -122,30 +122,75 @@ object CloudflowApplication {
...
+        Status.Pending
+      }
   }
 
-  case class Status(
-      appId: String,
-      appVersion: String,
-      streamletStatuses: Vector[StreamletStatus]
-  ) {
+  // the status is created with the expected number of streamlet statuses, derived from the CloudflowApplication.Spec, see companion
```
olofwalker: So this takes into account the pods also created by the spark operator that has no underlying deployment (if I remember correctly, that is how it works with spark drivers/executors)?

core/cloudflow-operator/src/main/scala/cloudflow/operator/CloudflowApplication.scala
```diff
@@ -122,30 +122,75 @@ object CloudflowApplication {
...
+        streamletStatuses,
+        Unknown
       )
+    }
+
+    def createStreamletStatuses(spec: CloudflowApplication.Spec) = {
+      // TODO this code needs to be improved.
+      // There are some plans to make replicas mandatory
+      // and to indicate extraPods required in the deployment to prevent this issue
+      import cloudflow.operator.runner._
```
olofwalker: Does this take into account the current scaling factor?

core/cloudflow-operator/src/main/scala/cloudflow/operator/CloudflowApplication.scala
```diff
@@ -122,30 +122,75 @@ object CloudflowApplication {
...
+      }
   }
 
-  case class Status(
-      appId: String,
-      appVersion: String,
-      streamletStatuses: Vector[StreamletStatus]
-  ) {
+  // the status is created with the expected number of streamlet statuses, derived from the CloudflowApplication.Spec, see companion
+  case class Status private (appId: String, appVersion: String, streamletStatuses: Vector[StreamletStatus], appStatus: String) {
```
debasishg: enum (`case object`) for `appStatus` ?

core/cloudflow-operator/src/main/scala/cloudflow/operator/CloudflowApplication.scala
```diff
@@ -200,28 +261,29 @@ object CloudflowApplication {
...
     }
   }
 
   /**
    * Status of the pod.
    * ready can be "True", "False" or "Unknown"
    */
-  case class PodStatus(name: String, status: String, restarts: Int, ready: String)
-
+  case class PodStatus(name: String, status: String, restarts: Int, ready: String) {
```
debasishg: Can't we use `ready` from `object PodStatus` without making it a `String` ?

core/cloudflow-operator/src/main/scala/cloudflow/operator/CloudflowApplication.scala
RayRoestenburg: It takes into account the replicas which is the scaling factor.

core/cloudflow-operator/src/main/scala/cloudflow/operator/CloudflowApplication.scala
RayRoestenburg: complicates the json format, and it is just the result that must be shown.

core/cloudflow-operator/src/main/scala/cloudflow/operator/CloudflowApplication.scala
RayRoestenburg: there are always pods.

core/cloudflow-operator/src/main/scala/cloudflow/operator/CloudflowApplication.scala
RayRoestenburg: It must be deprecated.

RayRoestenburg: (or removed, but older kubectl cloudflow will need it)

core/cloudflow-operator/src/main/scala/cloudflow/operator/CloudflowApplication.scala
```diff
@@ -122,30 +122,71 @@ object CloudflowApplication {
...
+    def createStreamletStatuses(spec: CloudflowApplication.Spec) = {
+      // TODO not match on runtime, this is not great for extensibility.
+      // There are some plans to make replicas mandatory in the CR
+      // and to indicate extraPods required in the deployment to prevent the code below specific to runtimes
+      import cloudflow.operator.runner._
+      spec.deployments.map { deployment =>
+        val expectedPodCount = deployment.runtime match {
+          case AkkaRunner.runtime  ⇒ deployment.replicas.getOrElse(AkkaRunner.DefaultReplicas)
+          case SparkRunner.runtime ⇒ deployment.replicas.getOrElse(SparkRunner.DefaultNrOfExecutorInstances) + 1
+          case FlinkRunner.runtime ⇒ deployment.replicas.getOrElse(FlinkRunner.DefaultReplicas) + 1
```
maasg: I think that we would rather push this to the `Runner` itf rather than fixing it here.

RayRoestenburg: It's a temporary fix we'll handle these defaults completely differently soon.

core/cloudflow-operator/src/main/scala/cloudflow/operator/CloudflowApplication.scala
```diff
@@ -186,42 +234,76 @@ object CloudflowApplication {
...
+    val Running          = "Running"
+    val Terminating      = "Terminating"
+    val Terminated       = "Terminated"
+    val Succeeded        = "Succeeded"
+    val Failed           = "Failed"
+    val Unknown          = "Unknown"
+    val CrashLoopBackOff = "CrashLoopBackOff"
+
+    val ReadyTrue  = "True"
+    val ReadyFalse = "False"
```
debasishg: couldn't we have done with a single `val Ready = "False"` or something booleanish ?

RayRoestenburg: These are constants used to set ready to true or false as a string value. (terrible, but this is how it was done, so need to keep that around)

core/cloudflow-operator/src/main/scala/cloudflow/operator/CloudflowApplication.scala
```diff
@@ -154,30 +195,47 @@ object CloudflowApplication {
...
   }
 
-  case class StreamletStatus(streamletName: String, podStatuses: Vector[PodStatus] = Vector()) {
+  case class StreamletStatus(
+      streamletName: String,
+      // Added as Option for backwards compatibility
+      expectedPodCount: Option[Int],
+      podStatuses: Vector[PodStatus]
+  ) {
+    def hasExpectedPods(nrOfPodsDetected: Int) = expectedPodCount.getOrElse(0) == nrOfPodsDetected
```
maasg: How can `expectedPodCount` be `None` - I would think that the expected count is always known, no? (while `actual` might be unknown.)

RayRoestenburg: @maasg The issue is that existing apps don't have this field. If I make it mandatory the operator will crash. (cause it can't read the status without the field). So this is a best of worst cases, instead of the operator crashing on existing apps, it will badly report the status of those apps. We'll document this in the release and tell users to redeploy. If they don't read those notes their stuff will at least not crash.

RayRoestenburg: @maasg the same applies for nrOfContainers and nrOfContainersReady in PodStatus.. Welcome to the world of backwards compat by optional fields ;-)

core/cloudflow-operator/src/main/scala/cloudflow/operator/CloudflowApplication.scala
```diff
@@ -186,42 +244,89 @@ object CloudflowApplication {
...
+      // Optional for backwards compatibility
+      nrOfContainersReady: Option[Int] = None,
+      // Optional for backwards compatibility
+      nrOfContainers: Option[Int] = None,
+      // TODO can this be removed without breaking backwards-compatibility?
+      ready: Option[String] = None
+  ) {
+    def containers      = nrOfContainers.getOrElse(0)
+    def containersReady = nrOfContainersReady.getOrElse(0)
+    def isReady         = status == PodStatus.Running && containersReady == containers && containers != 0
```
maasg:
```suggestion
    def isReady         = containers > 0 && status == PodStatus.Running && containersReady == containers
```

kubectl-cloudflow/cmd/update_docker_registry_credentials.go
```diff
@@ -158,23 +165,22 @@ func createOrUpdateServiceAccount(k8sClient *kubernetes.Clientset, appID string,
 	return nil, fmt.Errorf("failed to update Cloudflow app service account in `%s`, %s", appID, err)
 }
 
-func newCloudflowServiceAccount(appID string, cloudflowOperatorOwnerReference metav1.OwnerReference) v1.ServiceAccount {
+func newCloudflowServiceAccount(appID string) v1.ServiceAccount {
```
maasg: is this a piggy-back from the owner-ref issue?  It doesn't look like it's related to the status bug.

core/cloudflow-operator/src/main/scala/cloudflow/operator/CloudflowApplication.scala
```diff
@@ -186,42 +244,89 @@ object CloudflowApplication {
...
-            .getOrElse("Unknown")
-
-          PodStatus(name, st, restarts, ready)
-        }
-        .getOrElse(PodStatus(name, "Unknown", 0, ""))
+      // See https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/
+      val name                = pod.metadata.name
+      val status              = pod.status.getOrElse(Pod.Status())
+      val nrOfContainers      = pod.spec.map(_.containers.size).getOrElse(status.containerStatuses.size)
+      val nrOfContainersReady = status.containerStatuses.filter(_.ready == true).size
```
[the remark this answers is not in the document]
RayRoestenburg: lol 👍

kubectl-cloudflow/cmd/update_docker_registry_credentials.go
RayRoestenburg: yes I made this change on top of the owner-ref issue, strange that this pops up though... Let me double check.

core/cloudflow-operator/src/main/scala/cloudflow/operator/CloudflowApplication.scala
RayRoestenburg: I fixed this too, out of date somehow so could not just commit it from github.

kubectl-cloudflow/cmd/update_docker_registry_credentials.go
RayRoestenburg: It's outdated, I'm doing this on top of the fix for owner-ref issue. (It is in no commit in this PR)

PR #211, opened 2020-03-31
Fix for unexpected undeploy of applications.

RayRoestenburg:

The code was setting a bad owner reference, to the cloudflow operator deployment, which resides in a different namespace. This is not allowed.
- Removed cross namespace owner reference to cloudflow operator deployment.
- service account created now refers to cloudflow application as owner
- cloudflow application needs no owner. When CloudflowApplication CRD is deleted, all CloudflowApplications are deleted, the cloudflow operator cleans up everything.

Deploying the same app would undeploy the app. The owner reference was set incorrectly causing K8s to GC the app.

Bugfix: No unexpected undeploy of apps will occur anymore.

On GKE.


kubectl-cloudflow/cmd/deploy.go
```diff
@@ -171,21 +165,25 @@ func (opts *deployOptions) deployImpl(cmd *cobra.Command, args []string) {
...
+	if _, err := createOrUpdateServiceAccount(k8sClient, namespace, serviceAccount, ownerReference); err != nil {
+		util.LogAndExit("%s", err)
+	}
 
 	streamletNameSecretMap = deploy.UpdateSecretsWithOwnerReference(ownerReference, streamletNameSecretMap)
 	createOrUpdateStreamletSecrets(k8sClient, namespace, streamletNameSecretMap)
 
 	serviceAccount.ObjectMeta.OwnerReferences = []metav1.OwnerReference{ownerReference}
-	if _, err := createOrUpdateServiceAccount(k8sClient, namespace, serviceAccount); err != nil {
+	if _, err := createOrUpdateServiceAccount(k8sClient, namespace, serviceAccount, ownerReference); err != nil {
```
skyluc: `createOrUpdateServiceAccount` is called twice with the same parameters, which are in the same state.

kubectl-cloudflow/cmd/deploy.go
```diff
@@ -171,21 +165,25 @@ func (opts *deployOptions) deployImpl(cmd *cobra.Command, args []string) {
...
+	ownerReference := storedCR.GenerateOwnerReference()
+
+	if _, err := createOrUpdateServiceAccount(k8sClient, namespace, serviceAccount, ownerReference); err != nil {
+		util.LogAndExit("%s", err)
+	}
 
 	streamletNameSecretMap = deploy.UpdateSecretsWithOwnerReference(ownerReference, streamletNameSecretMap)
 	createOrUpdateStreamletSecrets(k8sClient, namespace, streamletNameSecretMap)
 
 	serviceAccount.ObjectMeta.OwnerReferences = []metav1.OwnerReference{ownerReference}
```
skyluc: This line of code has been added in `createOrUpdateServiceAccount`. It should be remove from here.

kubectl-cloudflow/cmd/deploy.go
```diff
@@ -153,14 +153,8 @@ func (opts *deployOptions) deployImpl(cmd *cobra.Command, args []string) {
...
-	// Get the Cloudflow operator ownerReference
-	ownerReference := version.GetOwnerReferenceForCloudflowOperator()
-
-	serviceAccount := newCloudflowServiceAccountWithImagePullSecrets(namespace, ownerReference)
-
-	if _, err := createOrUpdateServiceAccount(k8sClient, namespace, serviceAccount); err != nil {
-		util.LogAndExit("%s", err)
-	}
+	// Delay the creation of the service account for after the ownerReferences has been generated.
+	serviceAccount := newCloudflowServiceAccountWithImagePullSecrets(namespace)
```
skyluc: The creation of `serviceAccount` and `streamletNameSecretMap` can be moved later in the code, where they are actually used.

kubectl-cloudflow/cmd/deploy.go
RayRoestenburg: Oh  that's weird, let me fix that. Thanks for spotting

kubectl-cloudflow/cmd/deploy.go
RayRoestenburg: Fixed.

kubectl-cloudflow/cmd/deploy.go
RayRoestenburg: Fixed.

kubectl-cloudflow/cmd/deploy.go
RayRoestenburg: Fixed.

RayRoestenburg:
> @maasg LGTM, but I'm not fully aware of the work involved in the owner refs.
> How should we test this from the outside? Is there an e2e test possible.
> e.g. add owner ref checks to the deploy test?

I've just tested this manually.
What I've done is repeatedly deploy the same app which would cause the undeploy problem. The problem did not occur anymore. Also, I've made sure that deleting the CRD deletes all the apps and undeploys these correctly, removing all the resources as it should.

agemooij:
@RayRoestenburg did you test an explicit command line undeploy to see whether the full application, minus perhaps its Kafka topics, is removed?

RayRoestenburg:
@agemooij Good point! I have not changed the operator, which I now see, also on its own fetches the owner ref of the cloudflow operator deployment.  I'll have to change that still. I don't think this caused issues, but you never know.

PR #216, opened 2020-04-01
increase timeout to patch this for now

maasg:
Increases the timeout of this test to allow for the processing to happen without stopping the test.
The root cause goes much deeper and a complete fix will follow, but this will prevent CI from failing for the moment.


maasg:
@olofwalker This is a borderline case. We only need to run the query once and wait for it to finish but the timeout condition we use for the general "wait for x millis" triggers before we complete the operation, leading to a failure of the test.
The 'full solution' is to separate these two cases cleanly, which requires some refactoring of the code. During that refactoring, I found a deeper issue on the streamlet query health check, so before I get lost in the rabbit hole, I wanted to unblock potential failures in CI.

RayRoestenburg:
Maybe a good idea to use a "dilated" version of times, for CI, and configure this differently there.

maasg:
@RayRoestenburg the timeout is not needed at all in this test. I'm working on a solution to get rid of it.

PR #217, opened 2020-04-01
Update app ref with rc5

maasg:
Updates the integration tests with the artifacts to run RC5 tests.


RayRoestenburg:
Checked that I could pull the image. LGTM

PR #219, opened 2020-04-02
This removes the node affinity from Strimzi.

olofwalker:


installer/create-cluster-gke.sh
```diff
@@ -52,30 +52,17 @@ if [ -z "$CLUSTER_VERSION" ]
 fi
 
 # Create cluster
-
 gcloud container clusters create $CLUSTER_NAME \
   --cluster-version $CLUSTER_VERSION  \
   --image-type cos \
   --machine-type n1-standard-4 \
-  --num-nodes 3 \
+  --num-nodes 5 \
```
agemooij, line 59: Shouldn't this be 6? We removed a three-node second node pool. Are we sure 5 nodes is enough?

olofwalker: It's enough from running the swiss-knife application, but I can up it to 6 if you want

agemooij: OK, that should be good then.

PR #222, opened 2020-04-02
Csp 1978 + 1987 - spark test method based on data availability

maasg:
Introduces a new method to test Spark Streamlets where a query monitor checks for the availability of data instead of 'brute-force' waits that are tricky to time correctly on different environments.

Now, it's possible to get information about the rows processed. Also note how the API doesn't depend on a wait time anymore:

Before:
```
testKit.run(instance, Seq(in0, in1, in2), Seq(out), 10.seconds)
```
Now: 
```
val run = testKit.run(instance, Seq(in0, in1, in2), Seq(out))
...
run.totalRows must be(30)
```
This new method should be resilient against different CPU timings as the test progresses only when data becomes available (within a certain delay tolerance)

CSP-1987: In this PR, there's also a fix to the monitoring logic that checks for failed streamlets at runtime. Now, failures are correctly monitored and will trigger an immediate downing of the failed streamlet.
See `SparkStreamlet` for more details.

### DIFF
--- a/core/cloudflow-runner/src/main/scala/cloudflow/runner/Runner.scala
+++ b/core/cloudflow-runner/src/main/scala/cloudflow/runner/Runner.scala
@@ -37,7 +37,6 @@ object Runner extends RunnerConfigResolver with StreamletLoader {
     case Some(os) if os.startsWith("Win") ⇒
       log.error("cloudflow.runner.Runner is NOT compatible with Windows!!")
     case None ⇒ log.warn("""sys.props.get("os.name") returned None!""")
-    case _    ⇒ // okay
   }
 
   val PVCMountPath: String               = "/mnt/spark/storage"
@@ -68,17 +67,20 @@ object Runner extends RunnerConfigResolver with StreamletLoader {
         val streamletExecution = loadedStreamlet.streamlet.run(withPodRuntimeConfig)
         loadedStreamlet.streamlet.logStartRunnerMessage(formatBuildInfo)
 
-        Try {
-          // the runner waits for the execution to complete
-          // In normal circumstances it will run forever for streaming data source unless
-          // being stopped forcibly or any of the queries faces an exception
-          Await.result(streamletExecution.completed, Duration.Inf)
-        } match {
-          case Success(_) ⇒ System.exit(0)
-          case Failure(ex @ ExceptionAcc(exceptions)) ⇒
+        // the runner waits for the execution to complete
+        // In normal circumstances it will run forever for streaming data source unless
+        // being stopped forcibly or any of the queries faces an exception
+        Await.ready(streamletExecution.completed, Duration.Inf)
+        streamletExecution.completed.value match {
+          case None => // can't happen b/c we wait for it to be ready
+            log.error("Unexpected streamletExecution state: No result after termination. This is most probably a bug.")
+            shutdownWithFailure(loadedStreamlet,
+                                new IllegalStateException("Unexpected streamletExecution state: No result after termination"))
+          case Some(Success(_)) ⇒ System.exit(0)
+          case Some(Failure(ex @ ExceptionAcc(exceptions))) ⇒
             exceptions.foreach(ErrorEvents.report(loadedStreamlet, withPodRuntimeConfig, _))
             shutdownWithFailure(loadedStreamlet, ex)
-          case Failure(ex) ⇒
+          case Some(Failure(ex)) ⇒
             ErrorEvents.report(loadedStreamlet, withPodRuntimeConfig, ex)
             shutdownWithFailure(loadedStreamlet, ex)
         }
--- a/core/cloudflow-runner/src/main/scala/cloudflow/runner/Runner.scala
+++ b/core/cloudflow-runner/src/main/scala/cloudflow/runner/Runner.scala
@@ -32,6 +32,7 @@ import cloudflow.events.errors.ErrorEvents
  */
 object Runner extends RunnerConfigResolver with StreamletLoader {
   lazy val log = LoggerFactory.getLogger(getClass.getName)
+  import scala.concurrent.ExecutionContext.Implicits.global
 
   sys.props.get("os.name") match {
     case Some(os) if os.startsWith("Win") ⇒
@@ -71,16 +72,14 @@ object Runner extends RunnerConfigResolver with StreamletLoader {
         // the runner waits for the execution to complete
         // In normal circumstances it will run forever for streaming data source unless
         // being stopped forcibly or any of the queries faces an exception
-        Await.ready(streamletExecution.completed, Duration.Inf)
-        streamletExecution.completed.value match {
-          case None => // can't happen b/c we wait for it to be ready
-            log.error("Unexpected streamletExecution state: No result after termination. This is most probably a bug.")
-            shutdown(loadedStreamlet, Some(new IllegalStateException("Unexpected streamletExecution state: No result after termination")))
-          case Some(Success(_)) ⇒ shutdown(loadedStreamlet)
-          case Some(Failure(ex @ ExceptionAcc(exceptions))) ⇒
+        try {
+          Await.result(streamletExecution.completed, Duration.Inf)
+          shutdown(loadedStreamlet)
+        } catch {
+          case ex @ ExceptionAcc(exceptions) ⇒
             exceptions.foreach(ErrorEvents.report(loadedStreamlet, withPodRuntimeConfig, _))
             shutdown(loadedStreamlet, Some(ex))
-          case Some(Failure(ex)) ⇒
+          case ex =>
             ErrorEvents.report(loadedStreamlet, withPodRuntimeConfig, ex)
             shutdown(loadedStreamlet, Some(ex))
         }
@@ -93,16 +92,14 @@ object Runner extends RunnerConfigResolver with StreamletLoader {
     Files.deleteIfExists(
       Paths.get(s"/tmp/${loadedStreamlet.config.streamletRef}.txt")
     )
-    maybeException
-      .map { ex =>
+    maybeException match {
+      case Some(ex) =>
         log.error("Fatal error has occurred:", ex)
         System.exit(-1)
-      }
-      .getOrElse {
-        log.info("Streamlet terminating due to successful termination of query:")
+      case None =>
+        log.info("Streamlet terminating without failure")
         System.exit(0)
-      }
-
+    }
   }
 
   private def formatBuildInfo: String = {
--- a/core/cloudflow-runner/src/main/scala/cloudflow/runner/Runner.scala
+++ b/core/cloudflow-runner/src/main/scala/cloudflow/runner/Runner.scala
@@ -94,7 +94,7 @@ object Runner extends RunnerConfigResolver with StreamletLoader {
     )
     maybeException match {
       case Some(ex) =>
-        log.error("Fatal error has occurred:", ex)
+        log.error("A fatal error has occurred. The streamlet is going to shutdown", ex)
         System.exit(-1)
       case None =>
         log.info("Streamlet terminating without failure")
--- a/core/cloudflow-runner/src/main/scala/cloudflow/runner/Runner.scala
+++ b/core/cloudflow-runner/src/main/scala/cloudflow/runner/Runner.scala
@@ -36,7 +36,8 @@ object Runner extends RunnerConfigResolver with StreamletLoader {
   sys.props.get("os.name") match {
     case Some(os) if os.startsWith("Win") ⇒
       log.error("cloudflow.runner.Runner is NOT compatible with Windows!!")
-    case None ⇒ log.warn("""sys.props.get("os.name") returned None!""")
+    case Some(os) => log.info(s"Runner running on $os")
+    case None     ⇒ log.warn("""sys.props.get("os.name") returned None!""")
   }
 
   val PVCMountPath: String               = "/mnt/spark/storage"
--- a/core/cloudflow-runner/src/main/scala/cloudflow/runner/Runner.scala
+++ b/core/cloudflow-runner/src/main/scala/cloudflow/runner/Runner.scala
@@ -74,28 +74,34 @@ object Runner extends RunnerConfigResolver with StreamletLoader {
         streamletExecution.completed.value match {
           case None => // can't happen b/c we wait for it to be ready
             log.error("Unexpected streamletExecution state: No result after termination. This is most probably a bug.")
-            shutdownWithFailure(loadedStreamlet,
-                                new IllegalStateException("Unexpected streamletExecution state: No result after termination"))
-          case Some(Success(_)) ⇒ System.exit(0)
+            shutdown(loadedStreamlet, Some(new IllegalStateException("Unexpected streamletExecution state: No result after termination")))
+          case Some(Success(_)) ⇒ shutdown(loadedStreamlet)
           case Some(Failure(ex @ ExceptionAcc(exceptions))) ⇒
             exceptions.foreach(ErrorEvents.report(loadedStreamlet, withPodRuntimeConfig, _))
-            shutdownWithFailure(loadedStreamlet, ex)
+            shutdown(loadedStreamlet, Some(ex))
           case Some(Failure(ex)) ⇒
             ErrorEvents.report(loadedStreamlet, withPodRuntimeConfig, ex)
-            shutdownWithFailure(loadedStreamlet, ex)
+            shutdown(loadedStreamlet, Some(ex))
         }
       case Failure(ex) ⇒ throw new Exception(ex)
     }
   }
 
-  private def shutdownWithFailure(loadedStreamlet: LoadedStreamlet, ex: Throwable) = {
+  private def shutdown(loadedStreamlet: LoadedStreamlet, maybeException: Option[Throwable] = None) = {
     // we created this file when the pod started running (see AkkaStreamlet#run)
     Files.deleteIfExists(
       Paths.get(s"/tmp/${loadedStreamlet.config.streamletRef}.txt")
     )
-    log.error("Fatal error has occurred:", ex)
+    maybeException
+      .map { ex =>
+        log.error("Fatal error has occurred:", ex)
+        System.exit(-1)
+      }
+      .getOrElse {
+        log.info("Streamlet terminating due to successful termination of query:")
+        System.exit(0)
+      }
 
-    System.exit(-1)
   }
 
   private def formatBuildInfo: String = {
--- a/core/cloudflow-spark-testkit/src/main/scala/cloudflow/spark/testkit/SparkStreamletTestkit.scala
+++ b/core/cloudflow-spark-testkit/src/main/scala/cloudflow/spark/testkit/SparkStreamletTestkit.scala
@@ -84,7 +84,7 @@ final case class ConfigParameterValue private (configParameterKey: String, value
  *      in.addData(data)
  *
  *      // 7. Run the test
- *      run(processor, Seq(in), Seq(out), 2.seconds)
+ *      run(processor, Seq(in), Seq(out))
  *
  *      // 8. Get data from outlet tap(s) and assert
  *      val results = out.asCollection(session)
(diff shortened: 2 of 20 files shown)


core/cloudflow-spark-testkit/src/main/scala/cloudflow/spark/testkit/SparkStreamletTestkit.scala
```diff
@@ -212,32 +218,27 @@ final case class SparkStreamletTestkit(session: SparkSession, config: Config = C
...
-      sparkStreamlet: SparkStreamlet,
-      duration: Duration
+      sparkStreamlet: SparkStreamlet
   ): Unit = {
-    val t0             = System.currentTimeMillis()
-    def elapsedMillis  = System.currentTimeMillis() - t0
+    val t0           = System.currentTimeMillis()
+    def lap          = System.currentTimeMillis() - t0
+    val queryMonitor = new QueryExecutionMonitor()
+    ctx.session.streams.addListener(queryMonitor)
```
debasishg, line 232: 👍

core/cloudflow-spark-testkit/src/main/scala/cloudflow/spark/testkit/SparkStreamletTestkit.scala
```diff
@@ -256,3 +257,55 @@ case class SparkOutletTap[T: Encoder](
...
+  }
+
+  def waitForData(): Future[Long] = {
+    def hasData   = dataRows.nonEmpty && dataRows.forall { case (_, rows) => rows > 0 }
+    def totalRows = dataRows.values.sum
+    Future {
+      while (!hasData) {
+        Thread.sleep(1.second.toMillis)
+      }
+      totalRows
```
debasishg, line 286: How do we ensure here that we have all the rows that need to be fetched ? Should `waitForData` take an argument `expected_no_of_rows` so that it can match ?

debasishg: In testkit we always will be dealing with a finite data set - right ?

core/cloudflow-spark/src/main/scala/cloudflow/spark/SparkStreamlet.scala
```diff
@@ -84,59 +79,51 @@ trait SparkStreamlet extends Streamlet[SparkStreamletContext] with Serializable
...
+        Future {
+          val t0 = System.currentTimeMillis()
+          // Step 1: Wait for all queries to stop
+          while (streamletQueryExecution.queries.exists(_.isActive)) {
+            val elapsedTime = System.currentTimeMillis() - t0
+            if (elapsedTime < StopTimeout) {
+              Thread.sleep(1.second.toMillis)
+            } else {
+              throw new QueryTerminationTimeout(streamletQueryExecution.queries.filter(_.isActive), elapsedTime)
+            }
```
debasishg, line 116: Can we replace this with something like https://gist.github.com/viktorklang/9414163 .. implement a retry with backoff semantics ?

core/cloudflow-spark-testkit/src/main/scala/cloudflow/spark/testkit/SparkStreamletTestkit.scala
```diff
@@ -107,7 +113,7 @@ final case class SparkStreamletTestkit(session: SparkSession, config: Config = C
   def withConfigParameterValues(configParameterValues: ConfigParameterValue*): SparkStreamletTestkit = {
     val parameterValueConfig = ConfigFactory.parseString(
       configParameterValues
-        .map(parameterValue ⇒ s"cloudflow.streamlets.streamlet-under-test.${parameterValue.configParameterKey} = ${parameterValue.value}")
+        .map(parameterValue ⇒ s"cloudflow.streamlets.$TestStreamletName.${parameterValue.configParameterKey} = ${parameterValue.value}")
```
RayRoestenburg, line 116: 👍

core/cloudflow-spark-testkit/src/main/scala/cloudflow/spark/testkit/SparkStreamletTestkit.scala
```diff
@@ -212,32 +218,27 @@ final case class SparkStreamletTestkit(session: SparkSession, config: Config = C
...
-      duration: Duration
+      sparkStreamlet: SparkStreamlet
   ): Unit = {
-    val t0             = System.currentTimeMillis()
-    def elapsedMillis  = System.currentTimeMillis() - t0
+    val t0           = System.currentTimeMillis()
+    def lap          = System.currentTimeMillis() - t0
+    val queryMonitor = new QueryExecutionMonitor()
+    ctx.session.streams.addListener(queryMonitor)
+    println("============================================== execution start =============" + lap)
```
RayRoestenburg, line 233: I think these printlns should be removed, it is not very common to force print something during a test, especially if users can't turn it off. (End-users also use the testkit of course)

RayRoestenburg: (maybe they are there for some debugging / testing?)

maasg: :-)
To quote myself from the PR description:
>Note:
>This Draft PR still contains plenty of heisen(de)bug output to observe how it behaves in the >CI machinery.

core/cloudflow-spark-testkit/src/main/scala/cloudflow/spark/testkit/SparkStreamletTestkit.scala
```diff
@@ -256,3 +257,55 @@ case class SparkOutletTap[T: Encoder](
...
+    val exceptions = status.values.collect { case Terminated(Some(reason)) => reason }
+    ExecutionReport(dataRows.values.sum, dataRows.keys.size, exceptions.toSeq)
+  }
+
+  def waitForData(): Future[Long] = {
+    def hasData   = dataRows.nonEmpty && dataRows.forall { case (_, rows) => rows > 0 }
+    def totalRows = dataRows.values.sum
+    Future {
+      while (!hasData) {
+        Thread.sleep(1.second.toMillis)
```
RayRoestenburg, line 284: Better to use promise and future here ( create a promise, complete it when data arrives, await on future of promise here, no need to sleep)

core/cloudflow-spark-testkit/src/main/scala/cloudflow/spark/testkit/SparkStreamletTestkit.scala
RayRoestenburg: I don't think we can say it is always finite. It depends on how the end user uses it.

core/cloudflow-spark-testkit/src/main/scala/cloudflow/spark/testkit/TestSparkStreamletContext.scala
```diff
@@ -52,16 +53,18 @@ private[testkit] class TestSparkStreamletContext(override val streamletRef: Stri
       .map(_.instream.asInstanceOf[MemoryStream[In]].toDF.as[In])
       .getOrElse(throw TestContextException(inPort.name, s"Bad test context, could not find source for inlet ${inPort.name}"))
 
+  // the test write stream implementation is blocking and returns once the query has produced a result.
```
RayRoestenburg, line 56: Can we do this without changing this to blocking?

core/cloudflow-spark/src/main/scala/cloudflow/spark/SparkStreamlet.scala
RayRoestenburg: It's on `stop` so it's not the most critical path, but yes, I agree it can be polished a bit and use some async-fu ;-)

core/cloudflow-spark-testkit/src/main/scala/cloudflow/spark/testkit/SparkStreamletTestkit.scala
maasg: Good point! Thanks.

core/cloudflow-spark-testkit/src/main/scala/cloudflow/spark/testkit/SparkStreamletTestkit.scala
RayRoestenburg: yeah apologies I obviously can't read text, only code 🤦‍♂

core/cloudflow-spark-testkit/src/main/scala/cloudflow/spark/testkit/SparkStreamletTestkit.scala
maasg: We have existing test cases using infinite data sources (`rate` source)

core/cloudflow-spark-testkit/src/main/scala/cloudflow/spark/testkit/SparkStreamletTestkit.scala
maasg: I created a `poll` for our prod code, but I find it's overkill here.
See also: https://github.com/lightbend/cloudflow/pull/222/files#diff-80785b7e8093f0a01524d8c1768e4347R126

core/cloudflow-runner/src/main/scala/cloudflow/runner/Runner.scala
```diff
@@ -68,17 +67,20 @@ object Runner extends RunnerConfigResolver with StreamletLoader {
...
+        // the runner waits for the execution to complete
+        // In normal circumstances it will run forever for streaming data source unless
+        // being stopped forcibly or any of the queries faces an exception
+        Await.ready(streamletExecution.completed, Duration.Inf)
+        streamletExecution.completed.value match {
+          case None => // can't happen b/c we wait for it to be ready
+            log.error("Unexpected streamletExecution state: No result after termination. This is most probably a bug.")
+            shutdownWithFailure(loadedStreamlet,
+                                new IllegalStateException("Unexpected streamletExecution state: No result after termination"))
+          case Some(Success(_)) ⇒ System.exit(0)
```
RayRoestenburg, line 79: Maybe add a log that it is shutting down? Or is that already done by this point?

core/cloudflow-runner/src/main/scala/cloudflow/runner/Runner.scala
```diff
@@ -68,17 +67,20 @@ object Runner extends RunnerConfigResolver with StreamletLoader {
...
-          // In normal circumstances it will run forever for streaming data source unless
-          // being stopped forcibly or any of the queries faces an exception
-          Await.result(streamletExecution.completed, Duration.Inf)
-        } match {
-          case Success(_) ⇒ System.exit(0)
-          case Failure(ex @ ExceptionAcc(exceptions)) ⇒
+        // the runner waits for the execution to complete
+        // In normal circumstances it will run forever for streaming data source unless
+        // being stopped forcibly or any of the queries faces an exception
+        Await.ready(streamletExecution.completed, Duration.Inf)
```
RayRoestenburg, line 73: Why not use Await.result, since you use the result a little further down?

core/cloudflow-runner/src/main/scala/cloudflow/runner/Runner.scala
```diff
@@ -68,17 +67,20 @@ object Runner extends RunnerConfigResolver with StreamletLoader {
...
+        streamletExecution.completed.value match {
```
RayRoestenburg, line 74: If you use `Await.result` you don't need this line?

core/cloudflow-spark-testkit/src/main/scala/cloudflow/spark/testkit/SparkStreamletTestkit.scala
RayRoestenburg: I don't understand really cause the hasData will flip once, so you can use promise future and no poll? Am I missing something?

core/cloudflow-runner/src/main/scala/cloudflow/runner/Runner.scala
maasg: If you use `result`, you will throw an exception when it terminates due to a failure. You'd need an additional `try...catch` like it was done before. Like this, we have a uniform result handling.

core/cloudflow-runner/src/main/scala/cloudflow/runner/Runner.scala
maasg: See note :arrow_down:

core/cloudflow-spark-testkit/src/main/scala/cloudflow/spark/testkit/SparkStreamletTestkit.scala
maasg: @RayRoestenburg one option I could think of is to move this check to `onQueryProgress` but that check code looks intrusive there as it mixes the work of the progress callback with the data availability reporting. 
It might be less obvious to maintain later.
Would you have a suggestion?

core/cloudflow-runner/src/main/scala/cloudflow/runner/Runner.scala
RayRoestenburg: Yes but in this way you are handling a case that will never happen otherwise (the None case).
Just feels weird to do a `Future.value`, (I never use it) and do this in two steps. Remove line 74 and everything still compiles.. Just feels wrong. my two cents.

core/cloudflow-runner/src/main/scala/cloudflow/runner/Runner.scala
```diff
@@ -68,32 +68,41 @@ object Runner extends RunnerConfigResolver with StreamletLoader {
...
       Paths.get(s"/tmp/${loadedStreamlet.config.streamletRef}.txt")
     )
-    log.error("Fatal error has occurred:", ex)
+    maybeException
+      .map { ex =>
+        log.error("Fatal error has occurred:", ex)
+        System.exit(-1)
+      }
+      .getOrElse {
+        log.info("Streamlet terminating due to successful termination of query:")
```
RayRoestenburg, line 102: This message is specific to spark (query), should not be here? Also ends on colon?

core/cloudflow-runner/src/main/scala/cloudflow/runner/Runner.scala
```diff
@@ -68,32 +68,41 @@ object Runner extends RunnerConfigResolver with StreamletLoader {
...
-  private def shutdownWithFailure(loadedStreamlet: LoadedStreamlet, ex: Throwable) = {
+  private def shutdown(loadedStreamlet: LoadedStreamlet, maybeException: Option[Throwable] = None) = {
     // we created this file when the pod started running (see AkkaStreamlet#run)
     Files.deleteIfExists(
       Paths.get(s"/tmp/${loadedStreamlet.config.streamletRef}.txt")
     )
-    log.error("Fatal error has occurred:", ex)
+    maybeException
+      .map { ex =>
+        log.error("Fatal error has occurred:", ex)
```
RayRoestenburg, line 98: [minor] Would be nice to indicate that the streamlet will shut down.

core/cloudflow-spark-testkit/src/main/scala/cloudflow/spark/testkit/SparkStreamletTestkit.scala
```diff
@@ -129,25 +138,23 @@ final case class SparkStreamletTestkit(session: SparkSession, config: Config = C
   def outletAsTap[Out: Encoder](out: CodecOutlet[Out]) = out.outletAsTap
 
   /**
-   * Runs the `sparkStreamlet` using `inletTaps` as the sources and `outletTaps` as the sinks. Each `inletTap` abstracts
-   * a `MemoryStream` and an inlet port, where the test input data gets added. The `outletTap` returns a port and
-   * a query name, which gives a handle to the Spark `StreamingQuery` name that gets executed.
+   * Runs the `sparkStreamlet` using `inletTaps` as the sources and `outletTaps` as the sinks.
+   * Each `inletTap` abstracts a `MemoryStream` and an inlet port, where the test input data gets added.
```
RayRoestenburg, line 142: [minor] Would prefer to not use the word abstracts here

core/cloudflow-spark/src/main/scala/cloudflow/spark/SparkStreamlet.scala
```diff
@@ -84,60 +80,69 @@ trait SparkStreamlet extends Streamlet[SparkStreamletContext] with Serializable
...
-          } else {
-            // succeed the future and we can now go and stop the runner
-            // Note: we still now have no way to restart any query
-            val _ = completionPromise.trySuccess(Dun)
+      def completed: Future[Dun] = completionFuture
+      val ready: Future[Dun]     = Future.successful(Dun)
+      def stop(): Future[Dun] = {
+        streamletQueryExecution.stop()
+        scheduledQueryCheck.cancel()
+        poll(streamletQueryExecution.queries.forall(!_.isActive), 1.second, StopTimeout, system.scheduler)
```
RayRoestenburg, line 109: FYI there is a cancellable scheduler as well. you could pass in a promise into the  runnable and complete it with failure when it times out or complete with result if it succeeds. Might simplify this code a bit.

core/cloudflow-spark-testkit/src/main/scala/cloudflow/spark/testkit/SparkStreamletTestkit.scala
```diff
@@ -256,3 +248,44 @@ case class SparkOutletTap[T: Encoder](
...
+  def executionReport: ExecutionReport = {
+    val exceptions = status.values.collect { case Terminated(Some(reason)) => reason }
+    ExecutionReport(dataRows.values.sum, dataRows.keys.size, exceptions.toSeq)
+  }
+
+  def waitForData(): Future[Long] = {
+    def hasData   = dataRows.nonEmpty && dataRows.forall { case (_, rows) => rows > 0 }
+    def totalRows = dataRows.values.sum
+    Future {
+      while (!hasData) {
```
RayRoestenburg, line 270: Why do we wait only until the first bit of data comes in (every dataRow has more than 0 rows?)? Do I understand that correctly? I guess I have the same question as Debasish

core/cloudflow-spark-testkit/src/main/scala/cloudflow/spark/testkit/SparkStreamletTestkit.scala
RayRoestenburg: You can create the promise in the body of the QueryExecutionMonitor, pass the future of it here, complete promise on first data (probably in onQueryProgress)? I don't see the problem with that (yet)

core/cloudflow-spark-testkit/src/main/scala/cloudflow/spark/testkit/SparkStreamletTestkit.scala
```diff
@@ -256,3 +248,44 @@ case class SparkOutletTap[T: Encoder](
...
+
+  def executionReport: ExecutionReport = {
+    val exceptions = status.values.collect { case Terminated(Some(reason)) => reason }
+    ExecutionReport(dataRows.values.sum, dataRows.keys.size, exceptions.toSeq)
+  }
+
+  def waitForData(): Future[Long] = {
+    def hasData   = dataRows.nonEmpty && dataRows.forall { case (_, rows) => rows > 0 }
+    def totalRows = dataRows.values.sum
+    Future {
```
RayRoestenburg, line 269: Another issue with this approach is if the data never completes, this will never complete, and you have no way to make it time out.

core/cloudflow-spark/src/main/scala/cloudflow/spark/SparkStreamlet.scala
```diff
@@ -98,11 +100,18 @@ trait SparkStreamlet extends Streamlet[SparkStreamletContext] with Serializable
         if (someQueryStopped) completionPromise.completeWith(stop())
       }
 
+      val sysCheck: Cancellable = system.scheduler.schedule(1.minute, 1.minutes) {
+        val totalMem = Runtime.getRuntime.totalMemory() / 1024.0 / 1024.0
+        println(s"jvm-total-mem: $totalMem Mb")
```
RayRoestenburg, line 105: nitpick: `MB`

core/cloudflow-spark-testkit/src/main/scala/cloudflow/spark/testkit/SparkStreamletTestkit.scala
```diff
@@ -286,6 +281,9 @@ class QueryExecutionMonitor()(implicit ec: ExecutionContext) extends StreamingQu
     val id   = queryProgress.progress.id
     val rows = queryProgress.progress.numInputRows
     dataRows = dataRows + dataRows.get(id).map(v => id -> (v + rows)).getOrElse(id -> rows)
+    if (hasData & !dataAvailable.isCompleted) {
+      dataAvailable.success(dataRows.values.sum)
```
RayRoestenburg, line 285: you could use trySuccess if you are worried it will happen more than once. (success will throw exception if already completed)

core/cloudflow-spark-testkit/src/main/scala/cloudflow/spark/testkit/SparkStreamletTestkit.scala
```diff
@@ -224,11 +224,9 @@ final case class SparkStreamletTestkit(session: SparkSession, config: Config = C
     val queryMonitor = new QueryExecutionMonitor()
     ctx.session.streams.addListener(queryMonitor)
     val queryExecution = sparkStreamlet.setContext(ctx).run(ctx.config)
-    val res = for {
-      _    <- queryMonitor.waitForData()
-      done <- queryExecution.stop()
-    } yield done
-    Await.result(res, maxDuration)
+    val gotData        = queryMonitor.waitForData().andThen { case _ => queryExecution.stop() }
```
RayRoestenburg, line 227: 👍

core/cloudflow-spark/src/main/scala/cloudflow/spark/SparkStreamlet.scala
```diff
@@ -123,26 +132,33 @@ trait SparkStreamlet extends Streamlet[SparkStreamletContext] with Serializable
...
-              } else {
-                val p = Promise[Unit]()
-                s.scheduleOnce(frequency) {
-                  p.completeWith(_poll())
-                }
-              }
-            }
-          }
-        _poll()
+      private def poll2(predicate: => Boolean, frequency: FiniteDuration, deadline: FiniteDuration, s: Scheduler): Future[Unit] = {
```
RayRoestenburg, line 135: I like this one the best ;-)

maasg:
Yeah! Merged this :japanese_ogre:

PR #228, opened 2020-04-06
Introduce remote docker image inspection

skonto:
# What changes were proposed in this pull request?
Adds support for remote inspect of docker images, instead of pulling them locally.

This is an optimization required because pulling images can be very slow at the beginning.

No.

Via integration tests and manually.


skonto:
Forgot to add the required flags for building it will do shortly.

kubectl-cloudflow/docker/client.go
```diff
@@ -147,53 +151,71 @@ func imageMatchesNameAndTag(image types.ImageSummary, imageNameAndTag string) bo
...
-	// need this till we have a way to use docker registry API to check labels
-	pulledImage, pullError := PullImage(cli, imageName)	
-	if pullError != nil {	
-		return cloudflowapplication.CloudflowStreamletDescriptorsDigestPair{}, "", nil, fmt.Errorf("Failed to pull image %s: %s", imageName, pullError.Error())
-	}
-
-	images, err := cli.ImageList(context.Background(), types.ImageListOptions{})
+func GetStreamletDescriptorsForImageWithRegistryInspection(imageName string)(cloudflowapplication.CloudflowStreamletDescriptorsDigestPair, string, *PulledImage, error) {
+	ins := InspectOptions{Image: &ImageOptions{}}
+	localInspectOuput, err := ins.InspectLocalDockerImage(imageName)
```
RayRoestenburg, line 158: There can be a difference between the local image and the remote image, we should at least check that the sha is the same (since the image that is remote will actually be deployed)

skonto: That means we will have to assume it exists at the remote registry. What if it does not exist and it has not been pushed yet, then we should allow verification anyway.

RayRoestenburg: That is correct. I think we need to think a bit about the UX. 
- What does it mean when you do "kubectl cloudflow verify", is it always remote, or, 
- only remote if local is not the same, or 
- can you pass in a flag --local or something.. 
- or do we use specific image refs "schemas", local://, and docker:// or something
- What if I want to runlocal, can I use my local images.. 
Something to discuss I think. (it's a not really part of this PR, maybe)
@agemooij

kubectl-cloudflow/cmd/verify.go
```diff
@@ -12,25 +12,33 @@ import (
...
+	verifyOpts := &verifyOptions{}
+	verifyOpts.cmd = &cobra.Command{
 		Use:     "verify",
 		Short:   "Verifies the blueprint of a Cloudflow Application.",
-		Example: "kubectl cloudflow verify <blueprint-file-name>",
-		Run:     verifyBlueprintCMD.verifyImpl,
+		Long: `Verifies a Cloudflow blueprint file. This command allows to explicitly verify a blueprint without deploying it.
+Blueprint verification will inspect the images in the "images"" section of the blueprint,
+get the streamlet descriptors from the image labels and will proceed to the verification of the rest of the blueprint
+sections which are "streamlets" and "connections". Note here that it will pull the required images if they are not available locally.
```
RayRoestenburg, line 27: It will not always pull, right, update this doc?

kubectl-cloudflow/docker/client.go
```diff
@@ -147,53 +151,71 @@ func imageMatchesNameAndTag(image types.ImageSummary, imageNameAndTag string) bo
...
-		// uncompressed data : []byte
-		uncompressed, err := ioutil.ReadAll(reader)
+// parseImageInspectData gets inspection data and the imageName and returns streamlet descriptors and other related info about the inspected image
+func parseImageInspectData(outputData InspectOutput, imageName string)(cloudflowapplication.CloudflowStreamletDescriptorsDigestPair, string, *PulledImage, error) {
+	label, found := outputData.Labels[streamletDescriptorsLabelName]
+	if found {
+		descriptors, err := getDescriptorsFromCompressedLabel(label)
 		if err != nil {
-			return cloudflowapplication.CloudflowStreamletDescriptorsDigestPair{}, "", nil, fmt.Errorf("Failed to read contents of the application descriptor, %s", err.Error())
+			return cloudflowapplication.CloudflowStreamletDescriptorsDigestPair{}, "", nil, fmt.Errorf("Failed to get streamlet descriptors from compressed Labelfor image: %s, %s", imageName, err.Error())
```
RayRoestenburg, line 187:
```suggestion
			return cloudflowapplication.CloudflowStreamletDescriptorsDigestPair{}, "", nil, fmt.Errorf("Failed to get streamlet descriptors from compressed label for image: %s, %s", imageName, err.Error())
```

kubectl-cloudflow/docker/client.go
```diff
@@ -147,53 +151,71 @@ func imageMatchesNameAndTag(image types.ImageSummary, imageNameAndTag string) bo
...
 	}
-	return cloudflowapplication.CloudflowStreamletDescriptorsDigestPair{}, "", nil, fmt.Errorf("Unable to inspect image '%s'. It could not be found locally", imageName) 
+	var descriptors cloudflowapplication.Descriptors
+	err = json.Unmarshal([]byte(uncompressed), &descriptors)
+
+	defer func() {
+		err := reader.Close()
+		if err != nil {
+			// most likely we dont want to fail here, skip the error
+			fmt.Println("Failed to close zlib reader")
```
RayRoestenburg, line 217: UX question: What is the user going to do with this message?

kubectl-cloudflow/docker/inspect.go
```diff
@@ -0,0 +1,196 @@
+// This code part is copied and modified from the skopeo project,
+// https://github.com/containers/skopeo.
```
RayRoestenburg, line 2: Do you need to include some copyright header?

kubectl-cloudflow/docker/inspect.go
```diff
@@ -0,0 +1,196 @@
...
+}
+
+// optionalString is a string with a separate presence flag.
+type optionalString struct {
+	present bool
+	value   string
+}
+
+func (opts *InspectOptions) InspectLocalDockerImage(imageName string)(imageData *InspectOutput, retErr error) {
+	return opts.inspectImage("docker-daemon:" + imageName)
```
RayRoestenburg, line 57: maybe we can link this to a flag in the command.. (just an idea, 'verify local')

kubectl-cloudflow/docker/client.go
```diff
@@ -147,53 +151,71 @@ func imageMatchesNameAndTag(image types.ImageSummary, imageNameAndTag string) bo
...
 // and the image with digest in a struct
-func GetStreamletDescriptorsForImage(cli *client.Client, imageName string) (cloudflowapplication.CloudflowStreamletDescriptorsDigestPair, string, *PulledImage, error) {
-	// need this till we have a way to use docker registry API to check labels
-	pulledImage, pullError := PullImage(cli, imageName)	
-	if pullError != nil {	
-		return cloudflowapplication.CloudflowStreamletDescriptorsDigestPair{}, "", nil, fmt.Errorf("Failed to pull image %s: %s", imageName, pullError.Error())
-	}
-
-	images, err := cli.ImageList(context.Background(), types.ImageListOptions{})
+func GetStreamletDescriptorsForImageWithRegistryInspection(imageName string)(cloudflowapplication.CloudflowStreamletDescriptorsDigestPair, string, *PulledImage, error) {
```
RayRoestenburg, line 156: Do you still need the pulledImage?

kubectl-cloudflow/docker/client.go
skonto: It should be a debug message but we havent setup logging proerly for the cli. This should be improved when refactor the cli code base. I might remove it for now so it does not look weird.

kubectl-cloudflow/docker/inspect.go
skonto: Not sure I thought a bit before but will double-check.

kubectl-cloudflow/docker/inspect.go
skonto: Let's address this when we decide upon the UX. Not fix it here.

kubectl-cloudflow/docker/inspect.go
```diff
@@ -0,0 +1,196 @@
...
+		DockerCertPath:           opts.dockerCertPath,
+		OCISharedBlobDirPath:     opts.sharedBlobDir,
+		AuthFilePath:             opts.authFilePath,
+		DockerDaemonHost:         opts.dockerDaemonHost,
+		DockerDaemonCertPath:     opts.dockerCertPath,
+	}
+
+	if opts.credsOption.present {
+		var err error
+		ctx.DockerAuthConfig, err = getDockerAuth(opts.credsOption.value)
```
RayRoestenburg, line 147: How does this work? where do the credsOption come from?

skonto: This an option in the API that we might want to use in the future not used at the moment. For example if we want to pass directly password and username the api is there. I added it because we had a discussion about username and password now passed directly in deploy for the old image format and was not sure if we want this as well in the future. If we see we dont need it we can remove it.

kubectl-cloudflow/docker/client.go
```diff
@@ -147,53 +151,71 @@ func imageMatchesNameAndTag(image types.ImageSummary, imageNameAndTag string) bo
...
+	ins := InspectOptions{Image: &ImageOptions{}}
```
RayRoestenburg, line 157: How does this work? (does it somehow find the credentials, if you pass in this empty ImageOptions struct?)

skonto: This is the default options (nothing special eg no `dockerCertPath`) and it is empty as the underlying lib relies by default on the `docker login` to have happened before, user only needs to supply the image name there.

kubectl-cloudflow/docker/client.go
agemooij: As I've said in the various JIRa discussions, I do not see much use for using local images for anything besides the (future, as yet unimplemented) multi-image local running feature. We should always use the version on the registry to verify and deploy the blueprint.
AFAIK there is not a significant performance benefit to look at local images, especially not if we first have to check SHAs against the server anyway to make sure the local image isn't stale. 

So I propose we remove any code to perform local inspections and iterate from there.

kubectl-cloudflow/docker/inspect.go
agemooij: See above. I would like to remove the local inspect unless we have a very good reason for it.

kubectl-cloudflow/docker/client.go
skonto: I will create a factory method for default options as it is common: https://stackoverflow.com/questions/18125625/constructors-in-go

kubectl-cloudflow/docker/client.go
skonto: @agemooij there is also another reason to avoid it for now. Talking to the local daemon takes consistently too much time, check this (copying from the github actions build): 
```
dockerhub inspect took 347.708373ms
local inspect took 24.204437677s
```
I noticed the same on my machine. I have no idea yet why.

kubectl-cloudflow/docker/client.go
RayRoestenburg: Ok great. that's cool. What happens if they have not logged in, fails?

kubectl-cloudflow/docker/client.go
RayRoestenburg: Wow that's counter-intuitive..

kubectl-cloudflow/docker/client.go
```diff
@@ -147,53 +151,71 @@ func imageMatchesNameAndTag(image types.ImageSummary, imageNameAndTag string) bo
...
-			return cloudflowapplication.CloudflowStreamletDescriptorsDigestPair{}, "", nil, fmt.Errorf("Failed to read contents of the application descriptor, %s", err.Error())
+			return cloudflowapplication.CloudflowStreamletDescriptorsDigestPair{}, "", nil, fmt.Errorf("Failed to get streamlet descriptors from compressed Labelfor image: %s, %s", imageName, err.Error())
+		} else {
+			var descriptorsDigest cloudflowapplication.CloudflowStreamletDescriptorsDigestPair
+			descriptorsDigest.ImageDigest = outputData.Digest.String()
+			descriptorsDigest.StreamletDescriptors = descriptors.StreamletDescriptors
+			return descriptorsDigest, descriptors.APIVersion, &PulledImage{ImageName:imageName, Authenticated:true}, nil
 		}
+	} else {
+		return cloudflowapplication.CloudflowStreamletDescriptorsDigestPair{}, "", nil, fmt.Errorf("No label %s is found in local image: %s", streamletDescriptorsLabelName, imageName)
```
RayRoestenburg, line 195:
```suggestion
		return cloudflowapplication.CloudflowStreamletDescriptorsDigestPair{}, "", nil, fmt.Errorf("No '%s' label is found in local image: %s", streamletDescriptorsLabelName, imageName)
```

kubectl-cloudflow/docker/inspect_test.go
```diff
@@ -0,0 +1,59 @@
...
+	"time"
+)
+
+var imageNameToTest = "docker.io/lightbend/spark-aggregation:134-d0ec286-dirty"
+
+func Test_inspect(t *testing.T) {
+	ins := NewInspectOptions()
+
+	start := time.Now()
+	output, err := ins.InspectRemoteDockerImage("docker.io/lightbend/spark-aggregation:134-d0ec286-dirty")
```
RayRoestenburg, line 18: Add some tests?
- inspect a not existing remote docker image
- inspect an image that you do not have credentials for
(Would also prefer gingko tests here ;-)

skonto: it is possible.

kubectl-cloudflow/docker/client.go
skonto: ok...

kubectl-cloudflow/docker/client.go
skonto: Yes I tested it with aws where the credentials expired.

skonto:
@RayRoestenburg @agemooij I updated the PR. I have asked internally about the copyright stuff ;)

PR #230, opened 2020-04-07
Increase memory limits of the Spark Driver

maasg:
As a temporary workaround for the crash issues with the call-record-aggregator/aggregator streamlet, I propose an increase of the spark driver memory size.
This change seems to provide enough room for the driver to work properly even under pressure.
All observations are captured here: https://docs.google.com/presentation/d/1gxNG_5qXK_hOYN30EMkrWkqw_Q3f00ltqz4NpU--J1o/edit?usp=sharing
A new load test is running to validate the setting. In the short time that's been running, it seems to be working fine.

### DIFF
--- a/core/cloudflow-operator/src/main/resources/application.conf
+++ b/core/cloudflow-operator/src/main/resources/application.conf
@@ -81,7 +81,7 @@ cloudflow {
       }
 
       spark-runner-driver {
-        requests-memory = "512M"
+        requests-memory = "1024M"
         requests-memory = ${?SPARK_DRIVER_REQUESTS_MEMORY}
 
         # memory-overhead is the amount of off-heap memory to allocate in cluster mode, in MiB unless otherwise specified.


core/cloudflow-operator/src/main/resources/application.conf
```diff
@@ -81,7 +81,7 @@ cloudflow {
       }
 
       spark-runner-driver {
-        requests-memory = "512M"
+        requests-memory = "1024M"
```
agemooij, line 84: @maasg  These are only the requests, not the limits. Will the Spark Operator turn this value into a limit somehow?

yuchaoran2011: I don't think so. This is a limit and will be treated by the Spark operator as such

maasg: @agemooij @yuchaoran2011 I think the name used here is for consistency. It does not directly relate to `request/limit` of a k8s pod. 
Here's the configuration flow: 
IMG_20200407_164849

yuchaoran2011:
I saw this [thread](https://kubernetes.slack.com/archives/CALBDHMTL/p1586773937027100) of discussions in the Spark operator Slack channel today. It turns out it's Spark itself that doesn't support specifying just the limit. It sets both the limit and request to the same value.

Screen Shot 2020-04-13 at 09 40 12

PR #238, opened 2020-04-15
Bail out early in verify if images section is empty in blueprint

debasishg:

* If user tries to verify with old styled blueprint, bail out early

* Without handling of this error, it results in a panic - bad ux

* Yes, the user sees a better error message now

* Manually

### DIFF
--- a/kubectl-cloudflow/verify/verify.go
+++ b/kubectl-cloudflow/verify/verify.go
@@ -21,6 +21,10 @@ import (
 func VerifyBlueprint(content string) (Blueprint, []*docker.PulledImage, []string, error) {
 	config := configuration.ParseString(content)
 	imageRefsFromBlueprint := getImageRefsFromConfig(config)
+	if len(imageRefsFromBlueprint) == 0 {
+		return Blueprint{}, nil, nil, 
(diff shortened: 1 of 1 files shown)


kubectl-cloudflow/verify/verify.go
```diff
@@ -21,6 +21,10 @@ import (
 func VerifyBlueprint(content string) (Blueprint, []*docker.PulledImage, []string, error) {
 	config := configuration.ParseString(content)
 	imageRefsFromBlueprint := getImageRefsFromConfig(config)
+	if len(imageRefsFromBlueprint) == 0 {
+		return Blueprint{}, nil, nil, 
+			fmt.Errorf("This blueprint format is unsupported in cli based verify - the [images] section seems to be missing. If you are using the old style blueprint format, please try verify from within sbt")
```
olofwalker, line 26:
```suggestion
			fmt.Errorf("This blueprint format is unsupported in CLI based verify - the [images] section seems to be missing. If you are using the old-style blueprint format, please try to verify from within Sbt")
```

debasishg: changed ..

Issue #204, opened 2020-03-29
Custom logic for JSON parsing in HttpServerLogic in AkkaServerStreamlets


wjglerum:
When I'm using an AkkaServerStreamlet I would like to define a custom object that the streamlet consumes. I followed the example in the [documentation](https://cloudflow.io/docs/current/develop/akka-streamlet-utilities.html#_define_a_custom_route) to customize my HttpServerLogic inside the AkkaServerStreamlet. In the AvroOutlet I would like to send `Data` objects and my Akka-http endpoint should accept `DataEvent` objects. Next we convert our `DataEvent` objects to `Data` objects when writing them to the sink, see below.

```scala
class ExampleStreamlet extends AkkaServerStreamlet with Directives {
  val out = AvroOutlet[Data]("out")
  final override val shape = StreamletShape.withOutlets(out)

  final override def createLogic: HttpServerLogic[Data] = new HttpServerLogic(this, out) {
    override def route(writableSinkRef: WritableSinkRef[Data]): Route =
      path("vessel") {
        put {
          entity(as[DataEvent]) { event =>
            onSuccess(writableSinkRef.write(convertToData(event))) { _ =>
              complete(StatusCodes.OK)
            }
          }
        }
      }
  }
  private def convertToData(dataEvent: DataEvent): Data = ???
}
```

We defined implicits for the `DataEvent` object for the JsonFormats. However we also need to define a implicit JsonFormat for our `Data` objects. Then everything works fine and our akka-http endpoint accepts `DataEvent` object and our AvroOutlet uses `Data` objects. 

If we don't define the implicit JsonFormat for our `Data` object the compilation will fail. This is because of the type signature of `HttpServerLogic` which requires a implicit `FromByteStringUnmarshaller[Out]`  to be present, see below https://github.com/lightbend/cloudflow/blob/27ac2b57888a74f2353bc198ec5c627c68f6123b/core/cloudflow-akka-util/src/main/scala/cloudflow/akkastream/util/scaladsl/HttpServerLogic.scala#L122-L125

How could we fix this?

Also see the discussion on [Gitter](https://gitter.im/lightbend/cloudflow?at=5e7cc9c2e4eb057b38711a79)

PR #243, opened 2020-04-16
Fix http server logic

RayRoestenburg:

Fixes https://github.com/lightbend/cloudflow/issues/204.

bugfix. HttpServerLogic forced the use of an outlet, which is not necessary.

Yes, `HttpServerLogic` API is changed. `HttpServerLogic` does not mandate the use of an outlet anymore. This is a breaking change.

Unit tests pass. Examples build correctly.


debasishg:
@RayRoestenburg 1 minor suggestion .. 

Instead of `HttpWriterLogic extends HttpServerLogic` can we have `HttpServerLogic` _has-a_ `HttpWriterLogic` and keep `HttpServerLogic` as the main interface to the user ? And use something like `HttpServerLogic.writer.defaultStreaming(..)` .. Then the naming looks more natural as we are dealing with a class `AkkaServerStreamlet` - hence dealing with a class having a name server makes more sense.

RayRoestenburg:
I'll have a look @debasishg good idea.

RayRoestenburg:
@debasishg I changed it quite a bit based on your feedback. It is now a lot better I think. Thanks!

PR #249, opened 2020-04-20
assert rows received from a trustable source

maasg:
Patch the pesky data assertion with the `rate` producer.
This approach is based on the data that we have effectively seen so far and should not fail randomly on some internal Spark process that makes the data visible on the temporary SQL table.


maasg:
Thanks, @olofwalker !

PR #374, opened 2020-04-28
Generating CR in target directory.

RayRoestenburg:

Minor change to generate CR in the target dir.

"target" is where build artifacts should go

Output in sbt shows the new path where the CR will be.

tested with an example app.


RayRoestenburg:
@yuchaoran2011 also added a change, to rename `generateCR` to `buildApp`

yuchaoran2011:
Yes saw that. I agree that's a better name than `generateCR`

Issue #376, opened 2020-04-29
Migrate cloudflow-docs to cloudflow


maasg:
We want to merge cloudflow-docs into cloudflow.
From a project perspective, it reduces the moving parts and makes clear to contributors where the docs are.
From the docs perspective, this allows us to align the versions of the examples + snippets with the documentation in a natural way. This paves the way to implement multi-version support for the documentation.

PR #377, opened 2020-04-29
Integrate docs

maasg:
Trying this merge again, now with a full commit history.
No need to review. That was done already in #376


maasg:
commit merge is not enabled. Need to try again.

PR #380, opened 2020-04-29
Some fixes

RayRoestenburg:

- Improved buildApp and verifyBlueprint commands so you can just call them directly (without calling build first). the application project does not have to depend or aggregate the streamlet library projects (which makes sense because different runtimes can  have different and conflicting class paths). 
- Fixed bug in kubectl cloudflow deploy, image digest were not used.

Added a couple of settings in the plugin:
- extraDockerInstructions - allows for adding more instructions to the docker images (for instance for copying certs to the image)
- ownerInDockerImage - the owner in the docker image that you can use when copying more files to the docker image
- Added a fix for CrashloopBackoff not showing up in kubectl cloudflow status.
- Minor fix for kubectl cloudflow list, removed an empty line.

Just some minor fixes. The extraDockerInstructions allows, for instance, for adding certs to the docker image, which you could need for connecting to some service secured with SSL.

Some new plugin settings.

Against GKE.


debasishg:
@RayRoestenburg This PR also includes the whole of doc changes. Aren't they already in the `master` ?

core/sbt-cloudflow/src/main/scala/cloudflow/sbt/BlueprintVerificationPlugin.scala
```diff
@@ -48,9 +48,20 @@ object BlueprintVerificationPlugin extends AutoPlugin {
...
+          .taskDyn {
+            val filter = ScopeFilter(inProjects(thisProject.value.uses: _*))
+            Def.task {
+              val allValues = cloudflowStreamletDescriptors.all(filter).value
+              allValues
+            }
+          })
+          .value
+          .flatten
+          .toMap,
```
debasishg, line 61: This is the real trick 👍 .. this `ScopeFilter` thingy ..

core/sbt-cloudflow/src/main/scala/cloudflow/sbt/BuildAppPlugin.scala
```diff
@@ -105,6 +112,8 @@ object BuildAppPlugin extends AutoPlugin {
     val file = new File(s"target/${appDescriptor.appId}.json")
     IO.write(file, cr.toJson.compactPrint)
     log.success(s"Cloudflow application CR generated in ${file.getAbsolutePath}")
+    log.success(s"Use the following command to deploy the Cloudflow application:")
+    log.success(s"kubectl cloudflow deploy ${file.getAbsolutePath}")
```
yuchaoran2011, line 116: This was missing compared to the output of old single image. Great to have it back!

core/sbt-cloudflow/src/main/scala/cloudflow/sbt/BlueprintVerificationPlugin.scala
RayRoestenburg: It's still not exactly right for all cases, for instance where you have one project, which is both blueprint and streamlets lib.. (spark-sensors).. it would be better to just be able to find all of them by plugin..

PR #383, opened 2020-04-30
Minor fix to scripts

yuchaoran2011:
Found a small issue with the scripts when running them on EKS.

### DIFF
--- a/installer/common/shared.sh
+++ b/installer/common/shared.sh
@@ -190,14 +190,13 @@ export requestsCpu="1"
 export limitsMemory="1024M"
 export limitsCpu="2"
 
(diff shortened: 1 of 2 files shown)


installer/common/shared.sh
```diff
@@ -208,7 +207,7 @@ install_efs_provisioner() {
 helm upgrade $EFS_SERVER_NAME stable/$EFS_CHART_NAME \
 --install \
 --namespace "$1" \
---timeout 600 \
+--timeout "$HELM_TIMEOUT" \
```
yuchaoran2011, line 210: Note that `600` doesn't work. Helm will complain that unit of time is missing.

PR #389, opened 2020-04-30
Version by branch

maasg:
This PR moves the version spec to versioned content, allowing for correct interpolation of the version when variables are used.


maasg:
merging this attempt at making versions work.

PR #426, opened 2020-05-19
Configuration fix.

RayRoestenburg:

Fix for config changes.

There were some issues with merging the configuration in the CLI. The ported lightbend/config library in golang is not complete enough, to sidestep this, all configuration for an app is appended to a input secret. The CLI does still do some validation. The operator deals with merging and config modification.

No.

Tested with many scenarios on GKE.

### DIFF
--- a/core/cloudflow-akka/src/main/scala/cloudflow/akkastream/AkkaStreamletContextImpl.scala
+++ b/core/cloudflow-akka/src/main/scala/cloudflow/akkastream/AkkaStreamletContextImpl.scala
@@ -68,7 +68,7 @@ final class AkkaStreamletContextImpl(
     val gId   = topic.groupId(streamletRef, inlet)
 
     val consumerSettings = ConsumerSettings(system, new ByteArrayDeserializer, new ByteArrayDeserializer)
-      .withBootstrapServers(topic.bootstrapServers.getOrElse(defaultBootstrapServers))
+      .withBootstrapServers(topic.bootstrapServers.getOrElse(internalKafkaBootstrapServers))
       .withGroupId(gId)
       .withProperty(ConsumerConfig.AUTO_OFFSET_RESET_CONFIG, "earliest")
       .withProperties(topic.kafkaConsumerProperties)
@@ -98,7 +98,7 @@ final class AkkaStreamletContextImpl(
   def committableSink[T](outlet: CodecOutlet[T], committerSettings: CommitterSettings): Sink[(T, Committable), NotUsed] = {
     val topic = findTopicForPort(outlet)
     val producerSettings = ProducerSettings(system, new ByteArraySerializer, new ByteArraySerializer)
-      .withBootstrapServers(topic.bootstrapServers.getOrElse(defaultBootstrapServers))
+      .withBootstrapServers(topic.bootstrapServers.getOrElse(internalKafkaBootstrapServers))
       .withProperties(topic.kafkaProducerProperties)
 
     Flow[(T, Committable)]
@@ -120,7 +120,7 @@ final class AkkaStreamletContextImpl(
                                                    committerSettings: CommitterSettings): Sink[(T, CommittableOffset), NotUsed] = {
     val topic = findTopicForPort(outlet)
     val producerSettings = ProducerSettings(system, new ByteArraySerializer, new ByteArraySerializer)
-      .withBootstrapServers(topic.bootstrapServers.getOrElse(defaultBootstrapServers))
+      .withBootstrapServers(topic.bootstrapServers.getOrElse(internalKafkaBootstrapServers))
       .withProperties(topic.kafkaProducerProperties)
 
     Flow[(T, CommittableOffset)]
@@ -142,7 +142,7 @@ final class AkkaStreamletContextImpl(
     val topic = findTopicForPort(inlet)
     val gId   = topic.groupId(streamletRef, inlet)
     val consumerSettings = ConsumerSettings(system, new ByteArrayDeserializer, new ByteArrayDeserializer)
-      .withBootstrapServers(topic.bootstrapServers.getOrElse(defaultBootstrapServers))
+      .withBootstrapServers(topic.bootstrapServers.getOrElse(internalKafkaBootstrapServers))
       .withGroupId(gId)
       .withProperty(ConsumerConfig.AUTO_OFFSET_RESET_CONFIG, resetPosition.autoOffsetReset)
       .withProperties(topic.kafkaConsumerProperties)
@@ -159,7 +159,7 @@ final class AkkaStreamletContextImpl(
   def plainSink[T](outlet: CodecOutlet[T]): Sink[T, NotUsed] = {
     val topic = findTopicForPort(outlet)
     val producerSettings = ProducerSettings(system, new ByteArraySerializer, new ByteArraySerializer)
-      .withBootstrapServers(topic.bootstrapServers.getOrElse(defaultBootstrapServers))
+      .withBootstrapServers(topic.bootstrapServers.getOrElse(internalKafkaBootstrapServers))
       .withProperties(topic.kafkaProducerProperties)
 
     Flow[T]
@@ -180,7 +180,7 @@ final class AkkaStreamletContextImpl(
     new KafkaSinkRef(
       system,
       outlet,
-      defaultBootstrapServers,
+      internalKafkaBootstrapServers,
       topic,
       killSwitch,
       completionPromise
--- a/core/cloudflow-akka/src/main/scala/cloudflow/akkastream/AkkaStreamletContextImpl.scala
+++ b/core/cloudflow-akka/src/main/scala/cloudflow/akkastream/AkkaStreamletContextImpl.scala
@@ -62,15 +62,13 @@ final class AkkaStreamletContextImpl(
     def stop(): Future[Dun]    = AkkaStreamletContextImpl.this.stop()
   }
 
-  private val bootstrapServers = system.settings.config.getString("cloudflow.kafka.bootstrap-servers")
-
   // internal implementation that uses the CommittableOffset implementation to provide access to the underlying offsets
   private[akkastream] def sourceWithContext[T](inlet: CodecInlet[T]): SourceWithContext[T, CommittableOffset, _] = {
     val topic = findTopicForPort(inlet)
     val gId   = topic.groupId(streamletRef, inlet)
 
     val consumerSettings = ConsumerSettings(system, new ByteArrayDeserializer, new ByteArrayDeserializer)
-      .withBootstrapServers(topic.bootstrapServers.getOrElse(bootstrapServers))
+      .withBootstrapServers(topic.bootstrapServers.getOrElse(defaultBootstrapServers))
       .withGroupId(gId)
       .withProperty(ConsumerConfig.AUTO_OFFSET_RESET_CONFIG, "earliest")
       .withProperties(topic.kafkaConsumerProperties)
@@ -100,7 +98,7 @@ final class AkkaStreamletContextImpl(
   def committableSink[T](outlet: CodecOutlet[T], committerSettings: CommitterSettings): Sink[(T, Committable), NotUsed] = {
     val topic = findTopicForPort(outlet)
     val producerSettings = ProducerSettings(system, new ByteArraySerializer, new ByteArraySerializer)
-      .withBootstrapServers(topic.bootstrapServers.getOrElse(bootstrapServers))
+      .withBootstrapServers(topic.bootstrapServers.getOrElse(defaultBootstrapServers))
       .withProperties(topic.kafkaProducerProperties)
 
     Flow[(T, Committable)]
@@ -122,7 +120,7 @@ final class AkkaStreamletContextImpl(
                                                    committerSettings: CommitterSettings): Sink[(T, CommittableOffset), NotUsed] = {
     val topic = findTopicForPort(outlet)
     val producerSettings = ProducerSettings(system, new ByteArraySerializer, new ByteArraySerializer)
-      .withBootstrapServers(topic.bootstrapServers.getOrElse(bootstrapServers))
+      .withBootstrapServers(topic.bootstrapServers.getOrElse(defaultBootstrapServers))
       .withProperties(topic.kafkaProducerProperties)
 
     Flow[(T, CommittableOffset)]
@@ -144,7 +142,7 @@ final class AkkaStreamletContextImpl(
     val topic = findTopicForPort(inlet)
     val gId   = topic.groupId(streamletRef, inlet)
     val consumerSettings = ConsumerSettings(system, new ByteArrayDeserializer, new ByteArrayDeserializer)
-      .withBootstrapServers(topic.bootstrapServers.getOrElse(bootstrapServers))
+      .withBootstrapServers(topic.bootstrapServers.getOrElse(defaultBootstrapServers))
       .withGroupId(gId)
       .withProperty(ConsumerConfig.AUTO_OFFSET_RESET_CONFIG, resetPosition.autoOffsetReset)
       .withProperties(topic.kafkaConsumerProperties)
@@ -161,7 +159,7 @@ final class AkkaStreamletContextImpl(
   def plainSink[T](outlet: CodecOutlet[T]): Sink[T, NotUsed] = {
     val topic = findTopicForPort(outlet)
     val producerSettings = ProducerSettings(system, new ByteArraySerializer, new ByteArraySerializer)
-      .withBootstrapServers(topic.bootstrapServers.getOrElse(bootstrapServers))
+      .withBootstrapServers(topic.bootstrapServers.getOrElse(defaultBootstrapServers))
       .withProperties(topic.kafkaProducerProperties)
 
     Flow[T]
@@ -182,7 +180,7 @@ final class AkkaStreamletContextImpl(
     new KafkaSinkRef(
       system,
       outlet,
-      bootstrapServers,
+      defaultBootstrapServers,
       topic,
       killSwitch,
       completionPromise
--- a/core/cloudflow-operator/src/main/scala/cloudflow/operator/CloudflowApplication.scala
+++ b/core/cloudflow-operator/src/main/scala/cloudflow/operator/CloudflowApplication.scala
@@ -109,6 +109,10 @@ object CloudflowApplication {
       applicationSpec
     )
 
+  def getOwnerReferences(app: CR) = List(
+    OwnerReference(app.apiVersion, app.kind, app.metadata.name, app.metadata.uid, Some(true), Some(true))
+  )
+
   def hash(applicationSpec: CloudflowApplication.Spec): String = {
     val jsonString = Json.stringify(Json.toJson(CloudflowApplication(applicationSpec)))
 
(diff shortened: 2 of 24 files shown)


kubectl-cloudflow/cmd/deploy_command.go
```diff
@@ -301,16 +301,14 @@ func dockerConfigEntryExists(k8sClient *kubernetes.Clientset, namespace string,
...
-		if _, err := k8sClient.CoreV1().Secrets(secret.ObjectMeta.Namespace).Get(secret.ObjectMeta.Name, metav1.GetOptions{}); err != nil {
-			if _, err := k8sClient.CoreV1().Secrets(namespace).Create(secret); err != nil {
-				printutil.LogAndExit("Failed to create secret %s, %s", streamletName, err.Error())
-			}
-		} else {
-			if _, err := k8sClient.CoreV1().Secrets(namespace).Update(secret); err != nil {
-				printutil.LogAndExit("Failed to update secret %s, %s", streamletName, err.Error())
-			}
+func createOrUpdateAppInputSecret(k8sClient *kubernetes.Clientset, namespace string, appInputSecret *corev1.Secret) {
+	if _, err := k8sClient.CoreV1().Secrets(appInputSecret.ObjectMeta.Namespace).Get(appInputSecret.ObjectMeta.Name, metav1.GetOptions{}); err != nil {
```
yuchaoran2011, line 305: Minor: ObjectMeta is not needed. "appInputSecret.Namespace" and "appInputSecret.Name" will work.

core/cloudflow-akka/src/main/scala/cloudflow/akkastream/AkkaStreamletContextImpl.scala
```diff
@@ -182,7 +180,7 @@ final class AkkaStreamletContextImpl(
     new KafkaSinkRef(
       system,
       outlet,
-      bootstrapServers,
+      defaultBootstrapServers,
```
maasg, line 183: Why is this called `default`?

core/cloudflow-operator/src/main/scala/cloudflow/operator/CloudflowApplication.scala
```diff
@@ -109,6 +109,10 @@ object CloudflowApplication {
       applicationSpec
     )
 
+  def getOwnerReferences(app: CR) = List(
+    OwnerReference(app.apiVersion, app.kind, app.metadata.name, app.metadata.uid, Some(true), Some(true))
```
maasg, line 113: Some true is not the whole truth, right? ;-)

core/cloudflow-operator/src/main/scala/cloudflow/operator/action/Action.scala
```diff
@@ -43,12 +43,23 @@ sealed trait Action[+T <: ObjectResource] {
    * Returns the created or modified resource, or None if the resource is deleted.
    */
   def execute(client: KubernetesClient)(implicit ec: ExecutionContext, lc: LoggingContext): Future[Action[T]]
+
+  protected def recoverFromConflict[O](future: Future[O], client: KubernetesClient, f: KubernetesClient => Future[O])(
+      implicit ec: ExecutionContext
+  ): Future[O] =
+    future.recoverWith {
+      case e: K8SException if (e.status.code == Some(Action.ConflictCode)) => {
+        f(client)
```
maasg, line 52: we might need to back off here, I think.

core/cloudflow-operator/src/main/scala/cloudflow/operator/action/Action.scala
```diff
@@ -118,16 +129,24 @@ class CreateAction[T <: ObjectResource](
...
       res ← existing
         .map { existingResource ⇒
           val resourceVersionUpdated =
             editor.updateMetadata(resource, resource.metadata.copy(resourceVersion = existingResource.metadata.resourceVersion))
-          client
-            .update(resourceVersionUpdated)(format, resourceDefinition, lc)
-            .map(o ⇒ new CreateAction(o, format, resourceDefinition, editor))
+          recoverFromConflict(
+            client
+              .update(resourceVersionUpdated),
```
maasg, line 144: weird indentation. Maybe scalafmt doing weird things?

core/cloudflow-operator/src/main/scala/cloudflow/operator/action/Action.scala
```diff
@@ -106,9 +117,9 @@ object Action {
  */
 class CreateAction[T <: ObjectResource](
     val resource: T,
-    val format: Format[T],
-    val resourceDefinition: ResourceDefinition[T],
-    val editor: ObjectEditor[T]
+    implicit val format: Format[T],
```
maasg, line 120: moving these implicits here cleaned up the code nicely! :+1:

core/cloudflow-operator/src/main/scala/cloudflow/operator/action/Action.scala
```diff
@@ -215,32 +222,21 @@ class UpdateStatusAction[T <: ObjectResource](
...
-          case e: K8SException => {
-            val errMsg = e.status.message.getOrElse("")
-            if (errMsg.contains("please apply your changes to the latest version and try again")) {
-              executeUpdateStatus(client)
-            } else {
-              throw (e)
-            }
-          }
-        }
+      res ← recoverFromConflict(client.updateStatus(resourceVersionUpdated), client, executeUpdateStatus)
```
maasg, line 238: :+1:

core/cloudflow-operator/src/main/scala/cloudflow/operator/event/ConfigInputChangeEvent.scala
```diff
@@ -0,0 +1,259 @@
...
+          val secret       = watchEvent._object
+          val metadata     = secret.metadata
+          val secretName   = secret.metadata.name
+          val namespace    = secret.metadata.namespace
+          val absoluteName = s"$namespace.$secretName"
+          val data         = getData(secret)
+
+          watchEvent._type match {
+            case EventType.DELETED ⇒
+              currentSecrets = currentSecrets - absoluteName
```
maasg, line 64: given that this starts from `empty`, this clause will potentially not delete anything (unless it gets an `ADDED` and `DELETED` in the same message).
To me, it looks like it's missing the current state somewhere.
ie., in like 53, something like:
```
var currentSecrets = getCurrentSecrets(namespace)
```

core/cloudflow-operator/src/main/scala/cloudflow/operator/event/ConfigInputChangeEvent.scala
```diff
@@ -0,0 +1,259 @@
...
+              List()
+            case EventType.ADDED | EventType.MODIFIED ⇒
+              if (currentSecrets.get(absoluteName).forall { existingEvent ⇒
+                    // the secret must have been updated
+                    existingEvent._object.resourceVersion != watchEvent._object.resourceVersion &&
+                    // the secret must change
```
maasg, line 70: could you structure this better? The comment in between the boolean expression breaks reading and it looks like two different expressions.
Maybe name them?

maasg: see https://github.com/lightbend/cloudflow/pull/428

core/cloudflow-akka/src/main/scala/cloudflow/akkastream/AkkaStreamletContextImpl.scala
RayRoestenburg: It is the bootstrapservers of the kafka cluster that is installed with cloudflow.

core/cloudflow-operator/src/main/scala/cloudflow/operator/CloudflowApplication.scala
RayRoestenburg: hehe, questionable DSL indeed.

core/cloudflow-operator/src/main/scala/cloudflow/operator/action/Action.scala
RayRoestenburg: it's not really necessary since a next get must get the new resource version, since we just stumbled on a conflict (we only use this for that purpose, but that's not really encoded, so in the general sense you are correct, but how we use it, it's fine).

RayRoestenburg: So what is implied is that `future` must always first get the resource.

RayRoestenburg: I've added a comment. Might be able to encode this, to do the get in this code to be safer. But this is just a internal helper so I'll leave it as is.

core/cloudflow-operator/src/main/scala/cloudflow/operator/action/Action.scala
RayRoestenburg: Fixed it.

core/cloudflow-operator/src/main/scala/cloudflow/operator/action/Action.scala
RayRoestenburg: Thanks!

core/cloudflow-operator/src/main/scala/cloudflow/operator/event/ConfigInputChangeEvent.scala
RayRoestenburg: Te currentSecrets is kept in statefulMapConcat across messages, this will only delete if earlier added or modified, which is what is intended.

core/cloudflow-operator/src/main/scala/cloudflow/operator/event/ConfigInputChangeEvent.scala
```diff
@@ -0,0 +1,259 @@
...
+   * Finds the associated [[CloudflowApplication.CR]]s for [[ConfigInputChangeEvent]]s.
+   * The resulting flow outputs tuples of the app and the streamlet change event.
+   */
+  def mapToAppInSameNamespace(
+      client: KubernetesClient
+  )(implicit ec: ExecutionContext): Flow[ConfigInputChangeEvent, (Option[CloudflowApplication.CR], ConfigInputChangeEvent), NotUsed] =
+    Flow[ConfigInputChangeEvent].mapAsync(1) { configInputChangeEvent ⇒
+      val ns = configInputChangeEvent.watchEvent._object.metadata.namespace
+      client.usingNamespace(ns).getOption[CloudflowApplication.CR](configInputChangeEvent.appId).map {
+        case a @ Some(_) ⇒ a    -> configInputChangeEvent
```
maasg, line 107: weird construction. It doesn't look like it's doing something with the value. I think it can be simplified to:
```
.map{cr => cr -> configInputChangeEvent
```
or, if possible,
```
configInputChangeEvent -> cr
```
I'm not a fan of having a `None` as a key.

core/cloudflow-operator/src/main/scala/cloudflow/operator/event/ConfigInputChangeEvent.scala
RayRoestenburg: Also note the forall after the get in Add / Modify, so add is done if no item is found.

core/cloudflow-operator/src/main/scala/cloudflow/operator/event/ConfigInputChangeEvent.scala
maasg: After seeing `toInputConfigUpdateAction`, I think you can safely `flatMap` the value and get rid of the `None` case

core/cloudflow-operator/src/main/scala/cloudflow/operator/event/ConfigInputChangeEvent.scala
RayRoestenburg: Fixed it.

core/cloudflow-operator/src/main/scala/cloudflow/operator/event/ConfigInputChangeEvent.scala
RayRoestenburg: I have to emit an element in the flow when it is not there.

RayRoestenburg: Ah I see what you mean.. must have been different before and not fully cleaned up.

core/cloudflow-operator/src/main/scala/cloudflow/operator/event/ConfigInputChangeEvent.scala
```diff
@@ -0,0 +1,259 @@
...
+  def getStreamletRuntimeConfig(runtime: String, streamletName: String, appConfig: Config) =
+    Try(appConfig.getConfig(globalRuntimeConfigPath(runtime))).toOption
+      .getOrElse(ConfigFactory.empty())
+      .atPath(streamletRuntimeConfigPath(streamletName))
+
+  def moveConfigParameters(config: Config, streamletName: String): Config = {
+    val key                        = streamletConfigPath(streamletName)
+    val configParametersKey        = "config-parameters"
+    val absoluteConfigParameterKey = s"$key.$configParametersKey"
+    val configParametersSection = Try(Some(config.getConfig(absoluteConfigParameterKey)))
```
maasg, line 199: `Try(config.getConfig(absoluteConfigParameterKey)).toOption` ?

RayRoestenburg: Weird. yes!

RayRoestenburg: (I use that everywhere.. brain glitch)

core/cloudflow-operator/src/main/scala/cloudflow/operator/event/ConfigInputChangeEvent.scala
```diff
@@ -0,0 +1,259 @@
...
+  }
+
+  def mergeRuntimeConfigToRoot(config: Config, streamletName: String): Config = {
+    val streamletKey = streamletConfigPath(streamletName)
+
+    val configKey                = "config"
+    val absoluteRuntimeConfigKey = s"$streamletKey.$configKey"
+    val runtimeConfigSection = Try(Some(config.getConfig(absoluteRuntimeConfigKey)))
+      .getOrElse(None)
+    // removing 'config' section and move it contents in the root of the config (akka, spark, flink, etc).
```
maasg, line 215:
```suggestion
    // removing 'config' section and move it contents in the root of the config (akka, spark, flink, etc).
```
// removing 'config' section and move its contents in the root of the config (akka, spark, flink, etc).

core/cloudflow-operator/src/main/scala/cloudflow/operator/event/ConfigInputChangeEvent.scala
```diff
@@ -0,0 +1,259 @@
...
+    runtimeConfigSection
+      .map { c =>
+        val configs = c
+          .root()
+          .entrySet()
+          .asScala
+          .map { entry =>
+            entry.getValue.atPath(entry.getKey)
```
maasg, line 223: I don't get this one.

RayRoestenburg: yeah it is a tricky one, the config has config values and sections under it, to get to these you need to use the entrySet, so for instance:
```
cloudflow.streamlet.my-streamlet {
  config {
    foo = bar
    akka.log-level = DEBUG
    akka.some {
      thing {
        else  = 10
      }
    }
  }
}
```
entrySet gets you key and value for `akka` config section and foo config value. We want those as root config objects / values, so the values are placed at purely the key found, so `akka` and `foo` in this case.

RayRoestenburg: Later these 'root configs' are merged with the rest of the config in the right place (where the `cloudflow.streamlet.my-streamlet.config` is removed.

core/cloudflow-operator/src/test/scala/cloudflow/operator/event/ConfigInputChangeEventSpec.scala
```diff
@@ -0,0 +1,67 @@
...
+    with OptionValues
+    with Inspectors
+    with TestDeploymentContext {
+
+  "ConfigInputChangeEvent" should {
+    "transform the config" in {
+      val appConfig = ConfigFactory.parseString("""
+      cloudflow {
+        streamlets.logger {
+          config-parameters {
```
maasg, line 37: this example is missing one case:
- A config-parameter in this config instance that is not overridden.
It should be present in the result.

core/cloudflow-streamlets/src/main/scala/cloudflow/streamlets/StreamletContext.scala
```diff
@@ -55,6 +55,12 @@ trait StreamletContext {
    */
   def config: Config = streamletDefinition.config
 
+  /**
+   * The default bootstrapServers for the Kafka broker that has been installed or configured
+   * to be used globally for all Cloudflow applications
+   */
+  def defaultBootstrapServers = config.getString("cloudflow.kafka.bootstrap-servers")
```
maasg, line 62: I think `internalKafkaBootstrapServers` or `cloudflowKafkaBootstrapServers` would be better options.

maasg: Feel free to disagree

kubectl-cloudflow/config/config.go
```diff
@@ -142,19 +156,21 @@ func loadAndMergeConfigs(configFiles []string) (*configuration.Config, error) {
...
 		}
 	}
-	return config, nil
+	return &config, nil
 }
 
-func validateConfig(config *configuration.Config, applicationSpec cfapp.CloudflowApplicationSpec) error {
-	if config.IsEmpty() {
+func validateConfig(config *Config, applicationSpec cfapp.CloudflowApplicationSpec) error {
+	if config.isEmpty() {
```
maasg, line 163: weird... an exported method name starting with smallcaps?

core/cloudflow-streamlets/src/main/scala/cloudflow/streamlets/StreamletContext.scala
RayRoestenburg: I think that is better, thanks

kubectl-cloudflow/config/config.go
RayRoestenburg: it's not exported

core/cloudflow-operator/src/main/scala/cloudflow/operator/event/ConfigInputChangeEvent.scala
RayRoestenburg: @maasg yeah there is a little copy/paste going on here.. the mapToAppInSameNamespace is used slightly differently in StreamletChangeEvents. will cleanup.

Issue #422, opened 2020-05-19
remove deprecated Cloudflow plugins


maasg:
Cloudflow still contains several plugins for specific combinations of application and backend (eg. `CloudflowAkkaStreamletApplicationPlugin`) that are not needed anymore and even more importantly, will not work with the new multi-image implementation.

We need to clean these up and update the documentation with the new way to structure cloudflow applications.

PR #429, opened 2020-05-20
clean deprecated plugins 

maasg:
This PR removes deprecated plugins after the implementation of the multi-image support. 
This improves clarity in what plugins need to be used.


maasg:
closes #422

PR #431, opened 2020-05-20
Kubernetes (pods and containers) configuration

RayRoestenburg:

Allow configuration of pods and runtimes
TODO:
- [x] Validate config in CLI
- [x] Translate Config objects in AkkaRunner.resource to Deployment
- [ ] Provide first only through `kubectl cloudflow configure` and fix issues with deploy in a separate ticket.
- [ ] Translate Config objects in SparkRunner.resource to Spark App.
- [ ] Translate Config objects in FlinkRunner.resource to Flink App.
- [ ] Test.

To allow the user to freely configure their application details.

yes, more config options.

Testing on GKE.

### DIFF
--- a/core/cloudflow-operator/src/main/scala/cloudflow/operator/action/Action.scala
+++ b/core/cloudflow-operator/src/main/scala/cloudflow/operator/action/Action.scala
@@ -124,7 +124,7 @@ class CreateOrUpdateAction[T <: ObjectResource](
     implicit val editor: ObjectEditor[T]
 ) extends Action[T] {
 
-  val name = "create"
+  val name = "create-or-update"
 
   /**
    * Creates the resources if it does not exist. If it does exist it updates the resource as required.
--- a/core/cloudflow-operator/src/main/scala/cloudflow/operator/action/Action.scala
+++ b/core/cloudflow-operator/src/main/scala/cloudflow/operator/action/Action.scala
@@ -65,11 +65,11 @@ object Action {
   val ConflictCode = 409
 
   /**
-   * Creates a [[CreateAction]].
+   * Creates a [[CreateOrUpdateAction]].
    */
-  def create[T <: ObjectResource](resource: T, editor: ObjectEditor[T])(implicit format: Format[T],
-                                                                        resourceDefinition: ResourceDefinition[T]) =
-    new CreateAction(resource, format, resourceDefinition, editor)
+  def createOrUpdate[T <: ObjectResource](resource: T, editor: ObjectEditor[T])(implicit format: Format[T],
+                                                                                resourceDefinition: ResourceDefinition[T]) =
+    new CreateOrUpdateAction(resource, format, resourceDefinition, editor)
 
   /**
    * Creates a [[DeleteAction]].
@@ -77,12 +77,11 @@ object Action {
   def delete[T <: ObjectResource](resource: T)(implicit resourceDefinition: ResourceDefinition[T]) =
     new DeleteAction(resource, resourceDefinition)
 
-  /**
-   * Creates an [[UpdateAction]].
-   */
-  def update[T <: ObjectResource](resource: T, editor: ObjectEditor[T])(implicit format: Format[T],
-                                                                        resourceDefinition: ResourceDefinition[T]) =
-    new UpdateAction(resource, format, resourceDefinition, editor)
+  def createOrPatch[T <: ObjectResource, O <: Patch](
+      resource: T,
+      patch: O
+  )(implicit format: Format[T], patchWriter: Writes[O], resourceDefinition: ResourceDefinition[T]) =
+    new CreateOrPatchAction(resource, patch, format, patchWriter, resourceDefinition)
 
   /**
    * Creates an [[PatchAction]].
@@ -115,10 +114,10 @@ object Action {
 }
 
 /**
- * Captures creation of the resource. This action does not fail if the resource already exists.
- * If the resource already exists, it will be updated, ensuring the desired effect of creating the resource.
+ * Captures create or update of the resource. This action does not fail if the resource already exists.
+ * If the resource already exists, it will be updated.
  */
-class CreateAction[T <: ObjectResource](
+class CreateOrUpdateAction[T <: ObjectResource](
     val resource: T,
     implicit val format: Format[T],
     implicit val resourceDefinition: ResourceDefinition[T],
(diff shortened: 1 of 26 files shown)


core/cloudflow-operator/src/main/scala/cloudflow/operator/action/Action.scala
```diff
@@ -65,24 +65,23 @@ object Action {
   val ConflictCode = 409
 
   /**
-   * Creates a [[CreateAction]].
+   * Creates a [[CreateOrUpdateAction]].
    */
-  def create[T <: ObjectResource](resource: T, editor: ObjectEditor[T])(implicit format: Format[T],
-                                                                        resourceDefinition: ResourceDefinition[T]) =
-    new CreateAction(resource, format, resourceDefinition, editor)
+  def createOrUpdate[T <: ObjectResource](resource: T, editor: ObjectEditor[T])(implicit format: Format[T],
```
RayRoestenburg, line 70: create and update shared where exactly the same, so made the name more explicit and removed duplication.

core/cloudflow-operator/src/main/scala/cloudflow/operator/action/Action.scala
```diff
@@ -65,24 +65,23 @@ object Action {
...
   def delete[T <: ObjectResource](resource: T)(implicit resourceDefinition: ResourceDefinition[T]) =
     new DeleteAction(resource, resourceDefinition)
 
-  /**
-   * Creates an [[UpdateAction]].
-   */
-  def update[T <: ObjectResource](resource: T, editor: ObjectEditor[T])(implicit format: Format[T],
-                                                                        resourceDefinition: ResourceDefinition[T]) =
-    new UpdateAction(resource, format, resourceDefinition, editor)
+  def createOrPatch[T <: ObjectResource, O <: Patch](
```
RayRoestenburg, line 80: The resource is not created immediately in runner actions so for Spark a patch will not be good enough, it will need to be created if it does not exist. This fixes the issue we saw @maasg.

maasg: I think that `patch`  was required for the Spark operator 0.7 issue with deployments. We should test (prob. later) whether we can do an `update`

core/cloudflow-operator/src/main/scala/cloudflow/operator/action/Action.scala
```diff
@@ -160,31 +159,30 @@ class CreateAction[T <: ObjectResource](
...
     for {
       existing ← client.getOption[T](resource.name)
       res ← existing
-        .map(existingResource ⇒
-          editor.updateMetadata(resource, resource.metadata.copy(resourceVersion = existingResource.metadata.resourceVersion))
-        )
-        .map(resourceToUpdate ⇒ recoverFromConflict(client.update(resourceToUpdate), client, executeUpdate))
-        .getOrElse(client.create(resource))
+        .map(_ ⇒ client.patch(resource.name, patch, Some(resource.ns)))
+        .getOrElse(recoverFromConflict(client.create(resource), client, executeCreateOrPatch))
```
maasg, line 185: this might be a tricky situation. the `patch` could also cause a conflict, no?

core/cloudflow-operator/src/main/scala/cloudflow/operator/action/RunnerActions.scala
```diff
@@ -71,13 +67,10 @@ abstract class RunnerActions[T <: ObjectResource](runner: Runner[T]) {
           val resource     = SparkRunner.resource(deployment, newApp, namespace)
           val patch        = SparkRunner.patch(deployment, newApp, namespace)
           val patchAction  = Action.patch(resource, patch)(SparkRunner.format, SparkRunner.patchFormat, SparkRunner.resourceDefinition)
-          val configAction = Action.update(runner.configResource(deployment, newApp, namespace), runner.configEditor)
-          Seq(configAction, patchAction)
+          val configAction = Action.createOrUpdate(runner.configResource(deployment, newApp, namespace), runner.configEditor)
+          Seq(configAction)
```
maasg, line 71: why is the `patchAction` not needed here? Will it be able to `scale` without it? (I think that I added it for that reason)

RayRoestenburg: No scale is not working yet.

RayRoestenburg: I'm still working on this issue where updates can happen in the wrong order. this patch could happen after the patch that occurs because of config secret change, meaning it uses the defaults for PodConfig and RuntimeConfig, which wipes out the changes to these (which are translated to CR and resource requirements).

RayRoestenburg: But it is a good point, I need to solve this differently.

core/cloudflow-operator/src/main/scala/cloudflow/operator/action/Action.scala
RayRoestenburg: good catch! yes it gets a little tricky

core/cloudflow-operator/src/main/scala/cloudflow/operator/event/ConfigInputChangeEvent.scala
```diff
@@ -180,23 +185,33 @@ object ConfigInputChangeEvent {
 
   val TopicsConfigPath                                  = "cloudflow.topics"
   def streamletConfigPath(streamletName: String)        = s"cloudflow.streamlets.$streamletName"
```
maasg, line 187: We changed the config path for the runner from `cloudflow.runner.streamlets` to `cloudflow.runner.streamlet`.
https://github.com/lightbend/cloudflow/blob/master/core/cloudflow-streamlets/src/main/scala/cloudflow/streamlets/StreamletDefinition.scala#L118
Do we need to do this on this path as well?

RayRoestenburg: No this is the user -level config file

core/cloudflow-operator/src/main/scala/cloudflow/operator/action/Action.scala
RayRoestenburg: fixed.

core/cloudflow-operator/src/main/scala/cloudflow/operator/action/RunnerActions.scala
RayRoestenburg: I put this back in.

PR #435, opened 2020-05-26
check that appCR.status is not nil before dereferencing

maasg:
Mini PR that checks that the appCR.status is present before dereferencing it.


maasg:
@yuchaoran2011 yes, we are hitting this b/c of a bug. I just want to avoid the NPE error dump.

PR #440, opened 2020-05-27
Moved NFS out of bootstrap script

### LINKED ISSUE #436: When using the new operator-based installer, it says "Error: Could not find tiller"
I'm trying to use the operator-based installer on the current `master`
Following the instructions in the `readme`, I downloaded https://github.com/lightbend/cloudflow/releases/download/v1.3.3/bootstrap-install-script-1.3.3.sh

When executing it, I get:
```
 $ ./bootstrap-install-script-1.3.3.sh 
Error: could not find tiller
```

yuchaoran2011:
Fixed #436. 

I investigated running the Helm command in a docker container containing the Helm 3 executable but it turned out to be quite tricky to get authentication to work. So instead I just removed the reference to Helm from the bootstrap script and point the user to a separate utility script that installs NFS. This re-uses some scripts from the old bash-based installer. 

Users on Helm 2 were failing to run the bootstrap script. This fixed it.

Yes. Helm 2 users should be able to run the bootstrap script.

GKE


olofwalker:
```
I investigated running the Helm command in a docker container containing the Helm 3 executable but it turned out to be quite tricky to get authentication to work.```

Can you elaborate on this?

File: installer/init_helm.sh [hunk 1]
```diff
@@ -0,0 +1,6 @@
+#!/usr/bin/env bash
```
maasg, line 1: let's call this `install_nfs.sh`

installer/release/bootstrap-install-script.sh.template
```diff
@@ -16,10 +16,6 @@ print_error_message() {
   echo "$(tput setaf 1)$errorMessage$(tput sgr0)"
 }
 
-##### Install NFS StorageClass
```
maasg, line 19: how can I generate a new installer script?

yuchaoran2011:
@olofwalker The image I tried to use was an Alpine image for Helm: https://hub.docker.com/r/alpine/helm. Based off of it, I needed to put `kubectl` and mount my local kubeconfig in the image. But it turns out that when connecting to GKE, that was not enough, I had to put the entire Google cloud SDK in the image, which requires a separate authentication. AWS might again need something else. So I just resorted to retooling our existing scripts.

installer/release/bootstrap-install-script.sh.template
yuchaoran2011: Run `make all` from inside `installer/release`. I'm going to add more doc about how to use it in another PR. Another pending task is to automatically build a new script when we cut a release.

yuchaoran2011:
The build failed with an error that doesn't seem to be caused by this PR. Merging...

olofwalker:
@yuchaoran2011 ok, I think we preferred having the gcloud package in the container and just use that when installing on GKE.

PR #444, opened 2020-05-27
Fix config system secrets

RayRoestenburg:

Fix for handling config changes. Instead of creating a a secret per config type, the data is added under different keys in the secret for the streamlet deployment.

There were some complications in the previous solution that caused out of order updates of the resources.

no.

GKE.

TODO: (in separate PR)
- [ ] translate PodsConfig to Flink resource requirements and env
- [ ] translate PodsConfig to Spark resource requirements and env

### DIFF
--- a/core/cloudflow-operator/src/main/scala/cloudflow/operator/action/Action.scala
+++ b/core/cloudflow-operator/src/main/scala/cloudflow/operator/action/Action.scala
@@ -29,14 +29,10 @@ import skuber.api.patch.Patch
 sealed trait Action[+T <: ObjectResource] {
 
   /**
-   * The Kubernetes resource to create, update, or delete
-   */
-  def resource: T
-
-  /**
    * The action name
    */
   def name: String
+  def namespace: String
 
   /**
    * Executes the action using a KubernetesClient.
@@ -56,6 +52,8 @@ sealed trait Action[+T <: ObjectResource] {
       }
     }
 
+  def executing: String
+  def executed: String
 }
 
 /**
@@ -74,8 +72,8 @@ object Action {
   /**
    * Creates a [[DeleteAction]].
    */
-  def delete[T <: ObjectResource](resource: T)(implicit resourceDefinition: ResourceDefinition[T]) =
-    new DeleteAction(resource, resourceDefinition)
+  def delete[T <: ObjectResource](resourceName: String, namespace: String)(implicit resourceDefinition: ResourceDefinition[T]) =
+    DeleteAction(resourceName, namespace, resourceDefinition)
 
   def createOrPatch[T <: ObjectResource, O <: Patch](
       resource: T,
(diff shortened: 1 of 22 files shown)


core/cloudflow-operator/src/main/scala/cloudflow/operator/action/Action.scala
```diff
@@ -102,15 +118,22 @@ object Action {
...
-  def executed(action: Action[ObjectResource]) =
-    s"Executed ${action.name} action for ${action.resource.kind} ${action.resource.metadata.name} in namespace ${action.resource.namespace}"
+  def executed(action: Action[ObjectResource]) = action.executed
+}
 
+abstract class ResourceAction[T <: ObjectResource] extends Action[T] {
+  def resource: T
+  def resourceName = resource.metadata.name
+  def namespace    = resource.metadata.namespace
+  def executing =
```
maasg, line 133: I like the idea of adding some state markers. Yet, I find that this approach could be better structured.
The difference in the messages between `executing` and `executed` is very small every time. It also follows a pattern we could exploit:
`<action> for <identifier>` 
Not really a blocker right now.

RayRoestenburg: yeah this is just refactoring what we had already.. The delete action was a bit weird that it needed the total resource to delete it, so I modified that.

core/cloudflow-operator/src/main/scala/cloudflow/operator/runner/Runner.scala
```diff
@@ -96,20 +103,54 @@ trait Runner[T <: ObjectResource] {
...
+      .fromConfig(ConfigFactory.parseString(str))
+      .recover {
+        case e =>
+          log.error(
+            s"Detected pod configs in secret '${secret.metadata.name}' that contains invalid configuration data, IGNORING configuration.",
+            e
+          )
+          PodsConfig()
+      }
+      .getOrElse(PodsConfig())
```
maasg, line 134: Given that `recover` returns a `PodConfig()` isn't `get` sufficient here?

core/cloudflow-operator/src/main/scala/cloudflow/operator/runner/Runner.scala
```diff
@@ -96,20 +103,54 @@ trait Runner[T <: ObjectResource] {
...
+    Try(ConfigFactory.parseString(str))
+      .recover {
+        case e =>
+          log.error(
+            s"Detected runtime config in secret '${secret.metadata.name}' that contains invalid configuration data, IGNORING configuration.",
+            e
+          )
+          ConfigFactory.empty
+      }
+      .getOrElse(ConfigFactory.empty)
```
maasg, line 148: same here

core/cloudflow-operator/src/main/scala/cloudflow/operator/runner/Runner.scala
```diff
@@ -188,7 +229,8 @@ object PodsConfig {
    *  }}}
    */
   def fromConfig(config: Config): Try[PodsConfig] =
-    Try(PodsConfig(asConfigObjectToMap[PodConfig](config.getConfig("kubernetes.pods"))))
+    if (config.isEmpty) Try(PodsConfig())
```
maasg, line 232: `Success(PodsConfig())` ?

integration-tests/swiss-knife/build.sbt
```diff
@@ -24,6 +24,9 @@ lazy val swissKnife = (project in file("."))
       
     )
 lazy val app = (project in file("./app"))
+  .settings(
+    name:= "swiss-knife"
```
maasg, line 28: I have this on a branch as well. 3rd time today, I think :-)

maasg:
merged :-)

PR #450, opened 2020-05-28
Adding env, java options and resource requirements for Spark

RayRoestenburg:

Passing through config to spark pods for resource requirements and java options.
Also see https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/blob/8c480acfdd09882ed2f00573f15e7830558de524/pkg/apis/sparkoperator.k8s.io/v1beta2/types.go

- java options are taken from env var `JAVA_OPTS` in config. This env var is removed from env list passed to pod specs, and added to JavaOptions in pod spec, falling back to defaults specified in cloudflow.
- env vars, except for JAVA_OPTS, are passed through to the pods
- if config just contains `kubernetes.pods.pod` this setting is used for both driver and executor.
- if config contains `kubernetes.pods.driver` and `kubernetes.pods.executor`, these settins will be used respectively for driver and executor
- resource requirements fallback to defaults defined in cloudflow
- resource requirements are passed through as follows:
  - cores is derived from requests.cpu, converted to int, minimum set to 1
  - coreRequest is set to requests.cpu for executor only , docs say this is only supported in driver for Spark 3.0
- coreLimit is set to limits.cpu
- memory is set to requests.memory
- memoryOverhead is only set from defaults in cloudflow installation, (and later it will be possible to set through spark runtime setting (`spark.driver.memoryOverhead`)

Users want to customize pod/container resource requirements and environment settings for Spark.

The user will be able to set env, java options and resource requirements for Spark

GKE


core/cloudflow-operator/src/main/scala/cloudflow/operator/runner/SparkRunner.scala
```diff
@@ -194,6 +197,141 @@ object SparkRunner extends Runner[CR] with PatchProvider[SpecPatch] {
...
+        podConfig.containers.get(PodsConfig.CloudflowContainerName).flatMap { containerConfig =>
+          containerConfig.resources.map { resources =>
+            updatedDriver.copy(
+              cores = toIntCores(resources.requests.get(Resource.cpu)).orElse(coresOpt),
+              memory = resources.requests.get(Resource.memory).map(_.toString).orElse(updatedDriver.memory),
+              coreLimit = resources.limits.get(Resource.cpu).map(_.toString).orElse(updatedDriver.coreLimit)
+            )
+          }
+        }
+      }
```
yuchaoran2011, line 234: So if the user just sets generic pod settings, that will override specific settings for the Spark driver? Isn't it more sensible to be the other way around (i.e. more specific settings like driver-level ones take precedence over higher level settings)?

core/cloudflow-operator/src/main/scala/cloudflow/operator/runner/SparkRunner.scala
```diff
@@ -194,6 +197,141 @@ object SparkRunner extends Runner[CR] with PatchProvider[SpecPatch] {
...
+          containerConfig.resources.map { resources =>
+            updatedExecutor.copy(
+              cores = toIntCores(resources.requests.get(Resource.cpu)).orElse(coresOpt),
+              coreRequest = resources.requests.get(Resource.cpu).map(_.value),
+              memory = resources.requests.get(Resource.memory).map(_.toString).orElse(updatedExecutor.memory),
+              coreLimit = resources.limits.get(Resource.cpu).map(_.toString).orElse(updatedExecutor.coreLimit)
+            )
+          }
+        }
+      }
```
yuchaoran2011, line 281: Same question for the setting precedence for the executor.

core/cloudflow-operator/src/main/scala/cloudflow/operator/runner/SparkRunner.scala
```diff
@@ -240,11 +378,16 @@ object SparkResource {
...
+  /**
+   * NOTE: coreRequest in Driver is only supported in Spark 3.0:
+   * https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/blob/8c480acfdd09882ed2f00573f15e7830558de524/pkg/apis/sparkoperator.k8s.io/v1beta2/types.go#L499
+   */
   final case class Driver(
       cores: Option[Int] = Some(1),
       coreLimit: Option[String] = None,
       memory: Option[String] = None,
       memoryOverhead: Option[String] = None,
+      env: Option[List[EnvVar]] = None,
```
yuchaoran2011, line 390: How about renaming it to `envs` because it's a list of envs?

core/cloudflow-operator/src/main/scala/cloudflow/operator/runner/SparkRunner.scala
```diff
@@ -257,9 +400,11 @@ object SparkResource {
   final case class Executor(
       instances: Int,
       cores: Option[Int] = Some(1),
+      coreRequest: Option[String] = None,
       coreLimit: Option[String] = None,
       memory: Option[String] = None,
       memoryOverhead: Option[String] = None,
+      env: Option[List[EnvVar]] = None,
```
yuchaoran2011, line 407: Plural here too

core/cloudflow-operator/src/main/scala/cloudflow/operator/runner/SparkRunner.scala
RayRoestenburg: The name in json has to be env

core/cloudflow-operator/src/main/scala/cloudflow/operator/runner/SparkRunner.scala
RayRoestenburg: This is the mapping to the CR so can't just be changed.

core/cloudflow-operator/src/main/scala/cloudflow/operator/runner/SparkRunner.scala
RayRoestenburg: the updatedDriver is already set with defaults, and here the cores, memory and limit are always set, taking the values from the podConfig first falling back to existing settings (from defaults), so  the precedence is as you would like it.

core/cloudflow-operator/src/main/scala/cloudflow/operator/runner/SparkRunner.scala
RayRoestenburg: it's podConfig resources or else executorSettings, where executorSettings are the defaults.

core/cloudflow-operator/src/main/scala/cloudflow/operator/runner/SparkRunner.scala
RayRoestenburg: That will not work, mapping to spark https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/blob/8c480acfdd09882ed2f00573f15e7830558de524/pkg/apis/sparkoperator.k8s.io/v1beta2/types.go#L431

core/cloudflow-operator/src/main/scala/cloudflow/operator/runner/SparkRunner.scala
RayRoestenburg: Same

maasg:
there's one thing that I struggle with. We don't actually control the driver/exec pods of Spark. Not even the Spark operator does that. It's the  `spark-submit` internal implementation that directs the deployment. 
As we are offering more control on the specific configuration of the different subsystems, I wonder whether we should allow the configuration of the kubernetes resource through the spark specific configuration (spark.driver.cores, spark.executor.cores, ...) and not through `kubernetes.pods` which we don't actually directly control.

RayRoestenburg:
Yes, as discussed we keep it like this for now:
- the defaults are also translated this way
- the container resources settings is more advanced, and we will have to explain.
I'm going to do another test to make sure what the precedence is for setting cores both in spark runtime config and the kubernetes section.

RayRoestenburg:
So the precedence did not work as expected and I fixed this by adding the config to runtime config.

PR #462, opened 2020-06-03
Test configure app resources

maasg:
Adds an integration test to validate the update of resources for an Akka streamlet through a configuration file.

PS: If you know a better way to deal with the parsing of the kubectl/go output, pls let me know.
I tried the Json way but `kubectl -o json-template` does not output JSON but go structs as well :-(


integration-tests/itest/deploy_test.go
```diff
@@ -143,12 +145,48 @@ var _ = Describe("Application deployment", func() {
 			close(done)
 		}, LongTimeout)
 
-		It("should have configured a flink streamlet", func(done Done) {
+		XIt("should have configured a flink streamlet", func(done Done) {
```
RayRoestenburg, line 148: Ok. so we keep this as pending for now so it does not break stuff, LGTM.

integration-tests/itest/deploy_test.go
```diff
@@ -143,12 +145,48 @@ var _ = Describe("Application deployment", func() {
...
+			Expect(someAkkaPod).NotTo(Equal(nil))
+			podRes, err := kubectl.GetPodResources(swissKnifeApp.Name, someAkkaPod.Pod)
+			Expect(err).NotTo(HaveOccurred())
+
+			By("Reconfigure a single Akka streamlet")
+			err = cli.Configure(swissKnifeApp, UpdateAkkaProcessResourcesFile)
+			Expect(err).NotTo(HaveOccurred())
+
+			By("Wait for the deployment of the new configuration")
+			oneSecond, _ := time.ParseDuration("5s")
```
RayRoestenburg, line 169: fiveSeconds ;-)

integration-tests/itest/deploy_test.go
maasg: Yeah. It produces a random failure. -- we should test it manually from time to time, though.

maasg:
@yuchaoran2011 I tried all options. `-o json` returns the complete resource in JSON. `-o jsonpath` returns the addressed element, but in go template format :shrug: (https://github.com/kubernetes/kubernetes/issues/16707) 
After the release pressure, I'll look into the go client for k8s as a better structured option.

PR #466, opened 2020-06-03
sort status report by streamlet name, desc

maasg:
Orders the output of the `kubectl cloudflow status` command by streamlet name , descending.
This provides a stable ordering to the command and makes it easy to follow the progress.
For example, before, when using `watch` the output would move around at each state change. Now, you can quickly find the streamlet of interest.

(Low hanging :cherries: )


maasg:
closes #466

PR #472, opened 2020-06-05
Upgrade test impl to 2.0.0 rc4

maasg:
This PR upgrades swiss-knife and the test impl to support cloudflow 2.0.0-RC4.
In particular:
- added a delay after deploy to ensure that the deploy status is known
- refactored the `Status` parsing to support the absence of pods
- added command output when a CLI command fails to improve debugging.


maasg:
closes #473

PR #481, opened 2020-06-08
remove focus

maasg:
forgot to "unfocus" the test on the last PR


maasg:
@yuchaoran2011 Thanks for the trick!

Issue #487, opened 2020-06-08
Build the installer image in Github Actions


yuchaoran2011:
**Is your feature request related to a problem? Please describe.**
The new operator-based installer can be unknowingly broken without testing it in PRs.

**Is your feature request related to a specific runtime of cloudflow or applicable for all runtimes?**
Not related to any specific runtime.

**Describe the solution you'd like**
Add the build step in Github Actions to build the installer image when a PR is created.

**Describe alternatives you've considered**
None

**Additional context**
None

PR #488, opened 2020-06-09
Build installer in Github Actions. Update Helm repo before installing NFS

yuchaoran2011:
#487 

To prevent accidentally making changes that would break installer build 

No.

Ran the bash snippet locally. But let's see if the Github Actions work as intended.

### DIFF
--- a/scripts/build-all.sh
+++ b/scripts/build-all.sh
@@ -25,16 +25,32 @@ then
   echo "Error: 'target' missing."
   echo "Usage: build-all.sh <target>  # where <target> is an sbt target: compile, test, ..."
   echo "==================================================================================="
-  exit -1
+  exit 1
 fi
 echo "========================================================================="
 echo "Runs 'sbt $TARGET' for core and examples"
 echo "========================================================================="
 
+cd $DIR/../installer/yaml
+if make all; then
+  echo "Successfully generated YAML for installer"
+else
+  echo "Failed generating YAML for installer"
+  exit 1
+fi
+
+cd ../
+if sbt docker; then
+  echo "Successfully built installer Docker image"
+else
+  echo "Failed building installer Docker image"
+  exit 1
+fi
+
 cd $DIR/../core
 sbt --supershell=false "; scalafmtCheck ; $TARGET  ; publishLocal"
 RETVAL=$?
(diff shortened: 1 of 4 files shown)


yuchaoran2011:
Well, the build failed with:
```
git clone --single-branch --branch master git@github.com:lightbend/cloudflow.git work/cloudflow-operator/git_repo
Cloning into 'work/cloudflow-operator/git_repo'...
Warning: Permanently added the RSA host key for IP address '140.82.114.4' to the list of known hosts.
git@github.com: Permission denied (publickey).
fatal: Could not read from remote repository.
Makefile:222: recipe for target 'helm-fetch-cloudflow-operator' failed
Failed generating YAML for installer

Please make sure you have the correct access rights
and the repository exists.
```
Any idea why it can't clone the cloudflow repo, which is public?

scripts/build-all.sh
```diff
@@ -67,6 +83,6 @@ for prj in $PROJECTS; do
       ;;
   esac
   RETVAL=$?
-  [ $RETVAL -ne 0 ] && echo "Failure in project $prj" && exit -1
+  [ $RETVAL -ne 0 ] && echo "Failure in project $prj" && exit 1
```
yuchaoran2011, line 86: What does a -1 return code mean? I don't think it's valid. `exit 1` will indicate the script failed.

maasg: Wow. That rabbit hole goes deep. Looks like this comes from `C`, where `resultCode < 0 ` was an error and `resultCode >= 0` had particular 'success' meanings.

RayRoestenburg:
@yuchaoran2011 the build fails on yaml generation.

scripts/build-all.sh
yuchaoran2011: Well, that's not what bash uses though. In bash scripts, return code of 0 means success and 1-255 indicates error.

yuchaoran2011:
@RayRoestenburg Yes it failed to generate YAML because somehow it couldn't clone a public repo. Weird. But I saw that you skipped doing that in your other PR, we don't have to worry about it.
I'll re-open to merge the bash script return code in this PR, after I cleanup the unneeded stuff.

RayRoestenburg:
Closed, fixed in another PR @yuchaoran2011.

PR #490, opened 2020-06-09
move update helm repo from #488 to a separate commit/PR

maasg:
Fix the repo requirement for the nfs installer.
Taken from #488 into a separate PR to solve this quickly.

FYI @yuchaoran2011


maasg:
tested on GKE - merging.

Issue #505, opened 2020-06-15
Make installer image configurable


yuchaoran2011:
**Is your feature request related to a problem? Please describe.**
The installer image is currently hard-coded as `lightbend/cloudflow-installer`. It should be configurable.

**Is your feature request related to a specific runtime of cloudflow or applicable for all runtimes?**
All runtimes.

**Describe the solution you'd like**
Make the image an optional argument to the bootstrap script.

**Describe alternatives you've considered**
None

**Additional context**
User request

PR #506, opened 2020-06-16
Made installer image configurable in bootstrap script

yuchaoran2011:
#505 Previously, the installer image is hard coded as `lightbend/cloudflow-installer`. The user can now pass a custom image (including the docker repository and the image name) to the bootstrap script. Something like: `./bootstrap-install-script-2.0.0.sh lightbend-docker-commercial-registry.bintray.io/lightbend/cloudflow-installer`. If not specified, the default image `lightbend/cloudflow-installer` will be used.

A user request

Yes. Bootstrap script now has an optional argument.

By running the script without the argument and with argument and making sure the installer image has been changed properly.

### DIFF
--- a/installer/release/bootstrap-install-script.sh.template
+++ b/installer/release/bootstrap-install-script.sh.template
@@ -16,6 +16,13 @@ print_error_message() {
   echo "$(tput setaf 1)$errorMessage$(tput sgr0)"
 }
 
+CLOUDFLOW_INSTALLER_IMAGE=$1
+if [ -z "$CLOUDFLOW_INSTALLER_IMAGE" ]; then
+    CLOUDFLOW_INSTALLER_IMAGE=lightbend/cloudflow-installer
+fi
+echo "Cloudflow installer image to be used: $CLOUDFLOW_INSTALLER_IMAGE"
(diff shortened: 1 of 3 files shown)


installer/release/bootstrap-install-script.sh.template
```diff
@@ -223,7 +230,8 @@ else
 fi
 
 echo "$CLOUDFLOW_INSTALLER_YAML" | \
-  sed -e "s/CLOUDFLOW_INSTALLER_IMAGE_TAG_PH/$CLOUDFLOW_INSTALLER_IMAGE_TAG/" | \
+  sed -e "s~CLOUDFLOW_INSTALLER_IMAGE_TAG_PH~$CLOUDFLOW_INSTALLER_IMAGE_TAG~" \
+      -e "s~CLOUDFLOW_INSTALLER_IMAGE_PH~$CLOUDFLOW_INSTALLER_IMAGE~" | \
```
yuchaoran2011, line 234: Because `$CLOUDFLOW_INSTALLER_IMAGE` likely contains slashes, to avoid confusing `sed`, a new separator is needed.

PR #520, opened 2020-06-19
Fixes and improvements to AKS cluster creation script

yuchaoran2011:
* The latest fully supported version of Kubernetes on AKS is 1.16.9. Higher versions are also available but are still in preview. Updated the default version to be 1.16.9
* A8 VMs will be deprecated next year. Changed the default VM type to A4.
* Enable autoscaler when creating the cluster.
* Fixed a bug where the values of the parameters could be not set.

Bug fixes.

Yes. Changes are in a user-facing script.

Running the script.

### DIFF
--- a/installer/create-cluster-aks.sh
+++ b/installer/create-cluster-aks.sh
@@ -22,12 +22,12 @@ function usage {
   Options:
   The following options determine the cluster configuration.
   -c | --cluster-name name              The name used for the cluster
-  --resource-group group                The resource group to use
+  -r | --resource-group group           The resource group to use
   --location eastus                     AKS location to launch the cluster in. "eastus" by default
                                         For available locations, refer to https://azure.microsoft.com/en-us/global-infrastructure/services/?products=kubernetes-service.
-  --k8s-version 1.17.3                  Kubernetes version to use. "1.17.3" by default
+  --k8s-version 1.16.9                  Kubernetes version to use. "1.16.9" by default
                                         To view all supported versions, run " az aks get-versions --location <LOCATION>"
-  --node-vm-size Standard_A8_v2         Size of node VMs in the cluster. "Standard_A8_v2" by default.
+  --node-vm-size Standard_A4_v2         Size of node VMs in the cluster. "Standard_A4_v2" by default.
                                         For a list of supported sizes, refer to https://docs.microsoft.com/en-us/azure/virtual-machines/linux/sizes.
                                         using "provision-large.config.template".
   --node-count 5                        Number of nodes to launch. 5 by default.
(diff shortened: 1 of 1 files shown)


installer/create-cluster-aks.sh
```diff
@@ -45,30 +45,30 @@ function parse_args {
...
+      -r|--resource-group)
         shift
         RESOURCE_GROUP_NAME=$1
         [ -z "$RESOURCE_GROUP_NAME" ] && error '"--resource-group" requires a non-empty option argument.\n'
       ;;
       --location)
         shift
         LOCATION=$1
-        [ -z "$LOCATION" ] && error LOCATION=eastus
+        [ -z "$LOCATION" ] && error '"--location" requires a non-empty option argument.\n'
```
maasg, line 56: I don't completely understand the phrase "a non-empty option argument."
Do you mean "a non-empty argument." ? 
(same x3 below)

installer/create-cluster-aks.sh
```diff
@@ -113,9 +118,12 @@ az aks create --resource-group "$RESOURCE_GROUP_NAME" \
   --name "$CLUSTER_NAME" \
   --node-count $NODE_COUNT \
   --enable-addons monitoring \
+  --enable-cluster-autoscaler \
   --generate-ssh-keys \
   --kubernetes-version "$CLUSTER_VERSION" \
-  --node-vm-size $NODE_VM_SIZE
+  --node-vm-size $NODE_VM_SIZE \
+  --min-count 1 \
+  --max-count 7
```
maasg, line 126: How are min/max count related to `$NODE_COUNT` ? 
e.g. if the user gives a $NODE_COUNT > max-count, what's the expected behavior?

installer/create-cluster-aks.sh
```diff
@@ -45,30 +45,30 @@ function parse_args {
         CLUSTER_NAME=$1
         [ -z "$CLUSTER_NAME" ] && error '"--cluster-name" requires a non-empty option argument.\n'
```
RayRoestenburg, line 46: I think it is better to remove the "option" in these sentences. Can't we just say `--clustername requires an argument?`?

installer/create-cluster-aks.sh
RayRoestenburg: I think I would prefer ```--location" requires an argument.\n```

installer/create-cluster-aks.sh
yuchaoran2011: Good question. I haven't tested that scenario yet. Our GKE cluster creation script has the same issue. I'll update both scripts to check that node count is between and min and max counts.

installer/create-cluster-aks.sh
yuchaoran2011: Agreed. Will change.

installer/create-cluster-aks.sh
yuchaoran2011: In the script for GKE, we hard code the min, max and default node counts to be 1, 7 and 5, respectively. I updated this script to do the same. The user can go in the script and change the numbers if they want.

Issue #516, opened 2020-06-18
Upgrade to Akka 2.6.6


RayRoestenburg:
Upgrade Akka dependency to 2.6.6.

PR #530, opened 2020-06-23
Update to Akka 2.6.6, #516

patriknw:
* and enable Split Brain Resolver

Refs #516

### DIFF
--- a/core/project/Dependencies.scala
+++ b/core/project/Dependencies.scala
@@ -3,9 +3,9 @@ import sbt._
 // format: OFF
 object Version {
 
-  val Akka          = "2.6.5"
-  val AkkaHttp      = "10.1.11"
-  val AkkaMgmt      = "1.0.6"
+  val Akka          = "2.6.6"
+  val AkkaHttp      = "10.1.12"
+  val AkkaMgmt      = "1.0.8"
   val AlpakkaKafka  = "2.0.3"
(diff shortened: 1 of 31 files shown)


core/project/Dependencies.scala
```diff
@@ -3,9 +3,9 @@ import sbt._
 // format: OFF
 object Version {
 
-  val Akka          = "2.6.5"
-  val AkkaHttp      = "10.1.11"
-  val AkkaMgmt      = "1.0.6"
+  val Akka          = "2.6.6"
+  val AkkaHttp      = "10.1.12"
+  val AkkaMgmt      = "1.0.8"
```
patriknw, line 8: http and mgmt doesn't have to update, but I guess it will not hurt to use latest

RayRoestenburg: As long as it does not result in many versions of akka at runtime it's all good, but I don't have to explain that to you ;-) I'll run a test with this, thanks @patriknw!

RayRoestenburg:
@patriknw The CloudflowAkkaPlugin also references the Akka version: https://github.com/lightbend/cloudflow/blob/5b17da2ae8a4adbe2a3e5681a38cf436e85a54e3/core/sbt-cloudflow/src/main/scala/cloudflow/sbt/CloudflowAkkaPlugin.scala#L29

PR #534, opened 2020-06-24
Adding cache-dir option to all antora docker runs.

RayRoestenburg:
It did not work for me locally without this setting, got error:
```Error: EACCES: permission denied, mkdir '/.cache'```


RayRoestenburg:
Does this make sense @maasg or do I have something setup incorrectly? Did you run into this at some point?

maasg:
I don't know exactly. That folder is within the Docker run context. Why wouldn't it be able to create that folder?
One other difference I see between the different targets is that `html` uses `--privileged` and the others do not.

PR #550, opened 2020-06-30
Allow Docker images to be configurable in the installer

yuchaoran2011:
Allow Docker images installed by the installer configurable by the user.

To support use cases where images are stored in and need to be pulled from a custom Docker registry. 

Yes. Now the user can configure the images. If the user doesn't do anything, the default will be applied.

By building the installer image with `sbt dockerBuildAndPush`. Then update the values in `release/definitions.mk` to be values that are not the defaults. Run `make all` under `/release`. Then execute the generated bootstrap script. The user have the ability to modify images in the generated script.


installer/yaml/definitions.mk
```diff
@@ -3,10 +3,18 @@ strimzi_chart_version                := "0.16.2"
...
 cloudflow_operator_image_name        := lightbend/cloudflow-operator
+cloudflow_operator_image_tag         := 2.0.0
+spark_operator_image_name 	     := lightbend/sparkoperator
+spark_operator_image_tag 	     := 1.3.3-OpenJDK-2.4.5-1.1.0-cloudflow-2.12
+flink_operator_image_name	     := lyft/flinkk8soperator
+flink_operator_image_tag	     := v0.4.0
+kafka_image_name		     := strimzi/kafka
+kafka_image_tag			     := 0.16.2-kafka-2.4.0
+strimzi_operator_image_name	     := strimzi/operator
+strimzi_operator_image_tag	     := 0.16.2
```
yuchaoran2011, line 20: The formatting here is strange. In Vim, this file looks properly formatted with the right indentations.

RayRoestenburg: tabs / spaces?

yuchaoran2011: Ah yes! Fixed now.

installer/release/Makefile
```diff
@@ -27,12 +27,44 @@ bootstrap-install-script-${version}.sh: bootstrap-install-script.sh.template clo
...
+	}\
+	/CLOUDFLOW_FLINK_OPERATOR_IMAGE_NAME=""/ {\
+		print "CLOUDFLOW_FLINK_OPERATOR_IMAGE_NAME=\"${cloudflow_flink_operator_image_name}\""; \
+		next \
+	}\
+	/CLOUDFLOW_FLINK_OPERATOR_IMAGE_TAG=""/ {\
+		print "CLOUDFLOW_FLINK_OPERATOR_IMAGE_TAG=\"${cloudflow_flink_operator_image_tag}\""; \
+		next \
+	}\
+	/KAFKA_IMAGE_NAME=""/ {\
```
RayRoestenburg, line 54: So I guess this is strimzi kafka right?

installer/release/definitions.mk
```diff
@@ -1,5 +1,14 @@
...
-cloudflow_spark_operator_image := 1.3.3-OpenJDK-2.4.5-1.1.0-cloudflow-2.12
+cloudflow_installer_image_tag := 2.0.3
+
+cloudflow_operator_image_name := lightbend/cloudflow-operator
+cloudflow_operator_image_tag := 2.0.0
+cloudflow_spark_operator_image_name := lightbend/sparkoperator
+cloudflow_spark_operator_image_tag := 1.3.3-OpenJDK-2.4.5-1.1.0-cloudflow-2.12
+cloudflow_flink_operator_image_name := lyft/flinkk8soperator
+cloudflow_flink_operator_image_tag := v0.4.0
+kafka_image_name := strimzi/kafka
```
RayRoestenburg, line 11: ah right, never mind previous comment.

installer/src/main/scala/com/lightbend/cloudflow/installer/Action.scala
```diff
@@ -139,18 +139,37 @@ final case class CloudflowOperatorManagedStrimzi()(implicit cr: CloudflowInstanc
         "cloudflowOperator.persistentStorageClass"       -> spec.cloudflowOperator.persistentStorageClass,
         "kafkaClusterCr.kafkaPersistentStorageClass"     -> spec.kafkaClusterCR.kafkaPersistentStorageClass,
         "kafkaClusterCr.zookeeperPersistentStorageClass" -> spec.kafkaClusterCR.zooKeeperPersistentStorageClass,
-        "cloudflowOperator.imageTag"                     -> spec.cloudflowOperator.imageTag
```
RayRoestenburg, line 142: Ah right so we already had a 'map', cool, easy to extend.

installer/yaml/resources/strimzi-operator/strimzi-operator.yaml
```diff
@@ -0,0 +1,98 @@
...
+              value: __kafka.imageName__:__kafka.imageTag__
+            - name: STRIMZI_DEFAULT_TLS_SIDECAR_KAFKA_IMAGE
+              value: __kafka.imageName__:__kafka.imageTag__
+            - name: STRIMZI_DEFAULT_TLS_SIDECAR_ZOOKEEPER_IMAGE
+              value: __kafka.imageName__:__kafka.imageTag__
+            - name: STRIMZI_DEFAULT_KAFKA_EXPORTER_IMAGE
+              value: __kafka.imageName__:__kafka.imageTag__
+            - name: STRIMZI_KAFKA_IMAGES
+              value: |
+                2.3.1=strimzi/kafka:0.16.2-kafka-2.3.1
```
RayRoestenburg, line 55: Is it possible to add some comment here that we choos 2.4.0 and don't worry about 2.3.1?

yuchaoran2011:
I realized that ubuntu image is actually used in the installer. Its purpose is to serve as the base image for a Kubernetes job that we run to patch the MutatingWebhookConfiguration object with the right owner reference. My latest commit has made it configurable too. Now this PR, along with https://github.com/lightbend/cloudflow-enterprise-installer/pull/49 should cover all the images. I've verified that they could all be properly replaced on a cluster. Next we need to come up with a way to simulate a no-internet environment.

installer/yaml/resources/strimzi-operator/strimzi-operator.yaml
yuchaoran2011: Comment added.

installer/release/definitions.mk
```diff
@@ -9,7 +9,7 @@ cloudflow_spark_operator_image_tag := 1.3.3-OpenJDK-2.4.5-1.1.0-cloudflow-2.12
 cloudflow_flink_operator_image_name := lyft/flinkk8soperator
 cloudflow_flink_operator_image_tag := v0.4.0
 kafka_image_name := strimzi/kafka
-kakfa_image_tag := 0.16.2-kafka-2.4.0
+kafka_image_tag := 0.16.2-kafka-2.4.0
```
RayRoestenburg, line 12: tricky! ;-)

Issue #508, opened 2020-06-16
Make buildAndPublish sbt command private or rename to prevent confusion. 


RayRoestenburg:
buildAndPublish has been replaced by buildApp. (It might be better to make buildAndPublish private, to not confuse users that used to use it, if it is still used)

RayRoestenburg:
At least it should be made private, since it will confuse users that have used 1.x

PR #562, opened 2020-07-06
deprecate buildAndPublish

maasg:
Deprecates `buildAndPublish` in favor of `buildApp`
Adds an error message to  `buildAndPublish` 
and moves the internal implementation to a private key.


maasg:
closes #508

Issue #509, opened 2020-06-16
Make it possible to modify the docker repository where base images are located.


RayRoestenburg:
Currently base images for akka, spark and flink are published at docker.io/lightbend. A user should be able to take the base images at docker.io/lightbend, push them to a local docker repo. Right now the base image URI's are set in CloudflowAkkaPlugin, CloudflowFlinkPlugin and CloudflowSparkPlugin as simple values. This should be changed to take use settingKeys so the user can override the docker repository and registry part of the URI's.

PR #566, opened 2020-07-08
Made base images configurable

yuchaoran2011:
#509 Made base images for Akka, Flink and Spark configurable.

The user might want to use another image (e.g. an internal one)

Yes, Now in the build.sbt file of the project, the user needs to configure `cloudflowDockerParentImage` for each runtime plugin that gets used. 

Tested CRA and taxi-ride to make sure that images can indeed by configured and `sbt buildApp` works.


core/sbt-cloudflow/src/main/scala/cloudflow/sbt/CloudflowFlinkPlugin.scala
```diff
@@ -25,18 +25,14 @@ import cloudflow.sbt.CloudflowKeys._
...
-
   override def requires = CloudflowBasePlugin
 
   override def projectSettings = Seq(
     libraryDependencies ++= Vector(
           "com.lightbend.cloudflow" %% "cloudflow-flink"         % BuildInfo.version,
           "com.lightbend.cloudflow" %% "cloudflow-flink-testkit" % BuildInfo.version % "test"
         ),
-    cloudflowDockerParentImage := CloudflowFlinkDockerBaseImage,
+    cloudflowDockerParentImage := cloudflowDockerParentImage.value,
```
RayRoestenburg, line 35: It would be better if we can do something like:
```
cloudflowDockerParentImage := cloudflowFlinkBaseImage.value.getOrElse("s"lightbend/flink:${CloudflowBasePlugin.CloudflowVersion}-cloudflow-flink-$FlinkVersion-scala-${CloudflowBasePlugin.ScalaVersion}")
```

RayRoestenburg: where `cloudflowFlinkBaseImage` is a `settingKey[Option[String]]`. Similar for the other base images.

yuchaoran2011:
@RayRoestenburg Yeah I realized it's easy to miss or misconfigure this setting. I'll give a default and revert changes in the example projects.

core/sbt-cloudflow/src/main/scala/cloudflow/sbt/CloudflowFlinkPlugin.scala
yuchaoran2011: Looks like having a `settingKey[Option[String]]` doesn't make the setting optional even though `getOrElse` is used. If I don't specify `cloudflowFlinkBaseImage` in `build.sbt`, then I'll get the following error:
```
[error] References to undefined settings: 
[error] 
[error]   processor / cloudflowFlinkBaseImage from processor / cloudflowDockerParentImage ((cloudflow.sbt.CloudflowFlinkPlugin.projectSettings) CloudflowFlinkPlugin.scala:39)
```
It only works if I put `cloudflowFlinkBaseImage := Option.empty[String]` in `build.sbt`, in which case the default base image will be used. Is there a way to do it without putting anything extra in `build.sbt`?

RayRoestenburg: You have to set it to None in settings in the plugin.
add `cloudflowFlinkBaseImage := None` to projectSettings of the specific plugin  (in above case CloudflowFlinkPlugin) @yuchaoran2011

RayRoestenburg: Similar stuff happens here: https://github.com/lightbend/cloudflow/blob/2cd83d8a226fb086239a1a44df10054efbdbb2a1/core/sbt-cloudflow/src/main/scala/cloudflow/sbt/CloudflowApplicationPlugin.scala#L46

yuchaoran2011: Thanks for the suggestion. The PR has been modified based on it. I've verified that the user can now configure the base image using `cloudflowAkkaBaseImage`, `cloudflowFlinkBaseImage` or `cloudflowSparkBaseImage` in `build.sbt`. Default base images will be used in absence of user-provided setting.

core/sbt-cloudflow/src/main/scala/cloudflow/sbt/CloudflowFlinkPlugin.scala
```diff
@@ -32,11 +32,12 @@ object CloudflowFlinkPlugin extends AutoPlugin {
   override def requires = CloudflowBasePlugin
 
   override def projectSettings = Seq(
+    cloudflowFlinkBaseImage := None,
     libraryDependencies ++= Vector(
           "com.lightbend.cloudflow" %% "cloudflow-flink"         % BuildInfo.version,
           "com.lightbend.cloudflow" %% "cloudflow-flink-testkit" % BuildInfo.version % "test"
         ),
-    cloudflowDockerParentImage := CloudflowFlinkDockerBaseImage,
+    cloudflowDockerParentImage := cloudflowFlinkBaseImage.value.getOrElse(CloudflowFlinkDockerBaseImage),
```
RayRoestenburg, line 40: Is `cloudflowDockerParentImage` still used / needed?

examples/call-record-aggregator/project/cloudflow-plugins.sbt
```diff
@@ -1,5 +1,4 @@
 // Resolver for the cloudflow-sbt plugin
 //
 resolvers += Resolver.url("cloudflow", url("https://lightbend.bintray.com/cloudflow"))(Resolver.ivyStylePatterns)
-addSbtPlugin("com.lightbend.cloudflow" % "sbt-cloudflow" % "2.0.5")
-
+addSbtPlugin("com.lightbend.cloudflow" % "sbt-cloudflow" % "2.1.0-SNAPSHOT")
```
RayRoestenburg, line 4: 👍

core/sbt-cloudflow/src/main/scala/cloudflow/sbt/CloudflowFlinkPlugin.scala
yuchaoran2011: Good catch. Not used anywhere. Removed.

Issue #567, opened 2020-07-08
runLocal "Address already in use" with AkkaCluster


nolangrace:
**Describe the bug**
Akka streamlets are all trying to use port 25520 in runLocal

**To Reproduce**
runLocal for an application using clustering 

**Expected behavior**

**Screenshots**

**Additional context**

nolangrace:
see #568

PR #568, opened 2020-07-08
change akka-cluster-local.conf to randomize port and Join cluster wit…

nolangrace:
…hout seednodes

Akka Streamlet using clustering will be assigned a random port and join programmatically rather than using seed nodes to avoid conflicts between local Akka Streamlets

#567 

tested using runLocal and the connected car example


nolangrace:
fixes #567

PR #572, opened 2020-07-10
Allow a streamlet with an empty shape. Fixes #565

### LINKED ISSUE #565: Allow streamlet with 0 inlets and 0 outlets.

RayRoestenburg:

The StreamletShape is allowed to be empty.

There are cases where users do not use the Kafka-enabled inlets and outlets, but still want to use other streamlet features.
It should not be necessary to define one outlet that is never used, for this particular case. It also allows more iterative development on the streamlet, maybe at a later time inlets and / or outlets are added to the streamlet.

The user can specify `StreamletShape.empty`  (Scala API) or `StreamletShape.createEmpty()` (Java API)

Tested by added a Streamlet to an existing example with an empty StreamletShape, deployed on GKE.


djm1329:
lgtm, thanks for quick turnaround!

PR #573, opened 2020-07-10
Made it possible to modify the Flink StreamExecutionEnvironment in Fli…

RayRoestenburg:
…nkStreamlet.

Make it possible to setup the StreamExecutionEnvironment.

Some settings are not configurable through Flink config, so the user will want to set this.

Yes, the user can now override setupExecutionEnvironment and createExecutionEnvironment

All green.


yuchaoran2011:
There's a scalafmtCheck error that failed the build.

PR #575, opened 2020-07-10
Changing Cloudflow operator to use alpine base image

franciscolopezsancho:
changing base image from ubuntu to alpine

To avoid High Risk vulnerabilities found through Jfrog/Xray:

dpkg-source in dpkg 1.3.0 through 1.18.23 is able to use a non-GNU patch program and does not offer a protection mechanism for blank-indented diff hunks, which allows remote attackers to conduct directory traversal attacks via a crafted Debian source package, as demonstrated by use of dpkg-source on NetBSD.

In util-linux before 2.32-rc1, bash-completion/umount allows local users to gain privileges by embedding shell commands in a mountpoint name, which is mishandled during a umount command (within Bash) by a different user, as demonstrated by logging in as root and entering umount followed by a tab character for autocompletion.

Some HTTP/2 implementations are vulnerable to resource loops, potentially leading to a denial of service. The attacker creates multiple request streams and continually shuffles the priority of the streams in a way that causes substantial churn to the priority tree. This can consume excess CPU.
...

No

Deploying in GKE the operator and deploying example app [call-record-aggregator](https://github.com/lightbend/cloudflow/tree/master/examples/call-record-aggregator)

### DIFF
--- a/core/build.sbt
+++ b/core/build.sbt
@@ -1,8 +1,7 @@
 import sbt._
 import sbt.Keys._
-
 import Library._
-
+import sbtdocker.Instructions
 import sbtrelease.ReleaseStateTransformations._
 
 lazy val root =
@@ -431,9 +430,12 @@ lazy val operator =
         val targetDir    = "/app"
 
         new Dockerfile {
-          from("adoptopenjdk/openjdk8:latest")
+          from("adoptopenjdk/openjdk8:alpine")
+
           entryPoint(s"$targetDir/bin/${executableScriptName.value}")
           copy(appDir, targetDir, chown = "daemon:daemon")
+          addInstruction(
+            Instructions.Run("apk add bash; \\"))
         }
       },
       Test / fork := true,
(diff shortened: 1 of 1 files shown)


core/build.sbt
```diff
@@ -431,9 +430,11 @@ lazy val operator =
         val targetDir    = "/app"
 
         new Dockerfile {
-          from("adoptopenjdk/openjdk8:latest")
+          from("adoptopenjdk/openjdk8:alpine")
```
yuchaoran2011, line 433: Is there a version of Alpine we can specify here?

franciscolopezsancho: AFAIK that adoptopenjdk/openjdk8:alpine points to a 3.12

https://github.com/AdoptOpenJDK/openjdk-docker/blob/master/8/jdk/alpine/Dockerfile.hotspot.releases.full

Not sure we can add anything else @yuchaoran2011

yuchaoran2011: You verified that Alpine 3.9 didn't have critical vulnerabilities. Is that the case for 3.12 too?

franciscolopezsancho: Yep, I run the scan against this very image: adoptopenjdk/openjdk8:alpine

PR #579, opened 2020-07-13
Adding GRPC support for protobuf generation

blublinsky:
Current SBT plugin does not provide GRPC stub generation, which is required for some applications. To enable this generation, its necessary to add "grpc" to protocbridge.Target and a couple of libraries. I also removed "flat_package" to get a more user friendly code generation.

### DIFF
--- a/core/sbt-cloudflow/src/main/scala/cloudflow/sbt/CommonSettingsAndTasksPlugin.scala
+++ b/core/sbt-cloudflow/src/main/scala/cloudflow/sbt/CommonSettingsAndTasksPlugin.scala
@@ -64,8 +64,16 @@ object CommonSettingsAndTasksPlugin extends AutoPlugin {
           }.value,
       publishArtifact in (Compile, packageDoc) := false,
       publishArtifact in (Compile, packageSrc) := false,
-      libraryDependencies += "com.twitter"     %% "bijection-avro" % "0.9.7",
-      libraryDependencies += "org.apache.avro" % "avro"            % "1.8.2",
+      // Add Avro libraries
+      libraryDependencies ++= Seq(
+            "com.twitter"     %% "bijection-avro" % "0.9.7",
+            "org.apache.avro" % "avro"            % "1.8.2"
+          ),
+      // Add protoc libraries
+      libraryDependencies ++= Seq(
+            "io.grpc"              % "grpc-netty"            % scalapb.compiler.Version.grpcJavaVersion,
+            "com.thesamet.scalapb" %% "scalapb-runtime-grpc" % scalapb.compiler.Version.scalapbVersion
+          ),
       // TODO move all of this to schema plugins, possibly specific for runtime (when specific versions of libraries are needed, like Spark and Avro 1.8.2)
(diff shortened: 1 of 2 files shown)


core/sbt-cloudflow/src/main/scala/cloudflow/sbt/CommonSettingsAndTasksPlugin.scala
```diff
@@ -64,8 +64,16 @@ object CommonSettingsAndTasksPlugin extends AutoPlugin {
...
-      libraryDependencies += "com.twitter"     %% "bijection-avro" % "0.9.7",
-      libraryDependencies += "org.apache.avro" % "avro"            % "1.8.2",
+      // Add Avro libraries
+      libraryDependencies ++= Seq(
+            "com.twitter"     %% "bijection-avro" % "0.9.7",
+            "org.apache.avro" % "avro"            % "1.8.2"
+          ),
+      // Add protoc libraries
+      libraryDependencies ++= Seq(
+            "io.grpc"              % "grpc-netty"            % scalapb.compiler.Version.grpcJavaVersion,
```
RayRoestenburg, line 74: Can you use shaded grpc netty instead @blublinsky? https://github.com/akka/akka-grpc/issues/1 https://github.com/grpc/grpc-java/pull/2485

core/sbt-cloudflow/src/main/scala/cloudflow/sbt/CommonSettingsAndTasksPlugin.scala
```diff
@@ -133,7 +141,7 @@ object CommonSettingsAndTasksPlugin extends AutoPlugin {
     protocbridge.Target(
       ScalaGenerator,
       targetPath,
-      Seq("flat_package")
```
RayRoestenburg, line 136: akka grpc uses flat_package by default so I wonder why you want to change this? Can you elaborate @blublinsky ?

core/sbt-cloudflow/src/main/scala/cloudflow/sbt/CommonSettingsAndTasksPlugin.scala
blublinsky: Done

core/sbt-cloudflow/src/main/scala/cloudflow/sbt/CommonSettingsAndTasksPlugin.scala
blublinsky: the issue is that in this case it packages a lot into 1 class. as a result using generated code becomes problematic. In my case it was even using the inner class that was not defined. Typically flat_package is used to limit the amount of generated files, but it does not really help. I would, probably elevate it as a parameter, if you really want to do this

RayRoestenburg: Ok yeah can add some settings for this that the user can choose / override.

RayRoestenburg:
Duplicate of #636

PR #619, opened 2020-08-07
Check what Cloudflow dependencies are installed

olofwalker:
In a future PR we will provide a streamlined installation process for Cloudflow, where Cloudflow is installed and any dependency is installed separately. 

We will no longer install Strimzi, Spark, or Flink automatically, but instead, provide documentation on how to do it using `Helm`. 

The upside of this is that besides a streamlined installation process, users who not use Spark, will never have to have the Spark operator installed. It will also significantly reduce the complexity of upgrading Cloudflow.

This change, however, requires more validation when an application is deployed, for example, this PR adds those checks.

When deploying a CLoudflow application, the application descriptor will be inspected for runner types, and those will be used to validate the existence of these runners in the cluster before proceeding with the deployment.

### DIFF
--- a/kubectl-cloudflow/cmd/deploy_command.go
+++ b/kubectl-cloudflow/cmd/deploy_command.go
@@ -380,14 +380,14 @@ func validateDeployCmdArgs(cmd *cobra.Command, args []string) error {
 // The function validates that the operators for Spark and Flink are installed if the application uses any of those streamlet types
 func validateStreamletRunnersDependencies(applicationSpec cfapp.CloudflowApplicationSpec) {
 
-	runnerType := func(runnerTypeName string) bool {
-		for _, v := range applicationSpec.Streamlets {
-			if v.Descriptor.Runtime == runnerTypeName {
-				return true
-			}
-		}
-		return false
+	type RunnerRequirements struct {
+		crd     string
+		version string
 	}
+
+	runnerTypes := make(map[string]RunnerRequirements)
+	runnerTypes["Spark"] = RunnerRequirements{"sparkapplications.sparkoperator.k8s.io", version.RequiredSparkVersion}
+	runnerTypes["Flink"] = RunnerRequirements{"flinkapplications.flink.k8s.io", version.RequiredFlinkVersion}
 
 	validateRunnerType := func(crdName string, prettyName string, expectedVersion string) error {
 		version, err := exec.Command("kubectl", "get", "crds", crdName, "-o", "jsonpath={.spec.version}").Output()
@@ -400,15 +400,18 @@ func validateStreamletRunnersDependencies(applicationSpec cfapp.CloudflowApplica
 		return nil
 	}
 
-	if runnerType("spark") {
-		if err := validateRunnerType("sparkapplications.sparkoperator.k8s.io", "Spark", version.RequiredSparkVersion); err != nil {
-			printutil.LogErrorAndExit(err)
+	var result []error
+	for k, v := range runnerTypes {
+		if err := validateRunnerType(v.crd, strings.Title(k), v.version); err != nil {
+			result = append(result, err)
 		}
 	}
 
-	if runnerType("flink") {
-		if err := validateRunnerType("flinkapplications.flink.k8s.io", "Flink", version.RequiredFlinkVersion); err != nil {
-			printutil.LogErrorAndExit(err)
+	if len(result) != 0 {
+		for _, err := range result {
+			printutil.PrintError("%s", err.Error())
 		}
+		os.Exit(1)
 	}
+
 }
--- a/kubectl-cloudflow/cmd/deploy_command.go
+++ b/kubectl-cloudflow/cmd/deploy_command.go
@@ -377,7 +377,7 @@ func validateDeployCmdArgs(cmd *cobra.Command, args []string) error {
 	return nil
 }
 
-// The function validates that the operators for Spark and Flink is installed if the application uses any of those streamlet types
+// The function validates that the operators for Spark and Flink are installed if the application uses any of those streamlet types
 func validateStreamletRunnersDependencies(applicatonSpec cfapp.CloudflowApplicationSpec) {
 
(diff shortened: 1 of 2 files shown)


kubectl-cloudflow/cmd/deploy_command.go
```diff
@@ -373,3 +376,42 @@ func validateDeployCmdArgs(cmd *cobra.Command, args []string) error {
 
 	return nil
 }
+
+// The function validates that the operators for Spark and Flink is installed if the application uses any of those streamlet types
+func validateStreamletRunnersDependencies(applicatonSpec cfapp.CloudflowApplicationSpec) {
+
+	const requiredSparkVersion = "v1beta2"
+	const requiredFlinkVersion = "v1beta1"
```
RayRoestenburg, line 384: maybe these consts can go in a more central place, we have a version.go somewhere I think.

kubectl-cloudflow/cmd/deploy_command.go
```diff
@@ -373,3 +376,42 @@ func validateDeployCmdArgs(cmd *cobra.Command, args []string) error {
...
+		return false
+	}
+
+	validateRunnerType := func(crdName string, prettyName string, expectedVersion string) error {
+		version, err := exec.Command("kubectl", "get", "crds", crdName, "-o", "jsonpath={.spec.version}").Output()
+		if err != nil {
+			return fmt.Errorf("cannot detect that %s is installed, please install %s before continuing (%v)", prettyName, prettyName, err.Error())
+		}
+		if string(version) != expectedVersion {
+			return fmt.Errorf("%s is installed but the wrong version, required %s, installed %s", prettyName, requiredSparkVersion, string(version))
```
RayRoestenburg, line 401:
```suggestion
			return fmt.Errorf("Detected wrong version of %s, required %s, installed %s", prettyName, expectedVersion, string(version))
```

kubectl-cloudflow/cmd/deploy_command.go
```diff
@@ -373,3 +376,42 @@ func validateDeployCmdArgs(cmd *cobra.Command, args []string) error {
 
 	return nil
 }
+
+// The function validates that the operators for Spark and Flink is installed if the application uses any of those streamlet types
+func validateStreamletRunnersDependencies(applicatonSpec cfapp.CloudflowApplicationSpec) {
```
RayRoestenburg, line 381:
```suggestion
func validateStreamletRunnersDependencies(applicationSpec cfapp.CloudflowApplicationSpec) {
```

kubectl-cloudflow/cmd/deploy_command.go
```diff
@@ -126,6 +127,8 @@ func (opts *deployOptions) deployImpl(cmd *cobra.Command, args []string) {
 		printutil.LogAndExit("%s", err.Error())
 	}
 
+	validateStreamletRunnersDependencies(applicationSpec)
```
RayRoestenburg, line 130:
```suggestion
	validateStreamletRunnersInstalled(applicationSpec)
```

kubectl-cloudflow/cmd/deploy_command.go
```diff
@@ -373,3 +376,42 @@ func validateDeployCmdArgs(cmd *cobra.Command, args []string) error {
 
 	return nil
 }
+
+// The function validates that the operators for Spark and Flink is installed if the application uses any of those streamlet types
```
RayRoestenburg, line 380:
```suggestion
// The function validates that the operators for Spark and Flink are installed if the application uses any of those streamlet types
```

kubectl-cloudflow/cmd/deploy_command.go
```diff
@@ -373,3 +376,42 @@ func validateDeployCmdArgs(cmd *cobra.Command, args []string) error {
...
+			return fmt.Errorf("cannot detect that %s is installed, please install %s before continuing (%v)", prettyName, prettyName, err.Error())
+		}
+		if string(version) != expectedVersion {
+			return fmt.Errorf("%s is installed but the wrong version, required %s, installed %s", prettyName, requiredSparkVersion, string(version))
+		}
+		return nil
+	}
+
+	if runnerType("spark") {
+		if err := validateRunnerType("sparkapplications.sparkoperator.k8s.io", "Spark", requiredSparkVersion); err != nil {
```
RayRoestenburg, line 407: maybe move the crd names to consts too

kubectl-cloudflow/cmd/deploy_command.go
```diff
@@ -373,3 +376,42 @@ func validateDeployCmdArgs(cmd *cobra.Command, args []string) error {
...
+		if err != nil {
+			return fmt.Errorf("cannot detect that %s is installed, please install %s before continuing (%v)", prettyName, prettyName, err.Error())
+		}
+		if string(version) != expectedVersion {
+			return fmt.Errorf("%s is installed but the wrong version, required %s, installed %s", prettyName, requiredSparkVersion, string(version))
+		}
+		return nil
+	}
+
+	if runnerType("spark") {
```
RayRoestenburg, line 406: Here is another idea to do this: What if you would find the distinct runner names in the streamlets and then for every runtime found just do validateRunnerType, passing in the runtime found? (and have a const map, runtime -> crd, so you just get that crd from the map by runtime key)

kubectl-cloudflow/cmd/deploy_command.go
olofwalker: Moved to version

kubectl-cloudflow/cmd/deploy_command.go
olofwalker: They are anonymous consts already ;)

kubectl-cloudflow/cmd/deploy_command.go
olofwalker: ok, its an alternative

olofwalker:
Pushed updates

PR #626, opened 2020-08-12
[WIP] Amend ScalaPB targets rather than replacing

raboof:
# What changes were proposed in this pull request?

Rather than replacing the `PB.targets`, amend them.

To allow invoking the Akka gRPC code generators in a project also using Akka gRPC.

Previously the Cloudflow plugin would wipe out the `PB.generators`, possibly removing useful generators.

The new behaviour, appending, is not foolproof either: now we run the risk of running the same generators (e.g. the standard ScalaPB ones) multiple times. We need to think about how to deduplicate here.

Manually in https://github.com/raboof/cloudflow-grpc-experiment


raboof:
superseded by #642

PR #627, opened 2020-08-12
Made usage of strimzi entity operator optional.

RayRoestenburg:

The strimzi settings in the cloudflow-operator are now optional. If these are not provided, the cloudflow operator will not create Strimzi KafkaTopic CRs, which are used by the strimzi topic operator.

Users have asked if it is possible to install and run Cloudflow without strimzi kafka cluster and topic operator.

No.

all green, tested on gke.

What still has to happen (in a separate PR) is that the helm chart for installing cloudflow should make it possible to configure the default kafka broker that the cloudflow operator will use for managed topics.

### DIFF
--- a/core/cloudflow-operator/src/main/scala/cloudflow/operator/action/Actions.scala
+++ b/core/cloudflow-operator/src/main/scala/cloudflow/operator/action/Actions.scala
@@ -93,7 +93,7 @@ object Actions {
       currentApp: Option[CloudflowApplication.CR],
       deleteOutdatedTopics: Boolean = false
   )(implicit ctx: DeploymentContext): Seq[Action[ObjectResource]] =
-    if (ctx.kafkaContext.strimziTopicOperatorNamespace.nonEmpty && ctx.kafkaContext.strimziClusterName.nonEmpty) {
+    if (ctx.kafkaContext.useStrimzi) {
       StrimziTopicActions(
         newApp = newApp,
         currentApp = currentApp,
--- a/core/cloudflow-operator/src/main/scala/cloudflow/operator/action/Actions.scala
+++ b/core/cloudflow-operator/src/main/scala/cloudflow/operator/action/Actions.scala
@@ -93,7 +93,20 @@ object Actions {
       currentApp: Option[CloudflowApplication.CR],
       deleteOutdatedTopics: Boolean = false
   )(implicit ctx: DeploymentContext): Seq[Action[ObjectResource]] =
-    TopicActions(newApp, currentApp, deleteOutdatedTopics)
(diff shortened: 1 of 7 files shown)


core/cloudflow-operator/src/main/scala/cloudflow/operator/action/Actions.scala
```diff
@@ -93,7 +93,20 @@ object Actions {
       currentApp: Option[CloudflowApplication.CR],
       deleteOutdatedTopics: Boolean = false
   )(implicit ctx: DeploymentContext): Seq[Action[ObjectResource]] =
-    TopicActions(newApp, currentApp, deleteOutdatedTopics)
+    if (ctx.kafkaContext.strimziTopicOperatorNamespace.nonEmpty && ctx.kafkaContext.strimziClusterName.nonEmpty) {
```
ennru, line 96: Shouldn't this be a `useStrimzi` method on `KafkaContext`?

core/cloudflow-operator/src/test/scala/cloudflow/operator/TestDeploymentContext.scala
```diff
@@ -22,8 +22,8 @@ trait TestDeploymentContext {
   implicit val ctx: DeploymentContext =
```
ennru, line 22: Should this get a better name to express it relies on Strimzi?

core/cloudflow-operator/src/main/scala/cloudflow/operator/action/Actions.scala
RayRoestenburg: yeah could do that. This is really a quick fix..

core/cloudflow-operator/src/test/scala/cloudflow/operator/TestDeploymentContext.scala
RayRoestenburg: Do you mean DeploymentContext relies on strimzi? I'm not sure what you mean?

RayRoestenburg: The DeploymentContext is fine without strimzi, the user then just can't use the 'managed topic feature'.

core/cloudflow-operator/src/main/scala/cloudflow/operator/action/Actions.scala
RayRoestenburg: Done.

core/cloudflow-operator/src/test/scala/cloudflow/operator/TestDeploymentContext.scala
RayRoestenburg: (There is more work needed after this PR..)

Issue #610, opened 2020-07-30
OutofMemory issue in cloudflow-operator using custom skuber version on akka 2.6.8


RayRoestenburg:

RayRoestenburg:
Might decide to sidestep issue with OOM and stick with Akka 2.5.x (if this is related) but include the fixes for vulnerabilities.

raboof:
An existing theory was that this might have to do with the error handling while watching.

I created a test (https://github.com/michael-read/skuber/compare/mread-upgrade-to-akka-2.6.x...raboof:reproduce-memory-leak?expand=1) that seems to confirm this: after running the test (while it's sleeping), even after performing GC there are 5 PoolGateway instances on the heap.

It doesn't seem obvious that this was different on Akka 2.5 though yet, will test that next.

raboof:
> It doesn't seem obvious that this was different on Akka 2.5 though yet, will test that next.

(I also see the same problem on 2.5)

raboof:
> even after performing GC there are 5 PoolGateway instances on the heap.

Looking closer, I think those gateways/pools would have eventually (after the idle time-out) have been cleaned up - but perhaps they did not get cleaned up fast enough in this case.

A good improvement might be to make sure connection pools are reused across streams that share the same HTTPS configuration. Proposed upstream in https://github.com/doriordan/skuber/pull/317 .

RayRoestenburg:
After more investigation it looks like the operator is stable (forcing GC shows the memory getting collected), it is possible it needs some more memory when many apps are kept running. Also at the time some important JAVA_OPTS where missing. It is possible that some error condition in the operator can cause more memory issues, but some safety measures have been taken. Closing this issue for now, though we will keep an eye on it.

PR #628, opened 2020-08-13
Update skuber to use the same Akka version as the rest

raboof:
This is possible since the rest is now ahead of Skuber,
and Akka is binary compatible.

Use the same Akka version for the operator and the rest of the components

The version of Akka used with the operator is too low to support Akka HTTP
10.2.0, and instead of bumping the Akka version in the operator 'just enough'
it seems neater to just pull it all the way up to be consistent with the
other components.

No

`sbt test`

### DIFF
--- a/core/build.sbt
+++ b/core/build.sbt
@@ -408,16 +408,16 @@ lazy val operator =
     .settings(
       scalafmtOnCompile := true,
       libraryDependencies ++= Vector(
-            AkkaSlf4jOperator,
-            AkkaStreamOperator,
+            AkkaSlf4j,
+            AkkaStream,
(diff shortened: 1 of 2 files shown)


raboof:
We need to look into #610 before merging

raboof:
Ready to go AFAICS, but perhaps we want to wait for a release that contains https://github.com/doriordan/skuber/pull/317 ?

RayRoestenburg:
@raboof I think it would be nice to move to skuber 2.6.0, and when  doriordan/skuber#317 is merged, update to that, since 2.6.0 already has great benefits for us.

PR #629, opened 2020-08-13
Update core to Akka HTTP 10.2.0

raboof:

Update core to Akka HTTP 10.2.0 and use the new 'newServerAt' API.

In particular this means you can now configure HTTP/2 on the
`HttpServerLogic`, which is useful for gRPC.

This removes the `Flow` from the protected methods in `HttpServerLogic`,
instead relying only on `Route` (which matches better with HTTP/2).

`sbt test`


raboof:
to be revisited after #628 is merged first.

raboof:
Partly already done as part of #633, the other part will be included in #628

Issue #636, opened 2020-08-17
Akka gRPC support


raboof:
It would be nice to be able to easily introduce gRPC services based on Akka gRPC in a Cloudflow streamlet.

https://github.com/raboof/cloudflow-grpc-experiment is a basic experiment plugging this in in user code (not yet tested beyond 'runLocal').

To make this more smooth, we want to include the Akka gRPC plugin as a dependency of the Cloudflow plugin, and provide something similar to `HttpServerLogic` but for gRPC services.

As a sanity check it would be good to verify that use cases like https://github.com/lightbend/seldoncloudflow/blob/master/support/src/main/scala/com/lightbend/seldon/executors/tensor/SeldonTFGRPCExecutorTensor.scala, which include a gRPC client, are also still possible/easy.

PR #642, opened 2020-08-19
Embed the Akka gRPC plugin

raboof:

This PR embeds the Akka gRPC plugin in the Cloudflow plugin.

The language generated by Akka gRPC follows the model language by default

Refs #636

To make exposing your service as an Akka gRPC service smooth.

When you include a 'service' protobuf in your project, the relevant Akka gRPC code will be generated by default, and you can wire things up by providing an implementation and handing it to GrpcServerLogic

* https://github.com/raboof/cloudflow-grpc-experiment
* both locally and in GKE (with a port-forward directly to the pod)


docs/shared-content-source/docs/modules/develop/pages/akka-streamlet-utilities.adoc
```diff
@@ -6,6 +6,7 @@ include::ROOT:partial$include.adoc[]
 The `cloudflow.akka.util` library contains some predefined `StreamletLogic`{empty}s:
 
 - `HttpServerLogic`
+- `GrpcServerLogic`
```
raboof, line 9: Perhaps we should move `HttpServerLogic` and `GrpcServerLogic` to their own pages, WDYT?

RayRoestenburg: Yes I think that is a good idea.

RayRoestenburg: yes good idea

docs/shared-content-source/docs/modules/develop/pages/akka-streamlet-utilities.adoc
```diff
@@ -106,6 +107,68 @@ include::{cloudflow-examples-version}@docsnippets:ROOT:example$akkastreams-scala
...
+An `GrpcServerLogic` can be used to handle gRPC requests.
+You need to extend your streamlet from `AkkaServerStreamlet`
+so that Cloudflow will expose an HTTP endpoint in Kubernetes.
+
+Additionally, you need to enable HTTP/2 by adding the following
+to your configuration:
+
+[source,hocon]
+----
+akka.http.server.preview.enable-http2 = on
```
RayRoestenburg, line 123: Would it be possible to 'automate this' so that the user does not have to do it? We do some configuration changes in AkkaStreamlet for cluster config for instance: https://github.com/lightbend/cloudflow/blob/master/core/cloudflow-akka/src/main/scala/cloudflow/akkastream/AkkaStreamlet.scala#L45-L85

examples/snippets/modules/ROOT/examples/akkastreams-grpc-java/src/main/java/sensordata/SensorDataIngress.java
```diff
@@ -0,0 +1,35 @@
...
+        return new GrpcServerLogic(this, getContext()) {
+            public List<Function<HttpRequest, CompletionStage<HttpResponse>>> handlers() {
+                return List.of(
+                        SensorDataServiceHandlerFactory.partial(new SensorDataServiceImpl(), SensorDataService.name, system()),
+                        ServerReflection.create(List.of(SensorDataService.description), system()));
+            }
+        };
+    }
+    //end::logic[]
+}
```
RayRoestenburg, line 35: Could add a newline, not too important.

examples/snippets/modules/ROOT/examples/akkastreams-grpc-java/src/main/java/sensordata/SensorDataIngress.java
```diff
@@ -0,0 +1,35 @@
...
+import java.util.List;
+import java.util.concurrent.CompletionStage;
+
+public class SensorDataIngress extends AkkaServerStreamlet {
+    public StreamletShape shape() {
+        return StreamletShape.createEmpty();
+    }
+
+    //tag::logic[]
+    public AkkaStreamletLogic createLogic() {
```
RayRoestenburg, line 25: 👍 very nice!

examples/snippets/modules/ROOT/examples/akkastreams-grpc-java/src/main/protobuf/sensordata.proto
```diff
@@ -0,0 +1,19 @@
...
+    rpc Provide (SensorData) returns (SensorReply) {}
+}
+
+message SensorData {
+    string payload = 1;
+}
+
+message SensorReply {
+    string message = 1;
+}
```
RayRoestenburg, line 19: newline police

docs/shared-content-source/docs/modules/develop/pages/akka-streamlet-utilities.adoc
raboof: Since we're in charge of creating the ActorSystem this is indeed possible. It does raise some questions: should it be possible for the user to override this? If yes, Play and Lagom also have this 'problem' (and IIRC have different ways to solve it), should we introduce a 3rd approach or follow one of those?

RayRoestenburg (commented):
Looks really good! I think it would be nice if the configuration in application.conf could be done automatically. some nitpicks. I like the idea of separate pages for grpc and http.

docs/shared-content-source/docs/modules/develop/pages/akka-streamlet-utilities.adoc
RayRoestenburg: In this case, a streamlet always runs alone (in a pod /container) somewhere. The user indicates by using `GrpcServerLogic` what they want to do, server gRPC, which as I understand, requires http2. Not setting this would break something, correct? Also, we have a config system in place for cloudflow, they actually can already override anythingthey like at deployment if they really wanted to. Please see https://cloudflow.io/docs/current/develop/cloudflow-configuration.html (they could `set cloudflow.runtimes.akka.config.akka.http.server.preview.enable-http2 = off` in a config file and do `kubectl cloudflow deploy app.json --conf my.conf` although I don't think that makes sense here, though I know too little about gRPC at this point to say @raboof.

raboof: Since this means changing some relatively low-level cloudflow infrastructure, I think it'd be good to do this in a separate PR

RayRoestenburg: Sure no problem!

raboof: filed https://github.com/lightbend/cloudflow/issues/666

RayRoestenburg (approved):

blublinsky:
@raboof about test connecting to the GKE service via a K8s Ingress. HTTP2 support is widely used by a lot of kubernetes tools including NGNIX, Istio, etc. So I think that HTTP2 support is a must. Is this what you are saying by your comment?
Another question. I know that Akka GRPC is using ScalaPB, but it generates slightly different GRPC artifacts, which means that when using this plugin, we can use Akka GRPC, but not https://scalapb.github.io/docs/grpc/. Is this correct? 
Finally, can you, please verify that the client support is generated correctly so that we can use generated artifacts to implement GRPC client inside streamlet

raboof:
> about test connecting to the GKE service via a K8s Ingress. HTTP2 support is widely used by a lot of kubernetes tools including NGNIX, Istio, etc. So I think that HTTP2 support is a must. Is this what you are saying by your comment?

What I'm saying is that while we support HTTP/2 on our pod port, 'out of the box' a GKE Ingress will not allow HTTP/2 all the way from the client to the pod. It looks like this should be possible with additional (not completely trivial) configuration. The issue I created for this, #662, has links to further background information.

> I know that Akka GRPC is using ScalaPB, but it generates slightly different GRPC artifacts, which means that when using this plugin, we can use Akka GRPC, but not https://scalapb.github.io/docs/grpc/. Is this correct?

I expect the classes ScalaPB generates should be usable with https://scalapb.github.io/docs/grpc/, but currently we don't allow enabling the 'grpc' ScapaPB option via Akka gRPC. The reason for that was because we figured Akka gRPC users would want to use Akka gRPC instead of scalapb-grpc. If there are situations where it is useful to generate both then we should just allow it (though I wonder if it wouldn't lead to naming conflicts).

> can you, please verify that the client support is generated correctly so that we can use generated artifacts to implement GRPC client inside streamlet

I have verified the client code is indeed generated (though I haven't tested it works).

blublinsky:
Thanks Arnout

RayRoestenburg:
@raboof I'll merge this, would be great, when you have some time, to follow it up with automatically configuring the http2 preview setting!

raboof:
> would be great, when you have some time, to follow it up with automatically configuring the http2 preview setting!

Right, #666

Issue #620, opened 2020-08-10
Resolve conflicting log implementations in Cloudflow


olofwalker:
The problem here is that several logging frameworks (Logback and log4j) are pulled in by Cloudflow, this is something we have to fix.

There are ways to work around this for users of the current versions., for example, add the following configuration to your Cloudflow application `build.sbt` file.

.settings(
    useCoursier := false,
    excludeDependencies += "org.slf4j" % "slf4j-log4j12",
    excludeDependencies += "log4j"     % "log4j",
)

We have to temporarily disable `Coursier` because of this issue:
https://github.com/coursier/coursier/issues/1590

tringuyen-yw:
`log4j` is used by the local loggers (when testing the pipeline by `sbt runLocal`)

```
excludeDependencies += "org.slf4j" % "slf4j-log4j12",
excludeDependencies += "log4j" % "log4j"
```

will cause an exception when the streamlet loggers kick in:

```
/Users/abc/cloudflow/examples/snippets/modules/ROOT/examples/sensor-data-scala/src/main/blueprint/blueprint.conf verified.
[error] java.util.NoSuchElementException: None.get
[error]         at scala.None$.get(Option.scala:529)
[error]         at scala.None$.get(Option.scala:527)
[error]         at cloudflow.sbt.CloudflowLocalRunnerPlugin$.$anonfun$findLogLibsInPluginClasspath$2(CloudflowLocalRunnerPlugin.scala:183)
[error]         at scala.collection.TraversableLike.$anonfun$map$1(TraversableLike.scala:238)
[error]         at scala.collection.immutable.List.foreach(List.scala:392)
[error]         at scala.collection.TraversableLike.map(TraversableLike.scala:238)
[error]         at scala.collection.TraversableLike.map$(TraversableLike.scala:231)
[error]         at scala.collection.immutable.List.map(List.scala:298)
[error]         at cloudflow.sbt.CloudflowLocalRunnerPlugin$.findLogLibsInPluginClasspath(CloudflowLocalRunnerPlugin.scala:183)
[error]         at cloudflow.sbt.CloudflowLocalRunnerPlugin$.$anonfun$projectSettings$11(CloudflowLocalRunnerPlugin.scala:102)
[error]         at cloudflow.sbt.CloudflowLocalRunnerPlugin$.$anonfun$projectSettings$11$adapted(CloudflowLocalRunnerPlugin.scala:88)
[...]

PR #645, opened 2020-08-19
Fix examples logging

RayRoestenburg:

Fixed logback configuration of examples.

Because of fix https://github.com/lightbend/cloudflow/issues/620 Fix for runLocal, log4j slf4j binding is put in Runtime scope, which means it does not get added into the deployment (which is correct)
It seems that the examples depended on this, logback was not configured correctly. Some examples use `system.log` which is logging as an akka level logger, setting was set to WARN, so no output is then seen if the example logs on INFO level.

No.

examples have been deployed.


ennru (approved):
LGTM.

PR #649, opened 2020-08-20
Fix newline stripping in bootstrap-install-script

RoryStokes:
Fix for newlines being stripped from kubectl output in installer bootstrap script

Update bootstrap-install-script.sh.template to wrap kubectlOutput in quotes.

In certain shells, newlines in the kubectl output were being stripped, resulting in only the first storage class being available for selection.

No

Tested locally and via Azure DevOps agent, where the issue was originally noticed.


olofwalker (approved):
LGTM

RayRoestenburg (approved):
LGTM

PR #655, opened 2020-08-24
cloudflow.runtimes.flink.config support for local run

blublinsky:
Existing setupExecutionEnvironment is limited in the sense, that it does not update configuration in the environment, for example, Azure access key. Instead here we are assuming that all of the configurations necessary are specified in Streamlet config under the path "cloudflow.runtimes.flink.config.flink". In the case of local run, all of the parameters under this path are used to populate Flink configuration and build the environment based on them. In the case of cluster execution the same parameters will be included into flink-conf.yaml
Additionally, if "local.web" configuration is set in Flink.runtime or streamlet configuration, local environment is started with web support.
-->

Example are:

> EY request to run local with checkpointing to Azure

> EY request for local Web UI for local run.

No

Local runs

### DIFF
--- a/core/cloudflow-flink/src/main/scala/cloudflow/flink/FlinkStreamlet.scala
+++ b/core/cloudflow-flink/src/main/scala/cloudflow/flink/FlinkStreamlet.scala
@@ -19,21 +19,21 @@ package cloudflow.flink
 import java.nio.file.Path
 
 import scala.concurrent.ExecutionContext.Implicits.global
-import scala.concurrent.{ Future, Promise }
-import scala.util.{ Failure, Try }
-
+import scala.concurrent.{Future, Promise}
+import scala.util.{Failure, Try}
 import com.typesafe.config.Config
 import net.ceedubs.ficus.Ficus._
 import org.apache.flink.api.common.JobExecutionResult
 import org.apache.flink.api.common.typeinfo.TypeInformation
 import org.apache.flink.client.program.OptimizerPlanEnvironment
 import org.apache.flink.streaming.api.CheckpointingMode
-import org.apache.flink.streaming.api.datastream.{ DataStreamSink, DataStream ⇒ JDataStream }
+import org.apache.flink.streaming.api.datastream.{DataStreamSink, DataStream => JDataStream}
 import org.apache.flink.streaming.api.environment.CheckpointConfig
 import org.apache.flink.streaming.api.scala._
-
 import cloudflow.streamlets.BootstrapInfo._
 import cloudflow.streamlets._
+import org.apache.flink.configuration.Configuration
+import org.apache.flink.core.fs.FileSystem
 
 /**
  * The base class for defining Flink streamlets. Derived classes need to override `createLogic` to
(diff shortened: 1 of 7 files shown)


core/cloudflow-flink/src/main/scala/cloudflow/flink/FlinkStreamlet.scala
```diff
@@ -80,33 +80,47 @@ abstract class FlinkStreamlet extends Streamlet[FlinkStreamletContext] with Seri
...
+                                    createExecutionEnvironment(updatedConfig, streamletDefinition.streamletRef),
+                                    updatedConfig)
     }).recoverWith {
       case th ⇒ Failure(new Exception(s"Failed to create context from $config", th))
     }.get
 
+  /**
+   * Populate Flink configuration from Streamlet Flink configuration
+   */
+  private def populateFlinkConfiguration(configuration: Configuration, config: Config): Unit =
```
olofwalker, line 93: `configs: Config*` and then `configs.forEach` ?

blublinsky:
Add support for Access the Flink Web UI in local mode#208

core/cloudflow-flink/src/main/scala/cloudflow/flink/FlinkStreamlet.scala
blublinsky: Same thing

olofwalker: I meant that you can pass in more than one config with each call.

```
populateFlinkConfiguration(configuration, config.atPath(ClusterFlinkJobExecutor.flinkRuntime), config.atPath(ClusterFlinkJobExecutor.streamletRuntimeConfigPath(streamlet)))
```

blublinsky: Yea, but I need to preserve the order. Need to make sure that streamlet overwrite runtime. Also, need to know the path to remove it from the key (yesterday's update)

core/cloudflow-flink/src/main/scala/cloudflow/flink/FlinkStreamlet.scala
```diff
@@ -80,24 +80,64 @@ abstract class FlinkStreamlet extends Streamlet[FlinkStreamletContext] with Seri
...
     }.get
 
   /**
-   * Override this method to setup the StreamExecutionEnvironment. By default it does not modify the [[StreamExecutionEnvironment]] created
-   * by [[createExecutionEnvironment]].
-   * @see [[createExecutionEnvironment]] which creates the StreamExecutionEnvironment used in this FlinkStreamlet.
+   * Populate Flink configuration from Streamlet Flink configuration
    */
-  protected def setupExecutionEnvironment(env: StreamExecutionEnvironment): StreamExecutionEnvironment = env
+  private def populateFlinkConfiguration(configuration: Configuration, config: Config, path: String): Boolean = {
```
debasishg, line 93: I would love to see this as a method that returns the updated configuration instead of a `Boolean`. The name `populateFlinkConfiguration` always makes me think it will give me an updated configuration.

core/cloudflow-flink/src/main/scala/cloudflow/flink/FlinkStreamlet.scala
```diff
@@ -80,24 +80,64 @@ abstract class FlinkStreamlet extends Streamlet[FlinkStreamletContext] with Seri
...
+        // Ensures that filesystem is initialized with right configuration
+        FileSystem.initialize(configuration, null)
+        // Create local Flink environment
+        // Note here that a local web support is set through configuration by setting
+        // "local.web" either in the streamlet or Flink runtime context. We are not checking the value, but rather
+        // the fact that variable is defines.
+        enableWeb match {
+          case true => StreamExecutionEnvironment.createLocalEnvironmentWithWebUI(configuration)
+          case _    => StreamExecutionEnvironment.createLocalEnvironment(1, configuration)
+        }
```
debasishg, line 137: nitpick: I would do an if-then-else here. Pattern matching is probably more expensive

core/cloudflow-flink/src/main/scala/cloudflow/flink/FlinkStreamlet.scala
blublinsky: Well, possibly, but...
Boolean here is whether we need WebUI or not. Returning back configuration will make code a bit more verbose and will use more memory

core/cloudflow-flink/src/main/scala/cloudflow/flink/FlinkStreamlet.scala
blublinsky: Looks more "scalish" to me, but I might be wrong

core/cloudflow-flink/src/main/scala/cloudflow/flink/FlinkStreamlet.scala
olofwalker: I think @debasishg  has a point here, it would be nicer to return a new version of the config, you can also return any control information and use it like `val (newConfig, needWebUI) = populateFlinkConfiguration(...)`

core/cloudflow-flink/src/main/scala/cloudflow/flink/FlinkStreamlet.scala
olofwalker: It will preserve the order with an array, right_

core/cloudflow-flink/src/main/scala/cloudflow/flink/FlinkStreamlet.scala
justinhj: I think it's only really scala-ish if it's some kind of union type, with a simple boolean I think if is more common

blublinsky: You guys are so hard to please, done

RayRoestenburg (commented):
Added some more code, this still needs a documentation update.

RayRoestenburg:
@blublinsky Would be great if you can add some documentation to explain these runLocal FLink options, in a separate PR?

PR #667, opened 2020-08-26
Add Spark trigger support

blublinsky:
Adding Spark streaming execution trigger as a parameter for sparklet execution
As described https://dzone.com/articles/spark-trigger-options,  Spark  Structured Streaming supports different trigger options available in Spark Structured Streaming. Setting the right trigger for a stream will decide how quick your stream reads the next set of records. The trigger option enables you to control the latency and throughput of the job.

In the current implementation, no option was set for the execution trigger which means that Spark will try to process the next set of records as soon as it’s done with the current micro batch. The default behavior of write streams in Spark Structured Streaming is the micro batch. In a micro batch, incoming records are grouped into small windows and processed in a periodic fashion.

Although processing as soon as possible, is sometimes a good option, but the most widely used and recommended practice in Spark Structured Streaming. The trigger option of processing time gives you better control over how often micro batch jobs should get triggered. For example, if you set a trigger interval of “20 seconds” then for every 20 seconds a micro batch will process a batch of records. Based on the speed at which you want your job to process the records you can fine-tune the trigger interval and throughput of the job as well.
Another very useful processing option is continuous stream processing was introduced in Spark 2.3 as an experimental feature. As of Spark 2.4.0, it’s still experimental. It enables you to process records in milliseconds which would otherwise take seconds in micro batch. In this trigger option, records are not processed in micro batch, but, instead, a long-running task is created per write stream and processed as quickly as possible.

The ability to pass an explicit Trigger allows to better utilize Spark streaming capabilities and better utilize sparklets

No

Re running all the tests

### DIFF
--- a/core/cloudflow-spark-testkit/src/main/scala/cloudflow/spark/testkit/TestSparkStreamletContext.scala
+++ b/core/cloudflow-spark-testkit/src/main/scala/cloudflow/spark/testkit/TestSparkStreamletContext.scala
@@ -57,7 +57,7 @@ class TestSparkStreamletContext(override val streamletRef: String,
       stream: Dataset[Out],
       outPort: CodecOutlet[Out],
       outputMode: OutputMode,
-      optional_trigger: Option[Trigger] = None
+      optionalTrigger: Option[Trigger] = None
   )(implicit encoder: Encoder[Out], typeTag: TypeTag[Out]): StreamingQuery = {
     // RateSource can only work with a microBatch query because it contains no data at time zero.
     // Trigger.Once requires data at start to work.
--- a/core/cloudflow-spark-testkit/src/main/scala/cloudflow/spark/testkit/TestSparkStreamletContext.scala
+++ b/core/cloudflow-spark-testkit/src/main/scala/cloudflow/spark/testkit/TestSparkStreamletContext.scala
@@ -57,15 +57,18 @@ class TestSparkStreamletContext(override val streamletRef: String,
       stream: Dataset[Out],
       outPort: CodecOutlet[Out],
       outputMode: OutputMode,
-      trigger: Trigger = Trigger.ProcessingTime(0L)
+      optional_trigger: Option[Trigger] = None
   )(implicit encoder: Encoder[Out], typeTag: TypeTag[Out]): StreamingQuery = {
     // RateSource can only work with a microBatch query because it contains no data at time zero.
     // Trigger.Once requires data at start to work.
-    var etrigger = trigger
-    etrigger = if (isRateSource(stream)) {
-      Trigger.ProcessingTime(ProcessingTimeInterval)
-    } else {
-      Trigger.Once()
+    val trigger = optional_trigger match {
+      case Some(t) => t
+      case _ =>
+        if (isRateSource(stream)) {
+          Trigger.ProcessingTime(ProcessingTimeInterval)
+        } else {
+          Trigger.Once()
+        }
     }
     val streamingQuery = outletTaps
       .find(_.portName == outPort.name)
@@ -73,7 +76,7 @@ class TestSparkStreamletContext(override val streamletRef: String,
         stream.writeStream
           .outputMode(outputMode)
           .format("memory")
-          .trigger(etrigger)
+          .trigger(trigger)
           .queryName(outletTap.queryName)
           .start()
       }
--- a/core/cloudflow-spark-testkit/src/main/scala/cloudflow/spark/testkit/TestSparkStreamletContext.scala
+++ b/core/cloudflow-spark-testkit/src/main/scala/cloudflow/spark/testkit/TestSparkStreamletContext.scala
@@ -53,12 +53,16 @@ class TestSparkStreamletContext(override val streamletRef: String,
       .map(_.instream.asInstanceOf[MemoryStream[In]].toDF.as[In])
       .getOrElse(throw TestContextException(inPort.name, s"Bad test context, could not find source for inlet ${inPort.name}"))
 
-  override def writeStream[Out](stream: Dataset[Out],
-                                outPort: CodecOutlet[Out],
-                                outputMode: OutputMode)(implicit encoder: Encoder[Out], typeTag: TypeTag[Out]): StreamingQuery = {
+  override def writeStream[Out](
+      stream: Dataset[Out],
+      outPort: CodecOutlet[Out],
+      outputMode: OutputMode,
+      trigger: Trigger = Trigger.ProcessingTime(0L)
+  )(implicit encoder: Encoder[Out], typeTag: TypeTag[Out]): StreamingQuery = {
     // RateSource can only work with a microBatch query because it contains no data at time zero.
     // Trigger.Once requires data at start to work.
-    val trigger = if (isRateSource(stream)) {
+    var etrigger = trigger
+    etrigger = if (isRateSource(stream)) {
       Trigger.ProcessingTime(ProcessingTimeInterval)
     } else {
       Trigger.Once()
@@ -69,7 +73,7 @@ class TestSparkStreamletContext(override val streamletRef: String,
         stream.writeStream
           .outputMode(outputMode)
           .format("memory")
-          .trigger(trigger)
+          .trigger(etrigger)
           .queryName(outletTap.queryName)
           .start()
       }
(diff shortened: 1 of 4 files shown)


core/cloudflow-spark-testkit/src/main/scala/cloudflow/spark/testkit/TestSparkStreamletContext.scala
```diff
@@ -53,12 +53,16 @@ class TestSparkStreamletContext(override val streamletRef: String,
...
+      stream: Dataset[Out],
+      outPort: CodecOutlet[Out],
+      outputMode: OutputMode,
+      trigger: Trigger = Trigger.ProcessingTime(0L)
+  )(implicit encoder: Encoder[Out], typeTag: TypeTag[Out]): StreamingQuery = {
     // RateSource can only work with a microBatch query because it contains no data at time zero.
     // Trigger.Once requires data at start to work.
-    val trigger = if (isRateSource(stream)) {
+    var etrigger = trigger
+    etrigger = if (isRateSource(stream)) {
```
maasg, line 65: As we are making the trigger part of the interface, I don't think we can override it under the feet of the user.
At most, we could warn the user that the combination `RateSource && Trigger(Once)` will not work.

core/cloudflow-spark/src/main/scala/cloudflow/spark/SparkStreamletContext.scala
```diff
@@ -52,10 +52,13 @@ abstract case class SparkStreamletContext(
    * @param stream stream used to write the result of execution of the `StreamingQuery`
    * @param outPort the port used to write the result of execution of the `StreamingQuery`
    * @param outputMode the output mode used to write. Valid values Append, Update, Complete
+   * @param trigger Execution trigger (see https://dzone.com/articles/spark-trigger-options for details)
```
maasg, line 55: I would rather point to the official documentation.

blublinsky: Will do

maasg:
Hi @blublinsky!!! 
Thanks for adding support for triggers!. Looks good, except for the override in the `TestContext` - I left a note explaining the issue.

core/cloudflow-spark-testkit/src/main/scala/cloudflow/spark/testkit/TestSparkStreamletContext.scala
blublinsky: I am opened to your suggestions. I just hacked this to make it compile, but I will let you to do a better implementation

blublinsky: Ok, I think I fixed it

blublinsky:
Hi Gerard, I think this version is better. Trigger now is optional and if it is None, TestContext will set it as before. If a user wants to overwrite it, we are using provided value.

core/cloudflow-spark-testkit/src/main/scala/cloudflow/spark/testkit/TestSparkStreamletContext.scala
```diff
@@ -53,15 +53,22 @@ class TestSparkStreamletContext(override val streamletRef: String,
...
       .getOrElse(throw TestContextException(inPort.name, s"Bad test context, could not find source for inlet ${inPort.name}"))
 
-  override def writeStream[Out](stream: Dataset[Out],
-                                outPort: CodecOutlet[Out],
-                                outputMode: OutputMode)(implicit encoder: Encoder[Out], typeTag: TypeTag[Out]): StreamingQuery = {
+  override def writeStream[Out](
+      stream: Dataset[Out],
+      outPort: CodecOutlet[Out],
+      outputMode: OutputMode,
+      optional_trigger: Option[Trigger] = None
```
RayRoestenburg, line 60:
```suggestion
      optionalTrigger: Option[Trigger] = None
```

core/cloudflow-spark-testkit/src/main/scala/cloudflow/spark/testkit/TestSparkStreamletContext.scala
```diff
@@ -53,15 +53,22 @@ class TestSparkStreamletContext(override val streamletRef: String,
...
+  )(implicit encoder: Encoder[Out], typeTag: TypeTag[Out]): StreamingQuery = {
     // RateSource can only work with a microBatch query because it contains no data at time zero.
     // Trigger.Once requires data at start to work.
-    val trigger = if (isRateSource(stream)) {
-      Trigger.ProcessingTime(ProcessingTimeInterval)
-    } else {
-      Trigger.Once()
+    val trigger = optional_trigger match {
```
RayRoestenburg, line 64:
```suggestion
    val trigger = optionalTrigger match {
```

core/cloudflow-spark/src/main/scala/cloudflow/spark/SparkStreamlet.scala
```diff
@@ -221,10 +221,13 @@ abstract class SparkStreamletLogic(implicit val context: SparkStreamletContext)
   /**
    * Write a `StreamingQuery` into outlet using the specified `OutputMode`
    */
-  final def writeStream[Out](stream: Dataset[Out], outPort: CodecOutlet[Out], outputMode: OutputMode)(
+  final def writeStream[Out](stream: Dataset[Out],
+                             outPort: CodecOutlet[Out],
+                             outputMode: OutputMode,
+                             optional_trigger: Option[Trigger] = None)(
```
RayRoestenburg, line 227:
```suggestion
                             optionalTrigger: Option[Trigger] = None)(
```

core/cloudflow-spark/src/main/scala/cloudflow/spark/SparkStreamlet.scala
```diff
@@ -221,10 +221,13 @@ abstract class SparkStreamletLogic(implicit val context: SparkStreamletContext)
...
       implicit encoder: Encoder[Out],
       typeTag: TypeTag[Out]
-  ): StreamingQuery = context.writeStream(stream, outPort, outputMode)
+  ): StreamingQuery = context.writeStream(stream, outPort, outputMode, optional_trigger)
```
RayRoestenburg, line 230:
```suggestion
  ): StreamingQuery = context.writeStream(stream, outPort, outputMode, optionalTrigger)
```

core/cloudflow-spark/src/main/scala/cloudflow/spark/kafka/SparkStreamletContextImpl.scala
```diff
@@ -73,8 +73,10 @@ class SparkStreamletContextImpl(
     case (key, value) => s"kafka.$key" -> value
   }
 
-  def writeStream[Out](stream: Dataset[Out], outPort: CodecOutlet[Out], outputMode: OutputMode)(implicit encoder: Encoder[Out],
-                                                                                                typeTag: TypeTag[Out]): StreamingQuery = {
+  def writeStream[Out](stream: Dataset[Out], outPort: CodecOutlet[Out], outputMode: OutputMode, optional_trigger: Option[Trigger] = None)(
```
RayRoestenburg, line 76:
```suggestion
  def writeStream[Out](stream: Dataset[Out], outPort: CodecOutlet[Out], outputMode: OutputMode, optionalTrigger: Option[Trigger] = None)(
```

core/cloudflow-spark/src/main/scala/cloudflow/spark/kafka/SparkStreamletContextImpl.scala
```diff
@@ -87,15 +89,29 @@ class SparkStreamletContextImpl(
...
-    encodedStream.writeStream
-      .outputMode(outputMode)
-      .format("kafka")
-      .queryName(queryName)
-      .option("kafka.bootstrap.servers", brokers)
-      .options(kafkaProducerMap(topic))
-      .option("topic", destTopic)
-      .option("checkpointLocation", checkpointLocation)
-      .start()
+    optional_trigger match {
```
RayRoestenburg, line 92:
```suggestion
    optionalTrigger match {
```

PR #668, opened 2020-08-26
Akka cqrs, persistance, and projection sample

nolangrace:

New cloudflow sample build using Akka sample crqs shopping cart and Akka Projections

Help users understand that the full Akka ecosystem is available in cloudflow.  Addresses #664 

no


examples/shopping-cart-cqrs/akka-shopping-cart-streamlet/src/main/scala/shoppingcart/streamlets/ShoppingCartStreamlet.scala
```diff
@@ -0,0 +1,100 @@
...
+        .map( x =>
+          x.event match {
+            case e: ItemAdded => ShoppingCartEvent(e.cartId, e.itemId, e.quantity, 0, "ItemAdded")
+            case e: ItemRemoved => ShoppingCartEvent(e.cartId, e.itemId, 0, 0, "ItemRemoved")
+            case e: ItemQuantityAdjusted => ShoppingCartEvent(e.cartId, e.itemId, e.newQuantity, 0, "ItemQuantityAdjusted")
+            case e: CheckedOut => ShoppingCartEvent(e.cartId, "", 0, System.currentTimeMillis(), "CheckedOut")
+            case _ => ShoppingCartEvent("", "", 0, 0, "Error")
+          })
+        .mapAsync(1) { x =>
+          sinkRef(out).write(x)
```
nolangrace, line 90: @RayRoestenburg and @ennru this `sinkRef(out).write(x)` from the cloudflow api is what I am using to be able write a message to the cloudflow outlet before akka projections commits the offset.

This function uses the `KafkaProducer(...).send( )` from the Kafka client to write to kafka.  

For this example would it be better to use an AkkaProjection actor, Source.queue, and a MergeHub?  I was looking at the KafkaProducer documentation and I am pretty sure this operation can handle many parallel streams calling it at once but I wanted to hear your thoughts.

nolangrace:
@RayRoestenburg do you think examples should generate their own load in a streamlet or should we provide a script to make API calls.  The cloudflow examples have both so I was just curious what you prefer

RayRoestenburg:
@nolangrace It's easiest to just have some generators, that way someone can just start the thing and see it working.

blublinsky:
I do not see what projections adds here. Would it be simpler, instead of using a persistent Actor writing to Cassandra and then reading Cassandra journal to write to Kafka, to have a very simple streamlet, that reads HTTP and writes it directly to Kafka. It would be much easier and faster and use by far less resources.

nolangrace:
As far as Akka Projections is concerned if we could use a cloudflow kafka topic as a persistence backend then basically any streamlet reading from that persistence journal topic would be performing the same function as Akka Projections.  So we are in complete agreement there.

But long story short I agree that using kafka as the persistence backend would be great but until Akka persistence supports kafka I think this is the best solution.  Especially considering the goal of this sample is to demonstrate how the full world of Akka is available in cloudflow.

RayRoestenburg:
Closing this for now, will revisit once we pickup akka persistence support. Thanks @nolangrace !

Issue #671, opened 2020-08-30
Deployment of an application may create unnecessary resources.


olofwalker:
When deploying an application, the namespace is populated with all resources that may be used. The operator should inspect the application being deployed before creating resources. 

For example, if you deploy a Flink only application, there is no need for creating Spark `rolebindings`, etc.

PR #672, opened 2020-08-30
Prepare the namespace of the application depending on the runtime that is used.

olofwalker:
As detailed in #671, the Cloudflow operator currently creates a fixed set of resources when deploying an application, no matter if they are needed or not. 

This draft/sketch PR shows the problem and attempts on a fix.


RayRoestenburg:
@olofwalker I changed the approach a bit and rewrote the commit, due to some git rebase issues...

RayRoestenburg:
I refactored some stuff and then noticed that these actions are really runner specific and moved it to the runners.

RayRoestenburg:
Also changed it so that Flink and Spark get their own PVCs. We're working on an improvement so that these PVCs are easier to specify, and change, so the install is more modular (can be done in any order)

RayRoestenburg (approved):

RayRoestenburg:
I've tested this on GKE

PR #686, opened 2020-09-03
Update example to enable logging

blublinsky:
/## What changes were proposed in this pull request?
With the latest version of cloudflow (2.0.9), the log stopped working.
This seems to be related to Akka upgrade. The logger used here is from Akka logging
https://doc.akka.io/docs/akka/current/logging.html
The documentation https://doc.akka.io/docs/akka/current/logging.html#loggers suggest using Akka configuration to enable 
SLF, but it did not work for me, hence this change.
This is just a quick fix to make it work, but further investigation is required.

To enable working when running on the cluster. Local run still works

no

Running both on cluster and locally to ensure that it works


blublinsky:
One question. Do you have application.conf somewhere, which I could not overwrite?

examples/snippets/modules/ROOT/examples/sensor-data-scala/build.sbt
```diff
@@ -14,7 +14,7 @@ lazy val sensorData =  (project in file("."))
       libraryDependencies ++= Seq(
         "com.lightbend.akka"     %% "akka-stream-alpakka-file"  % "1.1.2",
         "com.typesafe.akka"      %% "akka-http-spray-json"      % "10.1.12",
-        "ch.qos.logback"         %  "logback-classic"           % "1.2.3",
+//        "ch.qos.logback"         %  "logback-classic"           % "1.2.3",
```
RayRoestenburg, line 17: Always remove commented lines please.

examples/snippets/modules/ROOT/examples/sensor-data-scala/src/main/resources/log4j.xml
```diff
@@ -0,0 +1,18 @@
...
+                   value="%d{yyyy-MM-dd HH:mm:ss} %-5p %c{1}:%L - %m%n" />
+        </layout>
+    </appender>
+
+    <root>
+        <level value="DEBUG" />
+        <appender-ref ref="console" />
+    </root>
+
+</log4j:configuration>
```
RayRoestenburg, line 18:
```suggestion
</log4j:configuration>

```

RayRoestenburg:
This is a quick fix. Making sure that log4j is not added to the classpath is an open issue: #620

Issue #685, opened 2020-09-03
Named Kafka cluster configurations for topic blueprint


seglo:
Kafka connection details can either be provided during installation (as simply, `bootstrap.servers`) or explicitly within a blueprint using advanced topic configuration.  In addition, we should let users define a reusable named Kafka connection configuration for as many clusters as they like.  The blueprint can then reference these configurations by a simple name.

A Kafka connection configuration will include the following information:

* name
* `bootstrap.servers` connection string
* producer client config
* consumer client config

During an application deployment the cloudflow operator will first lookup any referenced Kafka connection configurations and propagate them along to streamlets as they're created.

If a blueprint doesn't reference any Kafka cluster configurations then the current behaviour will be preserved. The order of precedence to load a configuration in this case will be:

* use the `default` Kafka cluster configuration, if it exists
* use the advanced connection info provided in the blueprint, if it exists
* use the `bootstrap.servers` configuration passed to the operator during install

Before deployment the CLI will verify that secrets containing any referenced Kafka cluster configurations exist.

RayRoestenburg:
Looks like Robert did already write some things down in #669 maybe combine this @seglo.

RayRoestenburg:
Would be nice, if you can set the TLS stuff in a shared connection config (not having to specify it as producer and consumer config) 

I would be fine with using hocon in the secret data  if that makes it easier.

RayRoestenburg:
(the blueprint and config files already have support for these kind of settings, so maybe it makes sense for consistency to use the same/similar format)

seglo:
> Would be nice, if you can set the TLS stuff in a shared connection config (not having to specify it as producer and consumer config)
> 
> I would be fine with using hocon in the secret data if that makes it easier.

Yes, I'm thinking about using a secret with a HOCON config file that uses the same config as in the blueprint.  We can update the chart to create a (`default`) if config is provided.  We can also use the chart to update the install with as many configs as we want, so the admin user doesn't have to figure out how to template and create the secrets themselves.

RayRoestenburg:
That sounds great.

seglo:
It looks like `helm` can't manage the creation of secrets. There's a [`helm-secrets`](https://github.com/zendesk/helm-secrets) plugin, but we don't want to saddle the user with installing that with helm.

I'll look into creating the secrets from the CLI as @olofwalker originally proposed in #669.

RayRoestenburg:
[...]

PR #687, opened 2020-09-04
Ray/681/include server kafka cm

RayRoestenburg:

- add bootstrap-servers to configmap created for managed kafka topic.
- made default bootstrap-servers optional, so you can use cloudflow without providing `cloudflow_operator.kafkaBootstrapservers` in helm charts install, as long as you do not use managed topics. (This will be improved with https://github.com/lightbend/cloudflow/issues/685

So it is clear where the topic is in the config map, and to make sure you can use cloudflow without specifying a `cloudflow_operator.kafkaBootstrapservers` 

No.

Unit tests. 
- [x] Testing on GKE.

### DIFF
--- a/core/cloudflow-streamlets/src/main/scala/cloudflow/streamlets/StreamletContext.scala
+++ b/core/cloudflow-streamlets/src/main/scala/cloudflow/streamlets/StreamletContext.scala
@@ -20,8 +20,11 @@ import java.nio.file.{ Path, Paths }
 
 import com.typesafe.config.Config
 import com.typesafe.config.ConfigFactory
+import org.slf4j.LoggerFactory
 
 trait StreamletContext {
+
+  private lazy val log = LoggerFactory.getLogger(getClass.getName)
 
   /**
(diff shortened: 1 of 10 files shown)


core/cloudflow-streamlets/src/main/scala/cloudflow/streamlets/StreamletContext.scala
```diff
@@ -59,7 +62,21 @@ trait StreamletContext {
    * The default bootstrapServers for the Kafka broker that has been installed or configured
    * to be used globally for all Cloudflow applications
    */
-  def internalKafkaBootstrapServers = config.getString("cloudflow.kafka.bootstrap-servers")
+  def internalKafkaBootstrapServers =
+    try {
+      config.getString("cloudflow.kafka.bootstrap-servers")
+    } catch {
+      case e: com.typesafe.config.ConfigException.Missing =>
```
ennru, line 69: Testing for the config path would show the intention cleaner then try/catch, IMHO.

ennru (approved):
LGTM.

PR #694, opened 2020-09-08
operator: separate Flink and Spark storage settings

ennru:

Make the default storage claim settings for Flink and Spark independent of each other.

The storage claims need to become easier adjustable per application.

### DIFF
--- a/core/cloudflow-operator/src/main/resources/application.conf
+++ b/core/cloudflow-operator/src/main/resources/application.conf
@@ -109,6 +109,8 @@ cloudflow {
         # Do not specify max heap, this is calculated by memory-overhead
         java-opts = ""
         java-opts = ${?SPARK_EXECUTOR_JAVA_OPTS}
+
+        persistent-storage = ${cloudflow.platform.deployment.persistent-storage}
       }
 
       flink-runner {
@@ -160,6 +162,8 @@ cloudflow {
           java-opts = ""
           java-opts = ${?FLINK_TASKMANAGER_JAVA_OPTS}
         }
+
+        persistent-storage = ${cloudflow.platform.deployment.persistent-storage}
       }
 
       persistent-storage {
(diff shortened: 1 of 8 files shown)


core/cloudflow-operator/src/main/scala/cloudflow/operator/DeploymentContext.scala
```diff
@@ -29,15 +29,13 @@ case class DeploymentContext(kafkaContext: KafkaContext,
...
-                             persistentStorageSettings: PersistentStorageSettings,
                              podName: String,
                              podNamespace: String) {
-  import kafkaContext._
   def infoMessage = s"""
    | pod-name:                         ${podName}
    | pod-namespace                     ${podNamespace}
    |
-   | kafka-bootstrap-servers:          ${bootstrapServers}
+   | kafka-bootstrap-servers:          ${kafkaContext.bootstrapServers}
```
RayRoestenburg, line 38: yeah more readable

core/cloudflow-operator/src/main/resources/application.conf
```diff
@@ -109,6 +109,8 @@ cloudflow {
         # Do not specify max heap, this is calculated by memory-overhead
         java-opts = ""
         java-opts = ${?SPARK_EXECUTOR_JAVA_OPTS}
+
+        persistent-storage = ${cloudflow.platform.deployment.persistent-storage}
```
RayRoestenburg, line 113: Maybe this is a good time to make this more consistent with flink-runner config, and have:
```hocon
spark-runner {
  driver {
    # contents of spark-runner-driver
  }
  executor {
    # contents of spark-runner-executor
  }
  persistent-storage = ${cloudflow.platform.deployment.persistent-storage}
}
```

RayRoestenburg (commented):
I think it would be nice to fix the conf file slightly to make spark runner settings more consistent with flink settings

ennru:
> I think it would be nice to fix the conf file slightly to make spark runner settings more consistent with flink settings

Agreed, but how would that be backwards compatible?

blublinsky:
Considering that we now allow overwriting of Flink checkpointing through configuration, do we still need this?

RayRoestenburg:
@blublinsky it's not only about Flink of course. https://github.com/lightbend/cloudflow/issues/710 improves on the approach here, the operator should not have storage settings at all.

PR #707, opened 2020-09-10
Recommend 'gcloud auth login' for authentication

raboof:
This is also the recommended way when you have installed via GCP Marketplace,
right?


raboof:
nevermind, I needed `--no-registry-credentials`

PR #719, opened 2020-09-16
Modified Kafka settings in strimzi example, noted this is purely for …

RayRoestenburg:
…testing.

Modified strimzi Kafka cluster configuration. 

Users quickly run into Kafka running out of space.

docs

### DIFF
--- a/docs/docs-source/docs/modules/administration/pages/how-to-install-and-use-strimzi.adoc
+++ b/docs/docs-source/docs/modules/administration/pages/how-to-install-and-use-strimzi.adoc
@@ -47,7 +47,7 @@ strimzi-cluster-operator-9968fd8c9-fhqmj   1/1     Running   0          17s
 
 To create a Kafka cluster using Strimzi, we have to create a `CustomResource` of the kind `Kafka` in the namespace `cloudflow`. 
 
-NOTE: The Kafka cluster configuration shown here is meant for development and testing purposes only.
+NOTE: The Kafka cluster configuration shown here is meant for development and testing purposes only. Please consider your storage requirements and modify the custom resource accordingly for your intended usage.
 
 Create the YAML file containing the Kafka custom resource using the following snippet:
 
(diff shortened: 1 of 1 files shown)


docs/docs-source/docs/modules/administration/pages/how-to-install-and-use-strimzi.adoc
```diff
@@ -62,8 +64,9 @@ spec:
     config:
       auto.create.topics.enable: false
       log.message.format.version: "2.3"
-      log.retention.bytes: 112197632
-      log.retention.hours: -1
+      log.retention.bytes: 1073741824
+      log.retention.hours: 1
```
huntc, line 68: I think this is good - at least there is a limit on retention now.

docs/docs-source/docs/modules/administration/pages/how-to-install-and-use-strimzi.adoc
```diff
@@ -80,7 +83,7 @@ spec:
     resources: {}
     storage:
       deleteClaim: false
-      size: 1Gi
+      size: 100Gi
```
huntc, line 86: My head is in the embedded world still. Crazy numbers! ;-)

huntc (approved):

huntc:
Actually, can we not use log compaction to achieve the retainment of just he last message?

RayRoestenburg:
Yes we can configure log.compaction , I think that would be ok. It would keep the last message for a key, so it depends on your partitioning logic if that helps in any way. I've added it though.

huntc:
> Yes we can configure log.compaction , I think that would be ok. It would keep the last message for a key, so it depends on your partitioning logic if that helps in any way. I've added it though.

Nice one. That should help a lot.

RayRoestenburg:
I tested the config on GKE, the kafka config in /tmp/strimzi.properties reflects this:
```
kafka.Kafka /tmp/strimzi.properties
```
```
tail -12 /tmp/strimzi.properties
##########
# User provided configuration
##########
auto.create.topics.enable=false
log.cleanup.policy=compact
log.message.format.version=2.3
log.retention.bytes=1073741824
log.retention.check.interval.ms=300000
log.retention.hours=1
offsets.topic.replication.factor=3
transaction.state.log.min.isr=2
transaction.state.log.replication.factor=3
```

Issue #722, opened 2020-09-16
operator: Response entity was not subscribed after 1 second


raboof:
**Describe the bug**

The operator crashes with:

```
2020-09-16 13:16:38,689 WARN  [PoolGateway] - [0 (WaitingForResponseEntitySubscription)] Response entity was not subscribed after 1 second. Make sure to read the response entity body or call `discardBytes()` on it. GET /apis/cloudflow.lightbend.com/v1alpha1/namespaces//cloudflowapplications Empty -> 200 OK Chunked
2020-09-16 13:16:40,289 INFO  [ActorSystemImpl] - [app: akkastreams-grpc-scala application configuration changed (Secret config-akkastreams-grpc-scala ADDED)]
2020-09-16 13:16:40,986 ERROR [Materializer] - [app-event] Upstream failed.
java.util.concurrent.TimeoutException: Response entity was not subscribed after 1 second. Make sure to read the response entity body or call `discardBytes()` on it. GET /apis/cloudflow.lightbend.com/v1alpha1/namespaces//cloudflowapplications Empty -> 200 OK Chunked
	at akka.http.impl.engine.client.pool.SlotState$WaitingForResponseEntitySubscription.onTimeout(SlotState.scala:317)
```

We have seen this before in #441. There the theory was that downgrading to Akka 2.5.29 would resolve the issue. However, here we see the issue again on version 704-d787efb of the operator, which is on Akka 2.5.29.

[logs.txt](https://github.com/lightbend/cloudflow/files/5232446/logs.txt)

raboof:
One possible explanation could be https://github.com/doriordan/skuber/blob/b551d675cbe2dd19c56442786d4e0143f64e3048/client/src/main/scala/skuber/api/watch/WatchSource.scala#L68, where a stream based on the entity is passed to `Flow.flatMapConcat` and backpressure might delay the materialization. On the other hand, I don't think we expect backpressure in that scenario, since I think a new request is only triggered when the previous request has completed...

PR #726, opened 2020-09-17
Higher timeout setting for subscription to response entity.

RayRoestenburg:

Increase timeout for subscribing to response in the operator.

Some users report:

```
TimeoutException: Response entity was not subscribed after 1 second. Make sure to read the response entity body or call `discardBytes()` on it. 
```
It could be, because of some backpressure (the idea is that flatMapConcat in [WatchSource](https://github.com/doriordan/skuber/blob/b551d675cbe2dd19c56442786d4e0143f64e3048/client/src/main/scala/skuber/api/watch/WatchSource.scala#L68), after doing request could result in some time between request and starting to read from response), or where at times the operator gets less resources in k8s, it just takes a little more time.
This PR sets the timeout to 60s.

No.

- [x] GKE.


olofwalker (approved):

raboof:
refs #722

Issue #356, opened 2020-04-23
Separate runtime operators


michaelpnash:
https://docs.google.com/document/d/1_NEOKnmrh42eXn46whhQ9Vc5yBG__7duH-sYu5evstM/edit

PR #739, opened 2020-09-23
spike of Akka Microservice operator

patriknw:
Only for discussion, not intended to be merged.

Based on Cloudflow operator.

So far this works with the shopping-cart-service. It can form an Akka Cluster with Bootstrap.

See commits for the incremental steps...

### DIFF
--- a/core/cloudflow-operator/src/test/kubernetes/shopping-cart-cr.json
+++ b/core/cloudflow-operator/src/test/kubernetes/shopping-cart-cr.json
@@ -65,7 +65,7 @@
       {
         "class_name": "shopping.cart.Main",
         "config": {},
-        "image": "docker.io/patriknw/shopping-cart-service-scala:0.1-39-141c65c6-20200916-1227",
+        "image": "docker.io/patriknw/shopping-cart-service-scala:0.2-1-2472dd18",
         "name": "shopping-cart-service.main",
         "port_mappings": {},
         "runtime": "akka-microservice",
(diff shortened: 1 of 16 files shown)


core/cloudflow-operator/src/test/kubernetes/shopping-cart-cr.json
```diff
@@ -0,0 +1,96 @@
+{
```
patriknw, line 1: I took this from the connected-cars example and modified it. Can be installed with kubectl cloudflow deploy src/test/kubernetes/shopping-cart-cr.json

blublinsky: For discussion as well. Currently Kubeflow operator is a bit asymmetric. It uses external Flink and Spark operators, but is managing Akka streamlets internally. It would be nice to factor out Akka management as a separate operator - similar to Spark and Flink and keep Cloudflow one as a manager - outsourcing management of specific runtimes to specific operators. This will make the overall maintenance of Cloudflow application simpler and will enable simpler integration with additional runtimes (when/if we will need it)

RayRoestenburg: We have a design for exactly this, the downside of it is obviously that you need to run more operators, requiring more resources.

RayRoestenburg: At the same time, having said that, this might be a very nice way to do it, integrating between akka microservices and cloudflow in a more flexible way.

blublinsky: Yea, I think this will a very nice separation of responsibilities and will make each operator simpler and more manageable

RayRoestenburg: See https://github.com/lightbend/cloudflow/issues/356 @blublinsky

blublinsky: That sounds like a great addition

patriknw:
this was just a learning experiment

PR #744, opened 2020-09-24
Respond with '200' on the root of gRPC routes

raboof:

For gRPC ingresses, serve an empty '200' on the root path.

GKE loadbalancers check for a '200' answer on the root for health checks by
default.

When using the PLAINTEXT_UPGRADE mode of grpc-java (i.e. with an insecure Nginx
ingress), the body must be empty for the 'upgrade' to work.

Manually on GKE


ennru (approved):
LGTM.

RayRoestenburg (approved):

Issue #538, opened 2020-06-24
Add a command to kubectl cloudflow to setup a TLS secured ingress for streamlets.


RayRoestenburg:
The cloudflow operator creates a k8s service for the streamlet. To access it from the outside, the user will have to setup an ingress. We'll have to check if this is made easier on GCP. Looks like you can use this: https://cloud.google.com/kubernetes-engine/docs/concepts/ingress

longmuir:
To make things easier, could we actually deploy an Ingress Controller as part of the marketplace installation? 

How would that compare to just telling people to just install it?
(open question - not insisting on one way or the other)

RayRoestenburg:
After discussion, we decided to use GKE-Ingress, and to add a feature to kubectl cloudflow to add ingresses for streamlets (TLS).

PR #745, opened 2020-09-24
Create endpoints with type NodePort

raboof:

Create endpoint service nodes with type NodePort instead of ClusterIP

To allow them to be exposed via gce loadbalancers as well

Not fully tested yet, first wanted to discuss whether this is a good idea in
theory before validating it in practice ;).


RayRoestenburg:
ClusterIP is only cluster internal, Nodeports are 'externally accessible', so this is not a great default, for all services. 
We've been thinking of adding a feature to kubectl cloudflow instead to add a nodeport service and an ingress, as well as TLS certs, specifically for when a user wants to make a streamlet externally accessible. See https://github.com/lightbend/cloudflow/issues/538 @raboof

raboof:
:+1:

Issue #586, opened 2020-07-16
Upgrade sbt-cloudflow and dependencies to Akka 2.6.8


RayRoestenburg:
Akka team announced Akka 2.6.8 which is a bugfix release for a regression introduced in Akka 2.6.7. It's always good to keep up with latest version of Akka.

raboof:
Done in #747

PR #747, opened 2020-09-24
Update Akka to 2.6.8

raboof:
This is also the Akka version Skuber 2.6.0 is using, so it should make updating that
smoother later.


raboof:
refs #586

PR #754, opened 2020-09-29
fixed misterious case in uppercase when running test in labels

franciscolopezsancho:
change the checking in the labels so it passes all the tests all the time. 

There was a bug when running the test. They were not passing from time to time, with no known cause. Were nondeterministic.

No

Running the test multiple times in a row. Before it would fail at least once every ten or so.

### DIFF
--- a/kubectl-cloudflow/config/config.go
+++ b/kubectl-cloudflow/config/config.go
@@ -335,23 +335,24 @@ func validateLabels(podConfig *configuration.Config, podName string) error {
 		}
 
 		for key, value := range labelsConfig.Root().GetObject().Items() {
-			label := strings.TrimSpace(key)
+			labelKey := strings.TrimSpace(key)
 			labelValue := strings.TrimSpace(value.String())
 
-			if labelHasPrefix(label) {
-				splitted := strings.Split(label, "/")
-				prefix := splitted[0]
-				label := splitted[1]
+			fmt.Printf("labelKey %s", labelKey)
 
-				if err := validateLabel(label, prefix); err != nil {
+			if labelHasPrefix(labelKey) {
+				splitted := strings.Split(labelKey, "/")
+				prefix := splitted[0]
+				suffix := splitted[1]
+
+				if err := validateLabelPrefix(labelKey, prefix); err != nil {
 					return err
 				}
-				return validateLabelValue(labelValue, label)
-			}
-			if err := validateLabel(label, ""); err != nil {
+				return validateLabelValue(labelValue, suffix)
+			} else if err := validateLabelValue(labelKey, ""); err != nil {
 				return err
 			}
-			return validateLabelValue(labelValue, label)
+			return validateLabelValue(labelValue, labelKey)
 		}
 	}
 	return nil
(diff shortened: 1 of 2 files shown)


franciscolopezsancho:
With the new test was failing all the time

kubectl-cloudflow/config/config.go
```diff
@@ -335,23 +335,24 @@ func validateLabels(podConfig *configuration.Config, podName string) error {
 		}
 
 		for key, value := range labelsConfig.Root().GetObject().Items() {
-			label := strings.TrimSpace(key)
+			labelKey := strings.TrimSpace(key)
 			labelValue := strings.TrimSpace(value.String())
 
-			if labelHasPrefix(label) {
-				splitted := strings.Split(label, "/")
+			fmt.Printf("labelKey %s", labelKey)
```
RayRoestenburg, line 341:
```suggestion
```

RayRoestenburg, line 341:
```suggestion
```

kubectl-cloudflow/config/config.go
```diff
@@ -335,23 +335,23 @@ func validateLabels(podConfig *configuration.Config, podName string) error {
...
 
-				if err := validateLabel(label, prefix); err != nil {
+				if err := validateLabelPrefix(labelKey, prefix); err != nil {
 					return err
 				}
-				return validateLabelValue(labelValue, label)
-			}
-			if err := validateLabel(label, ""); err != nil {
+				return validateLabelValue(labelValue, suffix)
+			} else if err := validateLabelValue(labelKey, ""); err != nil {
```
RayRoestenburg, line 351: I don;t get this part, why is it necessary to do validateLabelValue with a "" label, validating the labelKey? that is very confusing.

kubectl-cloudflow/config/config.go
```diff
@@ -361,34 +361,31 @@ func labelHasPrefix(label string) bool {
 	return strings.Count(label, "/") == 1 && !strings.HasPrefix(label, "/") && !strings.HasSuffix(label, "/")
 }
 
-func validateLabel(name string, prefix string) error {
-	// See https://github.com/kubernetes/kubernetes/issues/71140, IsDNS1123Subdomain and IsDNS1123Label do not allow uppercase letters.
-	labelPattern := regexp.MustCompile(`^[a-z0-9A-Z]{1}[a-z0-9\.\_\-]{0,61}[a-z0-9A-Z]{1}$`)
-
-	// TODO a DNS-1123 label must consist of lower case alphanumeric characters or '-', and must start and end with an alphanumeric character
+//Only the prefix of a value key has a different check
```
RayRoestenburg, line 364: I've read that prefixes can't have uppercase, is that what you mean?

RayRoestenburg (commented):
Let's quickly discuss this tomorrow morning, got some questions.

kubectl-cloudflow/config/config.go
franciscolopezsancho: Because there are two possible validation in the labelKey, keys that have prefix, and keys that don't have it. Makes sense?

kubectl-cloudflow/config/config.go
franciscolopezsancho: Oh, I don't think so. I was reading the code and didn't remember that. I will add a test for that otherwise may happen later on. Good catch.

kubectl-cloudflow/config/config.go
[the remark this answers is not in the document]
franciscolopezsancho: indeed

PR #765, opened 2020-10-05
Add methods to change Spark session and Flink streaming env from the streamlet.

RayRoestenburg:

Added updateSparkSession and updateStreamExecutionEnvironment, so the user can programmatically change respectively how the spark session and flink stream execution environment are set up.


blublinsky (approved):
LGTM

PR #773, opened 2020-10-08
Add MultiOutlet; base Splitter on it

ennru:

The current Akka Streamlet APIs make it impossible to achieve at-least-once processing with multiple outlets.

Users should not use multiple `committableSink`s in a streamlet as later elements might reach the outlet and committing before earlier elements when a stream diverges (eg. with `Partition`).

The existing `Splitter.sink`s may in rare cases backpressure for producing values to one outlet but still produce to the other outlet and commit the offset. If the streamlet/Pod is restarted before the earlier messages have been produced, they will not be reprocessed after the restart. 

To make correct committing when producing to multiple outlets easier, this PR introduces the `MultiOutlet.sink`. It accepts a collection per outlet and will commit first after all messages are produced to the outlets.

* Deprecate the `Spliter.sink` variant accepting `Sink`s
* Introduce `MultiOutlet.sink2` (and `sink3`) which produce to multiple outlets

Optional
* Introduce `flexiFlow` which produces to an outlet without terminating the stream. (The name stems from Alpakka Kafka, a different name might make more sense in Cloudflow.)

### DIFF
--- a/core/cloudflow-akka-util/src/main/scala/cloudflow/akkastream/util/scaladsl/MultiOutlet.scala
+++ b/core/cloudflow-akka-util/src/main/scala/cloudflow/akkastream/util/scaladsl/MultiOutlet.scala
@@ -1,0 +1,119 @@
+/*
+ * Copyright (C) 2016-2020 Lightbend Inc. <https://www.lightbend.com>
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package cloudflow.akkastream.util.scaladsl
+
+import akka.NotUsed
+import akka.kafka.CommitterSettings
+import akka.kafka.ConsumerMessage.Committable
+import akka.stream.scaladsl.{ Flow, GraphDSL, Sink, Unzip, ZipWith }
+import akka.stream.{ FlowShape, Graph }
+import cloudflow.akkastream.AkkaStreamletContext
+import cloudflow.streamlets.CodecOutlet
+
+import scala.collection.immutable
+
+object MultiOutlet {
+
+  /**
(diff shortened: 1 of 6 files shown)


core/cloudflow-akka-util/src/main/scala/cloudflow/akkastream/util/scaladsl/MultiOutlet.scala
```diff
@@ -0,0 +1,119 @@
...
+   */
+  def sink3[O1, O2, O3](
+      outlet1: CodecOutlet[O1],
+      outlet2: CodecOutlet[O2],
+      outlet3: CodecOutlet[O3],
+      committerSettings: CommitterSettings
+  )(implicit context: AkkaStreamletContext): Sink[((immutable.Seq[O1], immutable.Seq[O2], immutable.Seq[O3]), Committable), NotUsed] =
+    Flow
+      .fromGraph(graph3(context.flexiFlow(outlet1), context.flexiFlow(outlet2), context.flexiFlow(outlet3)))
+      .to(context.committableSink[Unit](committerSettings))
```
RayRoestenburg, line 56: nice reuse :-)

ennru:
Would we go as far as introducing 3-tuples for the Java DSL? Right now the Java DSL uses pairs of pairs when the user may pass three collections.

core/cloudflow-akka/src/main/scala/cloudflow/akkastream/AkkaStreamletContext.scala
```diff
@@ -61,6 +62,8 @@ trait AkkaStreamletContext extends StreamletContext {
   private[akkastream] def committableSink[T](outlet: CodecOutlet[T], committerSettings: CommitterSettings): Sink[(T, Committable), NotUsed]
   private[akkastream] def committableSink[T](committerSettings: CommitterSettings): Sink[(T, Committable), NotUsed]
 
+  private[akkastream] def flexiFlow[T](outlet: CodecOutlet[T]): Flow[(immutable.Seq[T], Committable), (Unit, Committable), NotUsed]
```
andreaTP, line 65: this is the very first API to take a `List`-like element, I wouldn't be shocked to use the `java.util.List` directly, to avoid unnecessary boxing/unboxing when it comes to `javadsl` and, also, for possible performance improvements, does it make sense?

andreaTP (commented):
Amazing work! Asked a couple of questions 😄 
Do you think is possible to create a unit test to guarantee that we are not going to broke the characteristics described in the docs? Or is it too much?

RayRoestenburg:
I think this can be closed, correct @ennru ?

Issue #767, opened 2020-10-06
Cloudflow operator is repeatedly trying to find and update secrets


RayRoestenburg:
The createOrUpdate does not use the right namespace to get the resource with `getOption`.
This causes resources that are not in the cloudflow namespace to not be found, and fail on create since they already exist.

RayRoestenburg:
Other calls like create and update to KubernetesClient use the namespace property on the resource. getOption just takes a name so cannot automatically get the resource from the right namespace. getInNamespace does not return Option.

RayRoestenburg:
Also realising that usingNamespace creates a new client. These new clients are not closed, which can cause a memory issue.

PR #774, opened 2020-10-08
Fix 'using namespace'

RayRoestenburg:
Fix for https://github.com/lightbend/cloudflow/issues/767

- [x] Tested on GKE


RayRoestenburg:
I initially thought this could be a memory issue, since usingNamespace creates a new KubernetesClient, but it actually shares the pool and other settings.

blublinsky (approved):
LGTM

PR #775, opened 2020-10-08
Fixed Flink filesystem initialization.

blublinsky:

Note, while testing. We tested with negative results, ensuring that execution fails correctly, when Azure account was not available. We also noticed, that due to the nature of the Flink operator, the observed results can be slightly deceiving. The job submission failed (as expected), but the cluster created for the job started and was running fine. The thing to remember here is that the fact of pods running correctly DOES NOT mean that the job is successfully deployed. You can use , for example

kubectl -n taxi-ride-fare describe FlinkApplication taxi-ride-fare-logger

to see the state of job submission

### DIFF
--- a/core/sbt-cloudflow/src/main/scala/cloudflow/sbt/CloudflowBasePlugin.scala
+++ b/core/sbt-cloudflow/src/main/scala/cloudflow/sbt/CloudflowBasePlugin.scala
@@ -38,7 +38,7 @@ object CloudflowBasePlugin extends AutoPlugin {
   final val DepJarsDir: String               = "dep-jars"
   final val OptAppDir                        = "/opt/cloudflow/"
   final val ScalaVersion                     = "2.12"
-  final val CloudflowVersion                 = BuildInfo.version
+  final val CloudflowVersion                 = "2.0.10"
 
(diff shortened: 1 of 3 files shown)


blublinsky:
It did fail on build, but this is because required base images with version 2.1.0-SNAPSHOT for Akka/Flink/Spark do not exist in the repo

core/sbt-cloudflow/src/main/scala/cloudflow/sbt/CloudflowBasePlugin.scala
```diff
@@ -38,7 +38,7 @@ object CloudflowBasePlugin extends AutoPlugin {
   final val DepJarsDir: String               = "dep-jars"
   final val OptAppDir                        = "/opt/cloudflow/"
   final val ScalaVersion                     = "2.12"
-  final val CloudflowVersion                 = "2.0.10"
+  final val CloudflowVersion                 = BuildInfo.version
```
RayRoestenburg, line 41:
```suggestion
  final val CloudflowVersion                 = "2.0.10"
```

RayRoestenburg: We'll fix this in a later PR about versions, which @andreaTP is working on.

RayRoestenburg (approved):

PR #777, opened 2020-10-09
[sbt-plugin] Remove the dependency on git and let user version

andreaTP:
Remove the dependency on git and remove the custom versioning plugin from the `sbt-plugin`

Let the user version their applications freely, we switched our examples to `sbt-dynver` but is not a requirement.

Yes, the users would need to manage versions by themselves.


core/sbt-cloudflow/src/sbt-test/sbt-cloudflow/version/helloworld1/src/main/blueprint/blueprint.conf
```diff
@@ -0,0 +1,6 @@
+blueprint {
+  streamlets {
+    hello-world = helloworld.HelloWorld1Shape
+  }
+  topics { }
```
RayRoestenburg, line 5:
```suggestion
```

core/sbt-cloudflow/src/sbt-test/sbt-cloudflow/version/helloworld2/src/main/blueprint/blueprint.conf
```diff
@@ -0,0 +1,6 @@
+blueprint {
+  streamlets {
+    hello-world = helloworld.HelloWorld2Shape
+  }
+  topics { }
```
RayRoestenburg, line 5: remove

andreaTP:
included in #784

PR #780, opened 2020-10-12
Multi out support as non-public API

ennru:
This is a minimal variant of #773 and does not offer multi-out flows or sinks to the public Akka Streamlet API but fixes the Splitter API to guarantee at-least-once.

* Introduce a `MultiData2` type which serves as a tuple of sequences. The idea is that such a type would make the Java API and the Scala API more uniform.

* Introduce `MultiProducer` (as internal API) which produces to two outlets in parallel and ensures proper committing after all elements are produced.

### DIFF
--- a/.github/workflows/labeler.yml
+++ b/.github/workflows/labeler.yml
@@ -1,12 +1,12 @@
 # https://github.com/actions/labeler
 name: "Pull Request Labeler"
 on:
-  - pull_request
+  - pull_request_target
 
 jobs:
   triage:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/labeler@master
+      - uses: actions/labeler@main
         with:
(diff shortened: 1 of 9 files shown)


.github/workflows/labeler.yml
```diff
@@ -1,12 +1,12 @@
...
 on:
-  - pull_request
+  - pull_request_target
 
 jobs:
   triage:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/labeler@master
+      - uses: actions/labeler@main
```
RayRoestenburg, line 10: We have no main (yet)?

RayRoestenburg: I think you mentioned it is not working great, should we remove, or does this fix things?

RayRoestenburg (approved):

Issue #782, opened 2020-10-12
Build cloudflow base images from upstream generic images


andreaTP:
**Is your feature request related to a problem? Please describe.**
Currently to build a Cloudflow application we need to have a specific base image uploaded matching version of Scala, Akka, Cloudflow itself, and more.
This is not strictly required e.g. the base image:
```
lightbend/akka-base:2.0.10-cloudflow-akka-2.6.9-scala-2.12
```
doesn't rely on any specific version of Akka, Scala and has very little to do with Cloudflow specifics as well:
https://hub.docker.com/layers/lightbend/akka-base/2.0.10-cloudflow-akka-2.6.9-scala-2.12/images/sha256-5f418d005be362cf847db272b66a7a83c2e7032c54ec36099d2a0b5671e35363?context=explore

**Is your feature request related to a specific runtime of cloudflow or applicable for all runtimes?**
All runtimes.

**Describe the solution you'd like**
The build of Cloudflow applications can be self-contained in the Sbt plugin depending on upstream base images.

RayRoestenburg:
@andreaTP I think this can be closed.

andreaTP:
yes! 🎉

PR #783, opened 2020-10-12
[sbt-plugin] Detach Cloudflow version from the version of the sbt plugin

andreaTP:
Here we detach the version of the `sbt-plugin` from the version of `cloudflow` used by the application.

This is a rough mechanism at the moment and it would require fixing #782 to give the user a nice experience.

This enables to decouple the sbt-plugin from the runtime.
I.e. a user becomes able to use a more recent version of the `sbt-plugin` (that, for example, fixes a bug) to target the cloudflow runtime that is already deployed.

Yes, it introduces an additional setting in the `sbt-plugin`

`scripted` tests are checking the result of this change


core/sbt-cloudflow/src/main/scala/cloudflow/sbt/CloudflowKeys.scala
```diff
@@ -44,6 +44,7 @@ trait CloudflowSettingKeys {
   val cloudflowFlinkBaseImage    = settingKey[Option[String]]("The base image for Cloudflow Flink plugin")
   val cloudflowSparkBaseImage    = settingKey[Option[String]]("The base image for Cloudflow Spark plugin")
   val cloudflowAkkaBaseImage     = settingKey[Option[String]]("The base image for Cloudflow Akka plugin")
+  val cloudflowVersion           = settingKey[String]("The default version of Cloudflow to be used")
```
RayRoestenburg, line 47: Add some discouraging remarks for users :-)

andreaTP:
included in #784

PR #792, opened 2020-10-14
Merge default named Kafka cluster config as fallback to any global topic config

### LINKED ISSUE #790: leaving out bootstrap.servers but adding topics in conf file does not work.
With config file:
```
cloudflow.topics {
  rides {
    topic.name = "my-rides"
    partitions = 10
    replicas = 2
  }
  fares {
    topic.name = "my-fares"
    partitions = 11
    replicas = 2
  }
  taxifares {
    topic.name = "my-tfares"
    partitions = 15
    replicas = 1
  }
}
```

You will get the error message:

```
  |Runtime Kafka bootstrap.servers is not set for topic ${topic.id}
  |A 'default' Kafka cluster, named cluster, or inline cluster configuration was not provided for this topic.
  |To set a 'default' Kafka cluster you must update or reinstall the cloudflow-operator Helm release with value `kafkaClusters.default.bootstrapServers`.
  |""".stripMargin)
```

But with: 

```
cloudflow.topics {
  rides {
    bootstrap.servers = "cloudflow-strimzi-kafka-bootstrap.cloudflow:9092"
    topic.name = "my-rides"
    partitions = 10
    replicas = 2
  }
  fares {
    bootstrap.servers = "cloudflow-strimzi-kafka-bootstrap.cloudflow:9092"
    topic.name = "my-fares"
    partitions = 11
    replicas = 2
  }
  taxifares {
    bootstrap.servers = "cloudflow-strimzi-kafka-bootstrap.cloudflow:9092"
    topic.name = "my-tfares"
    partitions = 15
    replicas = 1
  }
}
```
Everything is ok.
it should also work without specifying bootstrap servers, and in this case it should use the default.

seglo:
Fixes #790 

Bonus: link to storage pre-reqs in admin docs for Spark and Flink support

### DIFF
--- a/core/cloudflow-operator/src/main/scala/cloudflow/operator/event/ConfigurationScopeLayering.scala
+++ b/core/cloudflow-operator/src/main/scala/cloudflow/operator/event/ConfigurationScopeLayering.scala
@@ -132,26 +132,27 @@ object ConfigurationScopeLayering {
                                                     streamletConfig: Config,
                                                     appConfig: Config,
                                                     clusterSecretConfigs: Map[String, Config]): Config = {
+    val defaultClusterConfig = clusterSecretConfigs.get(TopicActions.DefaultConfigurationName)
     val portMappingConfigs = deployment.portMappings.flatMap {
       case (port, topic) =>
         Try {
-          val portMappingConfig = if (appConfig.hasPath(s"$TopicsConfigPath.${topic.id}")) {
-            appConfig
-              .getConfig(s"$TopicsConfigPath.${topic.id}")
-              .withFallback(topic.config)
-          } else {
-            val clusterSecretConfig =
-              topic.cluster
-                .flatMap(clusterName => clusterSecretConfigs.get(clusterName))
-                .orElse(clusterSecretConfigs.get(TopicActions.DefaultConfigurationName))
-                .getOrElse(ConfigFactory.empty())
+          val clusterSecretConfig =
+            topic.cluster
+              .flatMap(clusterName => clusterSecretConfigs.get(clusterName))
+              .orElse(defaultClusterConfig)
+              .getOrElse(ConfigFactory.empty())
 
-            appConfig
-              .withFallback(topic.config)
-              .withFallback(clusterSecretConfig)
-          }
+          val portMappingConfig =
+            if (appConfig.hasPath(s"$TopicsConfigPath.${topic.id}"))
+              appConfig.getConfig(s"$TopicsConfigPath.${topic.id}")
+            else
+              appConfig
 
-          portMappingConfig
+          val portMappingWithFallbackConfig = portMappingConfig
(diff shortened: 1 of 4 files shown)


core/cloudflow-operator/src/main/scala/cloudflow/operator/event/ConfigurationScopeLayering.scala
```diff
@@ -132,26 +132,27 @@ object ConfigurationScopeLayering {
                                                     streamletConfig: Config,
                                                     appConfig: Config,
                                                     clusterSecretConfigs: Map[String, Config]): Config = {
+    val defaultClusterConfig = clusterSecretConfigs.get(TopicActions.DefaultConfigurationName)
```
andreaTP, line 135: Can we have a test, at least checking the `Config` in output contains the expected keys/values?

andreaTP (commented):
Asking for an example to understand what has been missing/wrong with the old way of building up the configuration (and to avoid breaking it in the future).
Do you think it's doable?

core/cloudflow-operator/src/main/scala/cloudflow/operator/event/ConfigurationScopeLayering.scala
seglo: I was having a git issue and forgot to include this. I amended the test in `ConfigurationScopeLayeringSpec` to include the new use case.

seglo:
The problem was that a topic configuration in `cloudflow.topics.{topic}` was assumed to include a `bootstrap.servers` when it should fallback to the `default` or a named cluster config.  The `ConfigurationScopeLayeringSpec` is supposed to catch regressions here, but I missed the use case in my initial impl.  The test has been updated.

andreaTP:
> The problem was that a topic configuration in `cloudflow.topics.{topic}` was assumed to include a `bootstrap.servers` when it should fallback to the `default` or a named cluster config. The `ConfigurationScopeLayeringSpec` is supposed to catch regressions here, but I missed the use case in my initial impl. The test has been updated.

Amazing finding! 😄  Is there an easy way to unit test the behavior?

andreaTP (approved):

core/cloudflow-operator/src/main/scala/cloudflow/operator/event/ConfigurationScopeLayering.scala
andreaTP: Nice! Thanks!

andreaTP (approved):

andreaTP:
I apologize again for the confusion @seglo , can you please rebase on top of:
https://github.com/lightbend/cloudflow/pull/793
to check that the CI job `build-and-test-scala` works properly for "external contributors"😜  like you (at least in this case XD)

andreaTP:
merged in #795 , thanks!

PR #793, opened 2020-10-14
fix the CI for external contributors

andreaTP:
User's repositories are not appropriately tagged, and this results in missing jars.


RayRoestenburg (approved):

andreaTP:
not the root cause

PR #798, opened 2020-10-15
Automate local it tests

andreaTP:
In an ideal world we should be able to run:
```
make version=<version>
```
and the it-tests are executed 😄  this is a step in that direction.

### DIFF
--- a/integration-tests/Makefile
+++ b/integration-tests/Makefile
@@ -5,6 +5,9 @@
 .PHONY: all
 all: | spawn-gke-cluster prepare-swiss-knife prepare-itest-osx prepare-cluster run-it-tests delete-gke-cluster
 
+.PHONY: on-cluster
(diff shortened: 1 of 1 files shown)


integration-tests/Makefile
```diff
@@ -32,7 +38,7 @@ prepare-cluster:
 	helm repo add strimzi https://strimzi.io/charts/ | true
 	helm repo update
 	helm upgrade -i strimzi strimzi/strimzi-kafka-operator --namespace cloudflow
-	kubectl apply -f kafka-cluster.yaml
+	(for i in 1 2 3; do kubectl apply -f kafka-cluster.yaml && break || sleep 2; done)
```
RayRoestenburg, line 41: 👍

RayRoestenburg (approved):

PR #802, opened 2020-10-16
Auto docs generation

andreaTP:
This is preparation work to automate the release of the docs.

The versions are propagated automatically to all the docs by generating the needed `antora.yml` files.


RayRoestenburg (commented):
You can remove the antora.yaml files and generate these (as you already do in the Makefile), as discussed. Great fixes!

RayRoestenburg (approved):

PR #812, opened 2020-10-21
Edit Apache Flink entry script to prevent task manager memory setting overriding

### LINKED ISSUE #805: Apache Flink memory settings overrided in flink-entrypoint.sh
**Describe the bug**

There are custom Apache Flink memory properties that can be defined using `FlinkApplicationSpec` at Lyft operator (see the [documentation](https://github.com/lyft/flinkk8soperator/blob/master/docs/crd.md#flink-application-custom-resource-definition)).

Apache Flink uses quite complex algorithm to calculate a dozens of memory parameters. AFAIK current version of Lyft doesn't support Apache Flink memory configuration higher than 1.9 (see the [documentation](https://ci.apache.org/projects/flink/flink-docs-stable/ops/memory/mem_migration.html#changes-in-configuration-options)).

All the aforementioned lead us to the necessity `taskmanager.heap.size` parameter value control since it pretends to be an upper limit for the major part of the rest memory parameters (also see the [Flink memory calculator](https://docs.google.com/spreadsheets/d/1mJaMkMPfDJJ-w6nMXALYmTc4XxiV30P5U7DzgwLkSoE/edit?usp=sharing) found at SO).

Cloudflow `flink-entrypoint.sh` at line 48 ([here](https://github.com/lightbend/cloudflow/blob/84d39dae097074038e959bc880b0aaa25544ea7a/external/multi-base-images/flink/flink-entrypoint.sh#L48)) seems to put paid to any efforts considering Apache Flink task manager memory management since it overrides all parameters with the hard-coded value of 1024 mb.

We've changed the `flink-entrypoint.sh` and moved the line 48 closer to the beginning of the script so it stopped overriding other values and after that Apache Flink session clusters have started to work as expected.

It was a very odd decision to use hardcoded "the-all-overriding" memory parameters in the entry script. We don't have a clue about the reasons for such solution.

Please also see a quick drawn drill-down schema which illustrates the Lyft operator behaviour considering the memory size calculation based on K8s requests.

**To Reproduce**
Try to increase request limits at Flink operator higher than 2 Gb with default memory fraction with the parallelism value set to '2' and task slots set to '2'.

**Expected behavior**
The total Task Manager memory is two times less than `taskmanager.heap.size` calculated at Lyft operator.

Lockdain:
The change Fixes #805 so now Apache Flink is able to use calculated memory values for Task Manager pods and they will not be overrided by the hard-coded value of 1024 mb. 

Before the fix all Apache Flink deployments were limited to use only 1024 Mb of `taskmanager.memory.flink.size` which is not enough for the majority of use cases.

No.

The change was tested on a local OpenShift cluster.

### DIFF
--- a/external/multi-base-images/flink/flink-entrypoint.sh
+++ b/external/multi-base-images/flink/flink-entrypoint.sh
@@ -32,6 +32,8 @@ drop_privs_cmd() {
 }
 
 # Add in extra configs set by the operator
+echo "taskmanager.memory.flink.size: 1024mb" >> "$FLINK_HOME/conf/flink-conf.yaml"
+
 if [ -n "$FLINK_PROPERTIES" ]; then
     echo "$FLINK_PROPERTIES" >> $FLINK_HOME/flink-conf-tmp.yaml
 fi
@@ -45,7 +47,6 @@ fi
 echo "web.upload.dir: $FLINK_HOME" >> "$FLINK_HOME/conf/flink-conf.yaml"
 echo "jobmanager.web.upload.dir: $FLINK_HOME" >> "$FLINK_HOME/conf/flink-conf.yaml"
 
-echo "taskmanager.memory.flink.size: 1024mb" >> "$FLINK_HOME/conf/flink-conf.yaml"
 
 # Add JMX metric reporter to config
 echo "metrics.reporters: jmx" >> "$FLINK_HOME/conf/flink-conf.yaml"


lightbend-cla-validator:
At least one pull request committer is not linked to a user. See https://help.github.com/en/articles/why-are-my-commits-linked-to-the-wrong-user#commits-are-not-linked-to-any-user

Lockdain:
@RayRoestenburg @yuchaoran2011

RayRoestenburg:
Please link the commit to your user as the CLA validator suggests.

Lockdain:
Hi @RayRoestenburg 
Already done, the commit is linked:

image

RayRoestenburg:
@Lockdain thanks must have missed that on my phone..

RayRoestenburg (approved):

PR #830, opened 2020-10-26
Reporting error status in kubectl cloudflow status for unrecoverable errors.

RayRoestenburg:

An unrecoverable error in the application results in an error status in the CR. Once an application is in an error state it can only be undeployed, after which a user can decide to deploy again. (if the operator would update the status after /during an error status, the error status would disappear/be overwritten because of some pod change)

Particular errors would cause the cloudflow-operator to crash loop, since it would continue to try to deploy applications that would cause an exception.

`kubectl cloudflow status` can now show an Error state, as well as an Error message, explaining the reason for the Error state.

GKE, forcing an error state in a application.

### DIFF
--- a/core/cloudflow-blueprint/src/main/scala/cloudflow/blueprint/Topic.scala
+++ b/core/cloudflow-blueprint/src/main/scala/cloudflow/blueprint/Topic.scala
@@ -20,7 +20,7 @@ import com.typesafe.config._
 object Topic {
   val LegalTopicChars   = "[a-zA-Z0-9\\._\\-]"
   val LegalTopicPattern = s"$LegalTopicChars+".r
-  val MaxLength         = 255
+  val MaxLength         = 249
 }
 
 /**
--- a/core/cloudflow-operator/src/main/scala/cloudflow/operator/CloudflowApplication.scala
+++ b/core/cloudflow-operator/src/main/scala/cloudflow/operator/CloudflowApplication.scala
@@ -234,15 +234,19 @@ object CloudflowApplication {
       )
     }
 
+    def ignoreOnErrorStatus(oldApp: Option[CloudflowApplication.CR], newApp: CloudflowApplication.CR) = {
+      val _ = newApp
+      oldApp.flatMap(_.status) match {
+        case Some(status) if status.appStatus == Some(Status.Error) => false
+        case _                                                      => true
+      }
+    }
+
     def toAction(app: CloudflowApplication.CR): Action[ObjectResource] =
       Action.updateStatus(
         app.withStatus(this),
         editor,
-        (oldApp: Option[CloudflowApplication.CR], _: CloudflowApplication.CR) =>
-          oldApp.flatMap(_.status) match {
-            case Some(status) if status.appStatus == Some(Status.Error) => false
-            case _                                                      => true
-          }
+        ignoreOnErrorStatus
       )
   }
 
--- a/core/cloudflow-operator/src/main/scala/cloudflow/operator/CloudflowApplication.scala
+++ b/core/cloudflow-operator/src/main/scala/cloudflow/operator/CloudflowApplication.scala
@@ -22,6 +22,7 @@ import scala.collection.immutable._
 import scala.util.Try
 import com.typesafe.config._
 
+import org.slf4j.LoggerFactory
 import play.api.libs.json._
 import play.api.libs.json.JsonNaming.SnakeCase
 
@@ -129,6 +130,7 @@ object CloudflowApplication {
     val Pending          = "Pending"
     val CrashLoopBackOff = "CrashLoopBackOff"
     val Error            = "Error"
+    private val log      = LoggerFactory.getLogger(Status.getClass)
 
     def apply(
         spec: CloudflowApplication.Spec
@@ -142,13 +144,15 @@ object CloudflowApplication {
       )
     }
 
-    def errorAction(app: CloudflowApplication.CR, msg: String): Action[ObjectResource] =
+    def errorAction(app: CloudflowApplication.CR, msg: String): Action[ObjectResource] = {
+      log.info(s"Setting error status for app ${app.spec.appId}")
       Status(app.spec)
         .copy(
           appStatus = Some(CloudflowApplication.Status.Error),
           appMessage = Some(msg)
         )
         .toAction(app)
+    }
 
     def createStreamletStatuses(spec: CloudflowApplication.Spec) = {
       // TODO not match on runtime, this is not great for extensibility.
@@ -189,7 +193,6 @@ object CloudflowApplication {
                     appMessage: Option[String] = None) {
     def aggregatedStatus = appStatus.getOrElse(Status.Unknown)
     def updateApp(newApp: CloudflowApplication.CR) = {
-      // copy PodStatus lists that already exist
       val newStreamletStatuses = Status.createStreamletStatuses(newApp.spec).map { newStreamletStatus =>
         streamletStatuses
           .find(_.streamletName == newStreamletStatus.streamletName)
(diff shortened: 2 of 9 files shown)


core/cloudflow-blueprint/src/main/scala/cloudflow/blueprint/Topic.scala
```diff
@@ -20,7 +20,7 @@ import com.typesafe.config._
 object Topic {
   val LegalTopicChars   = "[a-zA-Z0-9\\._\\-]"
   val LegalTopicPattern = s"$LegalTopicChars+".r
-  val MaxLength         = 255
+  val MaxLength         = 249
```
RayRoestenburg, line 23: The max number of chars has changed in recent version of Kafka.

core/cloudflow-operator/src/main/scala/cloudflow/operator/CloudflowApplication.scala
```diff
@@ -233,7 +235,15 @@ object CloudflowApplication {
...
     def toAction(app: CloudflowApplication.CR): Action[ObjectResource] =
-      Action.updateStatus(app.withStatus(this), editor)
+      Action.updateStatus(
+        app.withStatus(this),
+        editor,
+        (oldApp: Option[CloudflowApplication.CR], _: CloudflowApplication.CR) =>
+          oldApp.flatMap(_.status) match {
+            case Some(status) if status.appStatus == Some(Status.Error) => false
+            case _                                                      => true
+          }
```
ennru, line 245: This is very dense. Maybe find a good place and name for it?
`ignoreOnErrorStatus`, `requireOkStatus` ?

core/cloudflow-operator/src/main/scala/cloudflow/operator/action/Action.scala
```diff
@@ -306,7 +309,13 @@ class UpdateStatusAction[T <: ObjectResource](
           editor.updateMetadata(resource, resource.metadata.copy(resourceVersion = existingResource.metadata.resourceVersion))
         )
       res ← resourceVersionUpdated
-        .map(resourceToUpdate => recoverFromConflict(client.updateStatus(resourceToUpdate), client, retries - 1, executeUpdateStatus))
+        .map { resourceToUpdate =>
+          if (predicateForUpdate(existing, resourceToUpdate)) {
+            recoverFromConflict(client.updateStatus(resourceToUpdate), client, retries - 1, executeUpdateStatus)
+          } else {
+            Future.successful(resource)
```
ennru, line 316: Possibly log that this update action was ignored.

core/cloudflow-operator/src/main/scala/cloudflow/operator/action/TopicActions.scala
```diff
@@ -123,71 +123,90 @@ object TopicActions {
...
-        super.execute(client).flatMap { resourceCreatedAction =>
-          createTopic().map(_ => resourceCreatedAction)
-        }
+  def createAction(newApp: CloudflowApplication.CR,
+                   appNamespace: String,
+                   labels: CloudflowLabels,
+                   topic: TopicInfo): Action[ObjectResource] =
+    (topic.bootstrapServers, topic.partitions, topic.replicationFactor) match {
+      case (Some(bootstrapServers), Some(partitions), Some(replicas)) =>
+        val brokerConfig = topic.brokerConfig
```
ennru, line 163: Possibly move the good path into a method to make the `match` span over less code.

ennru (approved):
LGTM.

core/cloudflow-operator/src/main/scala/cloudflow/operator/CloudflowApplication.scala
andreaTP: The logic here is already pretty self-contained, maybe I'm missing the reason why this has been implemented as an additional argument and not directly in the `UpdateStatusAction` (e.g. [here](https://github.com/lightbend/cloudflow/pull/830/files#diff-2401f079f994e5a152de63295fa322693b201133df858e2d361ad912b765cd5bR313) )

I think this will automatically make the code less dense as well 😄

core/cloudflow-operator/src/main/scala/cloudflow/operator/Name.scala
```diff
@@ -58,15 +64,27 @@ object Name {
...
   /**
-   * Make a name compatible with DNS 1123 standard: like a full domain name.
+   * Makes a name compatible with DNS 1123 standard: like a full domain name, max 63 characters
    * Regex to follow: [a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*
-   * Limit the resulting name to 63 characters
+   * Limits the resulting name to 63 characters
+   * https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#dns-label-names
    */
-  private[operator] def makeDNS1123Compatible(name: String): String = {
+  private[operator] def makeDNS1123CompatibleLabelName(name: String): String = {
```
andreaTP, line 72: This code is exclusively used in the tests now, can we remove this method and test the new `makeDNS1123CompatibleSubDomainName` instead?

core/cloudflow-operator/src/main/scala/cloudflow/operator/event/StatusChangeEvent.scala
```diff
@@ -113,6 +113,8 @@ object StatusChangeEvent extends Event {
                 }
             }
             currentStatuses.get(appId).map(_.toAction(app)).toList
+          // app is in error state, no need to handle updates.
+          case (Some(_), _) => List()
```
andreaTP, line 117: Instead of leaving the reasoning it in a comment, we can improve by having it in code 🙂 .
You can move this condition above and flip the check:
```scala
case (Some(app), _) if app.status.flatMap(_.appStatus) == Some(CloudflowApplication.Status.Error) ⇒ List()
```

andreaTP (approved):
A few minor comments but looks solid work!

core/cloudflow-operator/src/main/scala/cloudflow/operator/event/StatusChangeEvent.scala
RayRoestenburg: Sure

core/cloudflow-operator/src/main/scala/cloudflow/operator/Name.scala
RayRoestenburg: done.

core/cloudflow-operator/src/main/scala/cloudflow/operator/action/TopicActions.scala
RayRoestenburg: done

core/cloudflow-operator/src/main/scala/cloudflow/operator/action/Action.scala
RayRoestenburg: done

core/cloudflow-operator/src/main/scala/cloudflow/operator/CloudflowApplication.scala
RayRoestenburg: `UpdateStatusAction` is for any `T <: ObjectResource`, not only for `CloudflowApplication`, which is why it is not implemented there.

RayRoestenburg: I'm not totally happy with the results of this so I might keep it a bit ugly :-)

RayRoestenburg: I refactored it into `ignoreOnErrorStatus`.

RayRoestenburg:
The solution here is not complete, when an application gets in an error state it _has to_ be undeployed. normal subsequent deployments will not clear the error state that is shown in `kubectl cloudflow status`. This will need more work, but in the meantime this is already an improvement. /cc @andreaTP

PR #831, opened 2020-10-26
Force incompatible change, users must update kubectl cloudflow and op…

RayRoestenburg:
…erator.

Since there were some issues in upgrade to 2.0.11, in the next version it will be required to update sbt-cloudflow, kubectl cloudflow and the cloudflow operator to the next version.

Yes, the user has to update the cloudflow-operator, rebuild and deploy apps.


andreaTP (approved):
LGTM!
As an improvement, we can think about injecting those values at compile-time (with sbt in Scala and something like [this](https://goenning.net/2017/01/25/adding-custom-data-go-binaries-compile-time/) in go) and keep the source of truth in a single place. No need to tackle this now though 👍

Issue #638, opened 2020-08-17
Expose Control Materialized value on sources that read from inlets, instead of using NotUsed.


RayRoestenburg:
See https://github.com/lightbend/cloudflow/pull/418#discussion_r471453147

PR #846, opened 2020-11-04
Expose the Kafka Consumer Control and improve shutdown behaviour

ennru:

References #638 

This improves the shutdown behaviour of Akka Streamlets as a first step to a controlled shutdown as descirbed in https://doc.akka.io/docs/alpakka-kafka/current/consumer.html#controlled-shutdown

The APIs change as they now expose the materialized value of the Alpakka Kafka sources. This is not binary-compatible.

### DIFF
--- a/core/cloudflow-akka/src/main/scala/cloudflow/akkastream/AkkaStreamletContextImpl.scala
+++ b/core/cloudflow-akka/src/main/scala/cloudflow/akkastream/AkkaStreamletContextImpl.scala
@@ -55,7 +55,19 @@ final class AkkaStreamletContextImpl(
   private val completionPromise = Promise[Dun]()
   private val completionFuture  = completionPromise.future
 
-  val killSwitch = KillSwitches.shared(streamletRef)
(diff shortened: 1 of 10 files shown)


core/cloudflow-akka/src/main/scala/cloudflow/akkastream/AkkaStreamletContextImpl.scala
```diff
@@ -362,18 +365,30 @@ final class AkkaStreamletContextImpl(
...
+            f().recover {
+              case cause ⇒
+                system.log.error(cause, "onStop callback failed.")
+                Dun
+            }
           }
-        }
-      )
+        )
+      }
```
ennru, line 391: We can do even better here and trigger `control.shutdown()` after a short while.
Alpakka Kafka uses a stop-timeout of 30s which is a bit too long for most users.

ennru:
Let's keep the APIs unchanged for now and focus on #847.

PR #847, opened 2020-11-04
Controlled shutdown of Kafka consumers

ennru:

This implements more controlled Consumer shutdown as described in https://doc.akka.io/docs/alpakka-kafka/current/consumer.html#controlled-shutdown

This lowers the risk of messages being half processed during streamlet shutdown.
It improves chances that commits reach the Kafka consumer during shutdown.

A new setting "consumer stop timeout" needs to be introduced.
The underlying `akka.kafka.consumer.stop-timeout` can be overwritten to 0s for streamlets.

### DIFF
--- a/core/cloudflow-akka-tests/src/test/resources/logback-test.xml
+++ b/core/cloudflow-akka-tests/src/test/resources/logback-test.xml
@@ -1,12 +1,25 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <configuration>
-  <logger name="akka" level="${LOGLEVEL_AKKA:-OFF}" />
+  <logger name="akka" level="DEBUG"/>
+  <logger name="akka.actor.TimerScheduler" level="INFO"/>
+  <logger name="akka.kafka" level="DEBUG"/>
+
+  <logger name="org.apache.zookeeper" level="WARN"/>
+  <logger name="org.I0Itec.zkclient" level="WARN"/>
+
+  <logger name="kafka" level="WARN"/>
+  <logger name="org.apache.kafka" level="WARN"/>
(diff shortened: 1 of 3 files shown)


ennru:
This is a non-API-changing alternative to #846.

RayRoestenburg:
Related to #849 #848

core/cloudflow-akka-tests/src/test/resources/logback-test.xml
```diff
@@ -1,12 +1,25 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <configuration>
-  <logger name="akka" level="${LOGLEVEL_AKKA:-OFF}" />
```
RayRoestenburg, line 3: I guess these changes were for debugging and will not be committed?

core/cloudflow-akka/src/main/scala/cloudflow/akkastream/AkkaStreamletContextImpl.scala
```diff
@@ -362,18 +381,49 @@ final class AkkaStreamletContextImpl(
     // we created this file when the pod started running (see AkkaStreamlet#run)
     Files.deleteIfExists(Paths.get(s"/tmp/$streamletRef.txt"))
 
-    killSwitch.shutdown()
     import system.dispatcher
+    val controls = KafkaControls.get
```
RayRoestenburg, line 385: `stop` is currently only called on AkkaStreamlet catching exception in run. This logic should also happen if it completes without error.

RayRoestenburg (commented):
Needs to also do the graceful shutdown when the streamlet stops without error. (not sure if you want to do that in separate PR)

ennru:
I'll re-do this on top of #853

PR #858, opened 2020-11-11
added fix to avoid confusion when /mnt/[runtime]/storage is already present in the conf but is not in mount-path

franciscolopezsancho:
allow mounting PVC in presence on /mnt/[runtime]/storage in the config that has nothing to do with mount-path

fix. Otherwise PVC weren't mounted as expected. 

no

unit tests

### DIFF
--- a/kubectl-cloudflow/config/config_test.go
+++ b/kubectl-cloudflow/config/config_test.go
@@ -349,11 +349,52 @@ func Test_mountingExistingClaim(t *testing.T) {
 			}
 		}
 	}`
-	expected = strings.Replace(expected, "\t", "", -1)
-	expected = strings.Replace(expected, "\n", "", -1)
-	config = strings.Replace(config, "\t", "", -1)
-	config = strings.Replace(config, "\n", "", -1)
-	assert.Equal(t, expected, config)
+	assert.Equal(t, trim(expected), trim(config))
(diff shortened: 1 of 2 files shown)


kubectl-cloudflow/config/config_test.go
```diff
@@ -349,11 +349,52 @@ func Test_mountingExistingClaim(t *testing.T) {
...
 	}`
-	expected = strings.Replace(expected, "\t", "", -1)
-	expected = strings.Replace(expected, "\n", "", -1)
-	config = strings.Replace(config, "\t", "", -1)
-	config = strings.Replace(config, "\n", "", -1)
-	assert.Equal(t, expected, config)
+	assert.Equal(t, trim(expected), trim(config))
+}
+
+//checks it adds it even there is some other /mnt/flink/storage
```
RayRoestenburg, line 355: Weird sentence?

RayRoestenburg (approved):

kubectl-cloudflow/config/config_test.go
franciscolopezsancho: indeed

RayRoestenburg: So you can you fix it? 🤣

PR #859, opened 2020-11-12
Execute status error action on action failure.

RayRoestenburg:

Ensuring that an action failure marks the app in error state.

More robust.

No.

- [ ] GKE

### DIFF
--- a/core/cloudflow-operator-actions/src/main/scala/cloudflow/operator/action/Action.scala
+++ b/core/cloudflow-operator-actions/src/main/scala/cloudflow/operator/action/Action.scala
@@ -15,6 +15,9 @@
  */
 
 package cloudflow.operator.action
+
+import scala.concurrent._
+import scala.concurrent.duration._
 import akka.actor.ActorSystem
 import akka.pattern._
 import org.slf4j.LoggerFactory
@@ -22,8 +25,7 @@ import play.api.libs.json._
 import skuber._
 import skuber.api.client._
 import skuber.api.patch.Patch
-import scala.concurrent._
-import scala.concurrent.duration._
+import cloudflow.operator.CloudflowApplication
 
 /**
  * Captures an action to create, delete or update a Kubernetes resource.
@@ -36,9 +38,9 @@ sealed trait Action {
   def name: String
 
   /**
-   * The application this action is executing for.
+   * The app this action is executed for.
    */
-  def appId: Option[String] = None
+  def app: CloudflowApplication.CR
 
   /*
    * The namespace that the action takes place in.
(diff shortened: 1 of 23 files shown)


core/cloudflow-operator-actions/src/main/scala/cloudflow/operator/action/Action.scala
```diff
@@ -427,6 +449,7 @@ final class ProvidedAction[T <: ObjectResource, R <: ObjectResource](
 final class ProvidedByLabelAction[T <: ObjectResource, R <: ObjectResource](
     val labelKey: String,
     val labelValues: Vector[String],
+    val app: CloudflowApplication.CR,
```
andreaTP, line 452: I do see some overlapping concepts here, e.g. I "think" that the Application.CR already contains the Namespace, isn't it?

core/cloudflow-operator-actions/src/main/scala/cloudflow/operator/action/EndpointActions.scala
```diff
@@ -50,12 +50,17 @@ object EndpointActions {
 
     val deleteActions = (currentEndpoints -- newEndpoints).flatMap { endpoint ⇒
       Seq(
-        Action.delete[Service](Name.ofService(StreamletDeployment.name(newApp.spec.appId, endpoint.streamlet)), namespace)
+        Action.delete[Service](Name.ofService(StreamletDeployment.name(newApp.spec.appId, endpoint.streamlet)), newApp, namespace)
```
andreaTP, line 53: I think we can simplify here:
```scala
Action.delete[Service](Name.ofService(StreamletDeployment.name(newApp.spec.appId, endpoint.streamlet)), newApp, namespace)
```

might become something similar to:

```scala
Action.delete[Service](newApp, endpoint.streamlet)
```

core/cloudflow-operator-actions/src/main/scala/cloudflow/operator/action/EndpointActions.scala
```diff
@@ -50,12 +50,17 @@ object EndpointActions {
...
       )
     }.toList
     val createActions = (newEndpoints -- currentEndpoints).flatMap { endpoint ⇒
       Seq(
-        createServiceAction(endpoint, StreamletDeployment.name(newApp.spec.appId, endpoint.streamlet), namespace, labels, ownerReferences)
+        createServiceAction(endpoint,
+                            newApp,
```
andreaTP, line 59: as well if we pass around the entire Cr this can be simpler maybe

core/cloudflow-operator/src/main/scala/cloudflow/operator/Operator.scala
```diff
@@ -98,6 +99,7 @@ object Operator {
 
   def handleConfigurationInput(
       client: KubernetesClient,
+      runners: Map[String, Runner[_]],
```
andreaTP, line 102: I don't understand, from the current diff, what we are getting by passing around the runners, am I missing some context?

andreaTP (changes requested):
I see the direction that you are taking here, it's a tradeoff to pass around the entire CR or just use the needed pieces and bits.
If we decide to have the entire CR around we can probably simplify other parts of the code if I understand it correctly. 😄

core/cloudflow-operator-actions/src/main/scala/cloudflow/operator/action/Action.scala
RayRoestenburg: yes

core/cloudflow-operator/src/main/scala/cloudflow/operator/Operator.scala
RayRoestenburg: yes, errorAction

core/cloudflow-operator-actions/src/main/scala/cloudflow/operator/action/Action.scala
RayRoestenburg: you might do something in another namespace for the app.

core/cloudflow-operator-actions/src/main/scala/cloudflow/operator/action/EndpointActions.scala
RayRoestenburg: you might delete in a different namespace.

core/cloudflow-operator-actions/src/main/scala/cloudflow/operator/action/EndpointActions.scala
RayRoestenburg: In this case we can cleanup a little bit, yes.

core/cloudflow-operator-actions/src/main/scala/cloudflow/operator/action/EndpointActions.scala
RayRoestenburg: Can add another method, let me see (default to app namespace is probably not bad).

RayRoestenburg: done.

RayRoestenburg:
Cleaned up so that the app namespace is used.

Issue #848, opened 2020-11-05
Graceful shutdown for Akka Streamlets


orElse:
**Is your feature request related to a problem? Please describe.**
Currently on pod shutdown usually the actor system goes down which makes the materializer stop which makes the stream stop "hard". Everything which is inflight in the stream at that time just gets dropped which leaves you in mixed state if you have asnyc steps inside the stream talking to external systems. 

A graceful shutdown would especially be nice in controlled maintenance scenarios (deployment, scale down, when kubernetes cluster moves pods ...)

**Is your feature request related to a specific runtime of cloudflow or applicable for all runtimes?**

**Describe the solution you'd like**
Being able to configure that a streamlet should stop gracefully when the system/pod stops. The source should complete (or a killswitch right after the source should trigger a shutdown) - so no new messages are consumed from kafka but messages already inflight are processed gracefully. (This would than also most likely require to being able to define a shutdown timeout and configure it also for the pod).

**Describe alternatives you've considered**
1. Hooking into the CoordinatedShutdown and calling stop() on the streamlet -> since the killswitch is placed at the end this will just cancel all messages and currently I also suspect the the streamlet "completes" once the cancel reaches the source and not the complete the sink

```
override def createLogic(): AkkaStreamletLogic = new RunnableGraphStreamletLogic() {
    override def runnableGraph(): RunnableGraph[_] = {
    
      context.onStop(ShutdownLoggingHook.shutdownHook(system, context.streamletRef))

      CoordinatedShutdown(system).addTask(CoordinatedShutdown.PhaseServiceUnbind, "ProcessingShutdown") { () =>
        log.info("Stopping Processing in CoordinatedShutdown")
        context.stop().map(_ => Done)
      }

      sourceWithCommittableContext(in)
        .via(simpleflow)
        .to(committableSink(out))
    }
  }
```
2. adding a killswitch after the source and routing all message to a second sink for completion detection
```
override def createLogic(): AkkaStreamletLogic = new RunnableGraphStreamletLogic() {
    override def runnableGraph(): RunnableGraph[_] = {
     
      val sharedKillSwitch = KillSwitches.shared("graceful-shutdown-killswitch")

      val alsoToPromise = Promise[Done]()
      CoordinatedShutdown(system).addTask(CoordinatedShutdown.PhaseServiceUnbind, "ProcessingShutdown") { () =>
        log.info("Stopping Processing in CoordinatedShutdown")
      
        sharedKillSwitch.shutdown()
[...]

PR #861, opened 2020-11-12
Controlled shutdown of Kafka consumers

ennru:

This implements a more controlled Consumer shutdown as described in https://doc.akka.io/docs/alpakka-kafka/current/consumer.html#controlled-shutdown

References #848 

This lowers the risk of messages being half processed during streamlet shutdown.
It improves chances that commits reach the Kafka consumer during shutdown.

A new streamlet setting `consumer-stop-timeout` is introduced.
The underlying `akka.kafka.consumer.stop-timeout` can be overwritten to 0s for streamlets.

### DIFF
--- a/core/cloudflow-akka/src/main/scala/cloudflow/akkastream/AkkaStreamletContextImpl.scala
+++ b/core/cloudflow-akka/src/main/scala/cloudflow/akkastream/AkkaStreamletContextImpl.scala
@@ -59,9 +59,7 @@ final class AkkaStreamletContextImpl(
   override def config: Config = streamletDefinition.config
 
   private val consumerStopTimeout: FiniteDuration =
-    if (config.hasPath("consumer-stop-timeout"))
-      FiniteDuration(config.getDuration("consumer-stop-timeout").toMillis, TimeUnit.MILLISECONDS).toCoarsest
-    else 5.seconds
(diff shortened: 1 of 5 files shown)


core/cloudflow-akka/src/main/scala/cloudflow/akkastream/AkkaStreamletContextImpl.scala
```diff
@@ -36,24 +39,86 @@ import org.apache.kafka.clients.consumer.ConsumerConfig
...
+  private val streamletDefinitionMsg: String = s"${streamletDefinition.streamletRef} (${streamletDefinition.streamletClass})"
 
   implicit val system: ActorSystem = sys
 
   override def config: Config = streamletDefinition.config
 
+  private val consumerStopTimeout: FiniteDuration =
+    if (config.hasPath("consumer-stop-timeout"))
+      FiniteDuration(config.getDuration("consumer-stop-timeout").toMillis, TimeUnit.MILLISECONDS).toCoarsest
+    else 5.seconds
```
ennru, line 64: Where in the configs would the default need to be added?

RayRoestenburg: In the reference.conf of cloudflow-akka?

ennru: That would work with an independent default.
Is there a place where the `consumer-stop-timeout` default would be propagated as a default to the streamlet configs? (Is there a config which all streamlet configs inherit from?)

Issue #862, opened 2020-11-12
Migrate off EmbeddedKafka


ennru:
The local runner starts Kafka embedded in the JVM. That does not work with Kafka 2.6+ as the Kafka libraries are compiled with inlining. 

Instead, Testcontainers Kafka could be an alternative.

andreaTP:
Fixed in #880

PR #873, opened 2020-11-16
Downgrade Sbt version for users

andreaTP:
Downgrade Sbt version for the users

`runLocal` is broken with sbt 1.4

Long story short:
 - we run Embedded Kafka from sbt
 - Embedded Kafka inline the Scala version
 - The inlined version happens to be 2.12.11
 - Everything works with sbt 1.3 that uses 2.12.10 (that probably gets evicted in favor of 2.12.11 declared in our own plugin)
 - It doesn't work with sbt 1.4 since Scala becomes 2.12.12

Yes and is documented accordingly in the docs

`runLocal` in taxi-ride


franciscolopezsancho (approved):

ennru (approved):
LGTM.

ennru:
See even https://github.com/lightbend/cloudflow/issues/862

RayRoestenburg (approved):

PR #886, opened 2020-11-19
fix project import on IntelliJ

octonato:
This fix the import in IntelliJ. 

The problem was that the operator projects were only building for 2.13, but the main scala version was 2.12. IntelliJ will then use 2.12 as default and when it comes to the operator projects, it can't find a build for the blueprint 2.13. 

If we have them all for 2.12 and operator projects for 2.12 and 2.13 the problem goes away. And I don't think we want to drop 2.12 for the operator either.


core/project/Dependencies.scala
```diff
@@ -8,8 +8,9 @@ object Version {
   val AkkaMgmt      = "1.0.8"
   val AkkaGrpc      = "1.0.1"
   val AlpakkaKafka  = "2.0.5"
-  val Scala         = "2.12.11"
-  val ScalaOperator = "2.13.3"
+  val Scala212      = "2.12.11"
+  val Scala213      = "2.13.3"
```
octonato, line 12: I renamed them to Scala212 and Scala213 so we clear see which version is behind.

core/build.sbt
```diff
@@ -4,63 +4,6 @@ import Library._
 import sbtdocker.Instructions
 import sbtrelease.ReleaseStateTransformations._
 
-val javadocDisabledFor = Set(
```
octonato, line 7: I move this down. Annoying to have it at the top. Need to scroll to the interesting parts and when doing a search we first hit the project names in this section.

RayRoestenburg:
thanks @octonato we stole most of this and it is merged already :-)

PR #888, opened 2020-11-22
Adding error handler for incoming messages. WIP

blublinsky:

### DIFF
--- a/core/cloudflow-akka-testkit/src/main/scala/cloudflow/akkastream/testkit/TestContext.scala
+++ b/core/cloudflow-akka-testkit/src/main/scala/cloudflow/akkastream/testkit/TestContext.scala
@@ -24,7 +24,6 @@ import akka.actor.typed.scaladsl.adapter._
 import akka.cluster.sharding.typed.scaladsl.{ ClusterSharding, Entity }
 import akka.kafka.CommitterSettings
 import akka.kafka.ConsumerMessage._
-import akka.stream._
 import akka.stream.scaladsl._
 import com.typesafe.config._
 import cloudflow.akkastream._
--- a/core/cloudflow-akka-testkit/src/main/scala/cloudflow/akkastream/testkit/TestContext.scala
+++ b/core/cloudflow-akka-testkit/src/main/scala/cloudflow/akkastream/testkit/TestContext.scala
@@ -50,10 +50,15 @@ private[testkit] case class TestContext(
     StreamletDefinition("appId", "appVersion", streamletRef, "streamletClass", List(), volumeMounts, config)
 
   @deprecated("Use `sourceWithCommittableContext` instead.", "1.3.4")
-  override def sourceWithOffsetContext[T](inlet: CodecInlet[T]): cloudflow.akkastream.scaladsl.SourceWithOffsetContext[T] =
+  override def sourceWithOffsetContext[T](
+      inlet: CodecInlet[T],
+      dataconverter: InletDataPConverter[T] = DefaultInletDataPConverter[T]
+  ): cloudflow.akkastream.scaladsl.SourceWithOffsetContext[T] =
     sourceWithContext(inlet)
 
-  override def sourceWithCommittableContext[T](inlet: CodecInlet[T]) = sourceWithContext(inlet)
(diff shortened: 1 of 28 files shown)


blublinsky:
This is an invitation to discussion for approach. There are still unresolved issues. There is still need to implement this is=n Spark

core/cloudflow-akka-testkit/src/main/scala/cloudflow/akkastream/testkit/TestContext.scala
```diff
@@ -50,10 +50,15 @@ private[testkit] case class TestContext(
     StreamletDefinition("appId", "appVersion", streamletRef, "streamletClass", List(), volumeMounts, config)
 
   @deprecated("Use `sourceWithCommittableContext` instead.", "1.3.4")
-  override def sourceWithOffsetContext[T](inlet: CodecInlet[T]): cloudflow.akkastream.scaladsl.SourceWithOffsetContext[T] =
+  override def sourceWithOffsetContext[T](
```
franciscolopezsancho, line 53: If we just change the name and create a brand new API method we'lll avoid a breaking change. What do you think?

blublinsky: We can, but I was hoping not to do this by introducing default values

blublinsky:
Added documentation, cleaned up the code. Has not run tests yet

blublinsky:
And now tested with all 3 runtimes. Its done

core/cloudflow-akka-testkit/src/main/scala/cloudflow/akkastream/testkit/TestContext.scala
RayRoestenburg: If we need to break things it seems it makes more sense to change `Codec`. The `DataConverter` introduces a too similar thing to codec, let's discuss.

RayRoestenburg:
```scala
val inlet = ProtoInlet(...).handleErrors { bytes =>
  // do what you like here, return unit
}
```

RayRoestenburg (changes requested):
Good for a discussion piece but I think we need to take a different approach for the API.

core/cloudflow-akka-testkit/src/main/scala/cloudflow/akkastream/testkit/TestContext.scala
blublinsky: The reason I was doing it this way is to allow users to overwrite default implementation, which is harder if we will try to build it in codec. It is not about extending codec capabilities, but rather allowing users to customize behavior. We can not really foresee all of possible things that the users will want to do

RayRoestenburg: it's the same right, just add a function that handles it to the codec, with some convenience. Also, there is not much you can choose really. all you can do is `Array[Byte] => Unit`.

RayRoestenburg: And I think it makes sense to (at least)  add a `BytesInlet`, which is like `Codec[Array[Byte]]`. Then you can read in any way you like, but that kind of inlet has no schema so will never match an outlet, for safety.

blublinsky: I was suggesting this to you several times, but you were always hesitant about it. There are several things that are possible here. Also you are correct, the only thing we can do is  Array[Byte] => Unit, but there are a couple of things here:
1. How do you get to this point - just syntactic validation or both syntactic and semantic.
2. Which side effect do you want - just logging, writing to Dead-letter topic, etc.

blublinsky: Also, the approach that I have taken provides more fine grained configuration. I can control error processing on per stream basis, rather then doing it globally (per streamlet or even pipeline)

PR #891, opened 2020-11-23
Less restrictions on actions.

RayRoestenburg:

### DIFF
--- a/core/cloudflow-operator-actions/src/main/scala/cloudflow/operator/action/Action.scala
+++ b/core/cloudflow-operator-actions/src/main/scala/cloudflow/operator/action/Action.scala
@@ -33,7 +33,7 @@ sealed trait Action {
 
   /**
    * Executes the action using a KubernetesClient.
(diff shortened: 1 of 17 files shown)


andreaTP (approved):

core/cloudflow-operator-actions/src/main/scala/cloudflow/operator/action/Action.scala
```diff
@@ -31,6 +31,12 @@ import skuber.api.patch.Patch
  */
 sealed trait Action {
 
+  /**
+   * Executes the action using a KubernetesClient.
+   * Returns the created or modified resource, or None if the resource is deleted.
+   */
+  def execute(client: KubernetesClient)(implicit sys: ActorSystem, ec: ExecutionContext, lc: LoggingContext): Future[Action]
```
andreaTP, line 38: not related to this PR, but, in general it would be nice to move:
```scala
(implicit sys: ActorSystem, ec: ExecutionContext, lc: LoggingContext)
```
at the constructor level instead of the methods

PR #893, opened 2020-11-24
Fix Flink Docker image creation with upstream artifacts

andreaTP:
Fix the creation of Flink Streamlets Docker images

`sbt-docker` names differently local/upstream binaries

No

Rebuilding `swiss-knife` and adding an sbt test


RayRoestenburg (approved):
LGTM!

PR #900, opened 2020-11-25
Creating streamlet secrets if missing in cli

RayRoestenburg:

Create empty secrets before creating input secret.
cc @andreaTP need to do this as well in the new cli.

This solves an ordering issue with the operator needing to get the secret before creating runner resource (deployment, flinkapp, sparkapp)

No.

- [x] GKE

### DIFF
--- a/kubectl-cloudflow/cmd/deploy_command.go
+++ b/kubectl-cloudflow/cmd/deploy_command.go
@@ -194,7 +194,7 @@ func (opts *deployOptions) deployImpl(cmd *cobra.Command, args []string) {
 	ownerReference := createOrUpdateCloudflowApplication(appClient, applicationSpec)
 
 	appInputSecret = config.UpdateSecretWithOwnerReference(ownerReference, appInputSecret)
-
+	createStreamletSecretsIfMissing(k8sClient, applicationSpec, ownerReference)
 	createOrUpdateAppInputSecret(k8sClient, namespace, appInputSecret)
 
 	serviceAccount := newCloudflowServiceAccountWithImagePullSecrets(namespace)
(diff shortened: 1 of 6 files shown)


kubectl-cloudflow/cmd/deploy_command.go
```diff
@@ -194,7 +194,7 @@ func (opts *deployOptions) deployImpl(cmd *cobra.Command, args []string) {
 	ownerReference := createOrUpdateCloudflowApplication(appClient, applicationSpec)
 
 	appInputSecret = config.UpdateSecretWithOwnerReference(ownerReference, appInputSecret)
-
+	createStreamletSecretsIfMissing(k8sClient, applicationSpec, ownerReference)
```
andreaTP, line 197: adding this method in this place makes the method below to always be an update, correct?

kubectl-cloudflow/config/config.go
```diff
@@ -694,13 +694,40 @@ func addCommandLineArguments(spec cfapp.CloudflowApplicationSpec, config *Config
...
+		Type: corev1.SecretTypeOpaque,
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      name,
+			Namespace: appID,
+			Labels:    labels,
+		},
+	}
+	secretMap := make(map[string]string)
+	secretMap["secret.conf"] = "{}"
+	secretMap["runtime-config.conf"] = "{}"
```
andreaTP, line 722: If I read this change correctly, this is the core part of it, have you tested if just adding those defaults to `createOrUpdateAppInputSecret` make it?

kubectl-cloudflow/cmd/deploy_command.go
RayRoestenburg: No that is a different secret. (input for all streamlets vs secret for streamlet). This has to be improved further TBH.

kubectl-cloudflow/config/config.go
RayRoestenburg: n streamlets -> n empty secrets. 1 app secret.

PR #903, opened 2020-11-27
Akka streamlet cluster sharding connected car example using protobuff…

sfali:
Akka streamlet cluster sharding connected car example using protobuffer instead of avro.

Adding another example using Protobuffer.

Adding another example using Protobuffer.

No

N/A


andreaTP:
Hi @sfali ! And thanks a lot for your PR!
First of all, let me apologize for getting back to you so late.

Although we truly appreciate the effort you pushed into this PR, adding as well paragraphs to the documentation, we have to acknowledge that maintaining examples is a huge effort for us maintainers.
We need to be careful since this is code that needs to be maintained over time and over the repository changes.
In this specific case, we are adding an example that already exists and changing the serialization protocol, the serialization protocol itself is already demoed in other examples.
Basically, AFAIU, here we are not, unfortunately, improving the matrix of configurations covered by the examples.

I should, therefore, ask you if there is any compelling argument that makes this PR a real value for the current codebase 😄

andreaTP:
@sfali thanks again for this contribution, but I'm going to close this PR due to inactivity.
Please feel free to re-open it or discuss here around the point I made in the comment above 🙂

PR #906, opened 2020-11-30
WIP, migrate Actions from Skuber to Fabric8

andreaTP:
Sharing for early comments and feedback (don't even know if it works ...)

Quick recap:
 - having separate Fabric8Actions breaks way too many signatures
 - I'm currently relying on some reflection magic, and don't know if I can get rid of it ...
 - having both the skuber and fabric8 clients around is not nice, but is a temporary solution

### DIFF
--- a/core/cloudflow-operator-actions/src/main/scala/cloudflow/operator/action/Action.scala
+++ b/core/cloudflow-operator-actions/src/main/scala/cloudflow/operator/action/Action.scala
@@ -389,7 +389,8 @@ class UpdateStatusAction[T <: ObjectResource](
 final case class DeleteAction[T <: ObjectResource, B <: Fabric8HasMetadata](val resourceName: String, _namespace: String)(
     implicit ct: ClassTag[B]
 ) extends ResourceAction[T] {
-  import io.fabric8.kubernetes.api.model.DeletionPropagation
+  import io.fabric8.kubernetes.api.model.{ DeletionPropagation, ObjectMetaBuilder }
+  import io.fabric8.kubernetes.client.utils.Serialization
 
   val name      = "delete"
   val namespace = Some(_namespace)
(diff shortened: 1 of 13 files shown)


core/cloudflow-operator-actions/src/main/scala/cloudflow/operator/action/Action.scala
```diff
@@ -208,195 +199,243 @@ abstract class SingleResourceAction[T <: ObjectResource] extends ResourceAction[
...
-        .getOrElse(recoverFromError(client.create(resource), client, nextRetries, executeCreateOrPatch))
+        .map(_ =>
+          recoverFromError(
+            Future {
+              client.secrets().inNamespace().withName()
+              client
+                .resource(resource)
+                .inNamespace(resource.getMetadata.getNamespace)
+                .get()
+                .asInstanceOf[Resource[T, Doneable]] // TODO: this is not going to work ...
```
andreaTP, line 301: @RayRoestenburg fyi:
This is still a problem, probably `Resource[T, Doneable]` should pop up somewhere in the signature.

core/cloudflow-operator-actions/src/main/scala/cloudflow/operator/action/Action.scala
```diff
@@ -208,195 +199,243 @@ abstract class SingleResourceAction[T <: ObjectResource] extends ResourceAction[
...
   override def executing =
     s"Deleting $resourceName resource"
   override def executed =
     s"Deleted $resourceName resource"
 
+  private val log = LoggerFactory.getLogger(this.getClass.getName)
+
+  private val meta = {
+    val default = {
+      Serialization
```
andreaTP, line 405: @RayRoestenburg fyi:
This looks like a somehow "better way" than using default constructors (this way all the properties injected via Json got populated)

core/cloudflow-operator-actions/src/main/scala/cloudflow/operator/action/Action.scala
```diff
@@ -208,195 +199,243 @@ abstract class SingleResourceAction[T <: ObjectResource] extends ResourceAction[
...
+        .readValue("{}", ct.runtimeClass)
+        .asInstanceOf[T]
+    }
+
+    val metadata = new ObjectMetaBuilder()
+      .withName(resourceName)
+      .withNamespace(_namespace)
+      .build()
+
+    default.setMetadata(metadata)
```
andreaTP, line 416: @RayRoestenburg fyi:
At this point Metadata is empty and we can fill it directly.

RayRoestenburg:
Decided not to do this.

PR #911, opened 2020-12-04
Deprecate runtime specific base images

andreaTP:
Deprecate the usage of runtime-specific base images in favor of giving full control to the user over the generation of the streamlets images.

To have a self-contained build of the streamlet's images and get rid of the separate generation of the runtime base images.

Yes in the sbt plugin:

 - `cloudflowDockerParentImage` (previously unused) has been renamed `cloudflowDockerBaseImage` and used to store the docker image we use as `from`
 - `cloudflowFlinkBaseImage`, `cloudflowSparkBaseImage` and `cloudflowAkkaBaseImage` are now deprecated, this means that users that are using them will not benefit from this change and they should migrate to `cloudflowDockerBaseImage`
 - the first time users will build and push they are going to miss caches and it will take some time

adding more sbt scripted tests, running the integration tests and `taxi-ride` with this change in.


core/sbt-cloudflow/src/main/resources/runtimes/flink/config.sh
```diff
@@ -405,21 +405,9 @@ if [ -z "${FLINK_PID_DIR}" ]; then
     FLINK_PID_DIR=$(readFromConfig ${KEY_ENV_PID_DIR} "${DEFAULT_ENV_PID_DIR}" "${YAML_CONF}")
 fi
 
+if [ -z "${FLINK_ENV_JAVA_OPTS}" ]; then
```
RayRoestenburg, line 408: @franciscolopezsancho

core/sbt-cloudflow/src/main/resources/runtimes/flink/flink-console.sh
```diff
@@ -70,4 +70,4 @@ if [[ ${JAVA_VERSION} =~ ${IS_NUMBER} ]]; then
 fi
 
 echo "Starting $SERVICE as a console application on host $HOSTNAME."
-exec $JAVA_RUN $JVM_ARGS "${log_setting[@]}" ${FLINK_ENV_JAVA_OPTS} -classpath "`manglePathList "$FLINK_TM_CLASSPATH:$INTERNAL_HADOOP_CLASSPATHS"`" ${CLASS_TO_RUN} "${ARGS[@]}"
```
RayRoestenburg, line 73: @franciscolopezsancho

core/sbt-cloudflow/src/main/resources/runtimes/flink/flink-entrypoint.sh
```diff
@@ -32,11 +32,6 @@ drop_privs_cmd() {
 }
 
 # Add in extra configs set by the operator
-# If not already present
-if  ! cat "$FLINK_HOME/conf/flink-conf.yaml" | grep -q "taskmanager.memory.flink.size" ; then
```
RayRoestenburg, line 36: This should go back in

core/sbt-cloudflow/src/main/resources/runtimes/flink/flink-console.sh
franciscolopezsancho: Well spotted @RayRoestenburg , this was in that order to allow the user to set logging options through conf. If `${FLINK_ENV_JAVA_OPTS}` goes after  `${log_setting[@]}` no override of logging would be possible.

PR #913, opened 2020-12-10
Debugging remote

franciscolopezsancho:
adding remote debug capability 

UX

yes, the user can attach a debugger

adding the pertinent configuration and adding breaking points into IntelliJ. Done for all the base images. Breaking points behaved as usual.


docs/shared-content-source/docs/modules/develop/pages/debugging-remote.adoc
```diff
@@ -0,0 +1,41 @@
...
+Once this is added and the application is deployed we'll need to do two things. 
+
+
+1. Adding a remote configuration in the IDE of our choice. When adding this remote Run/Debug configuration we have to make sure that it has these arguments for the remote JVM.
+	
+	-agentlib:jdwp=transport=dt_socket,server=y,suspend=n,address=*:5005
+
+ Also, this remote configuration shall point to 'localhost' in the port of our choice. 
+
+2. Kubernetes port forwarding from our local to the selected streamlet. Let's say the port of our choice in the IDE is 5005 and our application is called 'example-application'. Then we will need to port forwarding like the following:
```
andreaTP, line 39: Does this step work even if we don't expose the port while deploying the streamlet? Is it already exposed by default by the operator?

docs/shared-content-source/docs/modules/develop/pages/debugging-remote.adoc
```diff
@@ -0,0 +1,41 @@
...
+
+include::ROOT:partial$include.adoc[]
+
+
+Debugging a remote application is possible just by applying the https://docs.oracle.com/javase/8/docs/technotes/guides/jpda/conninv.html#Invocation[standard] configuration to debugging a remote JVM. That is, adding this conf to JAVA_OPTS through our xref:cloudflow-configuration.adoc[configuration model]. 
+	
+	-agentlib:jdwp=transport=dt_socket,server=y,suspend=n,address=5005
+
+
+NOTE: the address can NOT be changed, it shall remain 5005.
```
andreaTP, line 11: we can bind the address to an sbt `settingKey`, if a user wants to debug multiple applications or in different environments, this sounds like a limitation, WDYT?

core/sbt-cloudflow/src/main/scala/cloudflow/sbt/CloudflowAkkaPlugin.scala
```diff
@@ -90,7 +90,8 @@ object CloudflowAkkaPlugin extends AutoPlugin {
         Instructions.User(UserInImage),
         Instructions.EntryPoint.exec(Seq("bash", "/opt/akka-entrypoint.sh")),
         Instructions.Copy(sources = Seq(CopyFile(depJarsDir)), destination = OptAppDir, chown = Some(userAsOwner(UserInImage))),
-        Instructions.Copy(sources = Seq(CopyFile(appJarsDir)), destination = OptAppDir, chown = Some(userAsOwner(UserInImage)))
+        Instructions.Copy(sources = Seq(CopyFile(appJarsDir)), destination = OptAppDir, chown = Some(userAsOwner(UserInImage))),
+        Instructions.Expose(Seq(5005))
```
andreaTP, line 94: Thinking twice about it, I think we should make this feature optional.
For security reasons, I can easily see customers that do want to not expose ports in production environments.

andreaTP (commented):
I left a few suggestions, also, we can easily build up a simple `scripted` test to at least verify that the port is exposed at the docker level to avoid regressions in the future.
Available to help in case!

RayRoestenburg:
Fails on scalafmt.

core/sbt-cloudflow/src/main/scala/cloudflow/sbt/CloudflowAkkaPlugin.scala
RayRoestenburg: Can we put this in as an optional sbt setting?

RayRoestenburg (commented):
Would be nice to have the debug port as a setting in sbt.

docs/shared-content-source/docs/modules/develop/pages/debugging-remote.adoc
```diff
@@ -0,0 +1,41 @@
...
+----
+
+Once this is added and the application is deployed we'll need to do two things. 
+
+
+1. Adding a remote configuration in the IDE of our choice. When adding this remote Run/Debug configuration we have to make sure that it has these arguments for the remote JVM.
+	
+	-agentlib:jdwp=transport=dt_socket,server=y,suspend=n,address=*:5005
+
+ Also, this remote configuration shall point to 'localhost' in the port of our choice.
```
RayRoestenburg, line 37:
```suggestion
The above remote configuration example points to 'localhost' and port 5005. 
```
Is maybe better?

docs/shared-content-source/docs/modules/develop/pages/debugging-remote.adoc
```diff
@@ -0,0 +1,41 @@
...
+            { name = "JAVA_OPTS" 
+              value = "-agentlib:jdwp=transport=dt_socket,server=y,suspend=n,address=5005"
+            }
+          ]
+    }
+  }
+}
+----
+
+Once this is added and the application is deployed we'll need to do two things.
```
RayRoestenburg, line 30:
```suggestion
Once this is added and the application is deployed we'll need to do two things:
```

docs/shared-content-source/docs/modules/develop/pages/debugging-remote.adoc
```diff
@@ -0,0 +1,41 @@
...
+          ]
+    }
+  }
+}
+----
+
+Once this is added and the application is deployed we'll need to do two things. 
+
+
+1. Adding a remote configuration in the IDE of our choice. When adding this remote Run/Debug configuration we have to make sure that it has these arguments for the remote JVM.
```
RayRoestenburg, line 33:
```suggestion
1. Add a remote configuration in the IDE of our choice. When adding this remote Run/Debug configuration we have to make sure that it has these arguments for the remote JVM.
```

docs/shared-content-source/docs/modules/develop/pages/debugging-remote.adoc
RayRoestenburg, line 39:
```suggestion
2. Port-forward from our local to the selected streamlet in the Kubernetes cluster. Let's say the port of our choice in the IDE is 5005 and our application is called 'example-application'. Then we will need to port-forward like in the following example:
```

franciscolopezsancho:
I'll pursue open port and passing JVM params all through configuration at deploy time.

PR #922, opened 2020-12-14
adding container ports to akka streamlet through configuration, annotations for akka, spark and flink.

franciscolopezsancho:
adding remote debug capability 

UX

yes, the user can attach a debugger

adding the pertinent configuration and adding breaking points into IntelliJ. Done for all the base images. Breaking points behaved as usual.


RayRoestenburg:
It's interesting that the Flink CR and Spark CR have no ability to add container ports to the container.
@franciscolopezsancho this could still be interesting for akka, so you can at least provide the info that you are going to expose a port. But as discussed for debugging purposes we don't need this addition. 

For other readers: It's interesting that container ports on container is only for information purposes:

https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.20/#container-v1-core

```
List of ports to expose from the container. Exposing a port here gives the system additional information about the network connections a container uses, but is primarily informational. Not specifying a port here DOES NOT prevent that port from being exposed. Any port which is listening on the default "0.0.0.0" address inside a container will be accessible from the network. Cannot be updated.
```

RayRoestenburg (commented):
As discussed probably not needed.

franciscolopezsancho:
Do you think @RayRoestenburg  we should add this info in the docs?

RayRoestenburg:
@franciscolopezsancho no, those are kubernetes details I would say.

RayRoestenburg:
RE-opening this since we need this, and nice to add annotations too, for adding ports for prometheus to scrape.

andreaTP (approved):

andreaTP (commented):
A couple of questions but in general looks good

PR #924, opened 2020-12-16
Easy logging

andreaTP:


core/cloudflow-akka/src/main/resources/reference.conf
```diff
@@ -10,3 +10,11 @@ akka.kafka.consumer.stop-timeout = 0s
 # Configure the time to wait after stopping the inflow before shutting down the Kafka consumer
 # This helps to process in-flight messages (and commits).
 cloudflow.akka.consumer-stop-timeout = 10 s
+
+# logback configuration
+# https://doc.akka.io/docs/akka/current/logging.html#slf4j
+akka {
```
RayRoestenburg, line 16: FYI Patrik has pointed out that overrides of reference.confs in a reference.conf does not always work. (it works so far in Cloudflow but that is likely because of ordering)

core/build.sbt
```diff
@@ -117,6 +117,8 @@ lazy val streamlets =
             Ficus,
             Avro,
             Bijection,
+            LogbackClassic,
```
RayRoestenburg, line 120: Would prefer if we don't do this, and let users add logback, and we tell them they should.

core/sbt-cloudflow/src/main/scala/cloudflow/sbt/CloudflowFlinkPlugin.scala
```diff
@@ -94,6 +95,10 @@ object CloudflowFlinkPlugin extends AutoPlugin {
             Seq("rm", flinkTgz),
             Seq("cp", "/tmp/config.sh", s"${flinkHome}/bin/config.sh"),
             Seq("cp", "/tmp/flink-console.sh", s"${flinkHome}/bin/flink-console.sh"),
+            // logback configuration:
+            // https://ci.apache.org/projects/flink/flink-docs-stable/deployment/advanced/logging.html#configuring-logback
+            // logback is provided by the streamlet itself
```
RayRoestenburg, line 100:
```suggestion
            // logback must be provided by the streamlet itself
```

core/sbt-cloudflow/src/main/scala/cloudflow/sbt/CloudflowSparkPlugin.scala
```diff
@@ -58,10 +58,16 @@ object CloudflowSparkPlugin extends AutoPlugin {
             IO.delete(depJarDir)
 
             projectJars.foreach { jar =>
-              IO.copyFile(jar, new File(appJarDir, jar.getName))
+              // Logback configuration
+              // dependencies are filetered out here to preserve the behavior in runLocal
```
RayRoestenburg, line 62:
```suggestion
              // dependencies are filtered out here to preserve the behavior in runLocal
```

core/sbt-cloudflow/src/main/scala/cloudflow/sbt/CloudflowSparkPlugin.scala
```diff
@@ -58,10 +58,16 @@ object CloudflowSparkPlugin extends AutoPlugin {
...
+              if (jar.getName.startsWith("slf4j-log4j12-") && jar.getName.startsWith("log4j-"))
+                IO.copyFile(jar, new File(appJarDir, jar.getName))
             }
             depJars.foreach { jar =>
-              IO.copyFile(jar, new File(depJarDir, jar.getName))
+              // Logback configuration
+              // dependencies are filetered out here to preserve the behavior in runLocal
```
RayRoestenburg, line 68:
```suggestion
              // dependencies are filtered out here to preserve the behavior in runLocal
```

core/build.sbt
```diff
@@ -459,7 +458,8 @@ lazy val operatorActions =
       libraryDependencies ++= Vector(
             AkkaSlf4j,
             Ficus,
-            Logback,
+            LogbackClassic,
```
RayRoestenburg, line 461: I guess we should not do tis in operatorActions (we should remove it, add dependency to operator)

PR #927, opened 2020-12-17
Check whitesource in CI

andreaTP:
Forgot to include on the previous PR


RayRoestenburg (approved):

ennru:
You need to provide the secret.

I believe having a separate workflow makes sense. eg https://github.com/lightbend/akka-platform-operator/blob/main/.github/workflows/license-check.yml

PR #928, opened 2020-12-18
Update whitesource results from CI

andreaTP:


.github/workflows/build_and_publish.yaml
```diff
@@ -12,18 +12,19 @@ jobs:
...
-        make all
+        git fetch --depth=1 origin +refs/tags/*:refs/tags/*
     - name: Whitesource
       env:
         WHITESOURCE_PASSWORD: ${{ secrets.WHITESOURCE_PASSWORD }}
       run: |
-        git fetch --depth=1 origin +refs/tags/*:refs/tags/*
         cd core
         sbt "whitesourceCheckPolicies"
+        sbt "whitesourceUpdate"
```
ennru, line 22: A single start of sbt would be enough `sbt "whitesourceCheckPolicies; whitesourceUpdate"`

ennru (approved):
LGTM.

.github/workflows/build_and_publish.yaml
andreaTP: right, fixed!

PR #935, opened 2020-12-23
Remove a bunch of intermediate releases from docs to avoid Antora OOM

andreaTP:
The CI is failing with OOM while building the documentation, and I can reproduce locally.

This is a workaround to make the CI green again, any tip @rasummer @rstento ?


RayRoestenburg (approved):

rstento:
@andreaTP, I think it's a good idea, just for usability. I'm not sure why it was failing, but I'm glad you were able to fix.

PR #946, opened 2021-01-06
Index only current and remove development documentation

andreaTP:
This PR index only the `latest` branch, and we remove the `development` documentation as suggested by @rstento .

We have to note that `latest` is not forcefully matching `current` and this might cause quirks.

E.g.:
if we update an older version of the docs we probably need to push a commit to the `current` to re-index the correct version.
Fortunately, this doesn't happen very often and we can probably live with it.

### DIFF
--- a/docs/antora-docker/Dockerfile
+++ b/docs/antora-docker/Dockerfile
@@ -7,6 +7,7 @@ RUN yarn global add antora-site-generator-lunr
 ENV DOCSEARCH_ENABLED=true
 ENV DOCSEARCH_ENGINE=lunr
 ENV DOCSEARCH_LANGS=en
+ENV DOCSEARCH_INDEX_VERSION=latest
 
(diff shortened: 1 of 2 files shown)


RayRoestenburg (approved):

rstento (commented):
@andreaTP I can't get my local build to work correctly to test this. I did the same changes you did on a test branch and when I entered a search, I was still getting the wrong page version. I am done for today, so will try to look at this again tomorrow. I have a suspicion that lunr doesn't know how to deal with the antora.yml and base-antora.yml, but maybe I'm wrong.

andreaTP:
@rstento , we check tomorrow on the official site (I merged and it's going to end up in cloudflow.io 🙂 ), I'm done as well for today.
I have re-written the docker image on dockerhub, make sure you pull it again.

andreaTP:
`base-antora.yaml` is completely transparent to antora (e.g. it's useful just for our automation process)

PR #947, opened 2021-01-07
Use alpine moving tag for base images

andreaTP:
Alpine is already the latest in the operator versions:

```
docker run --rm -it --entrypoint=/bin/sh lightbend/cloudflow-operator:2.0.21
/ # cat /etc/os-release
NAME="Alpine Linux"
ID=alpine
VERSION_ID=3.12.3
PRETTY_NAME="Alpine Linux v3.12"
HOME_URL="https://alpinelinux.org/"
BUG_REPORT_URL="https://bugs.alpinelinux.org/"
```

Here we set the moving tag for base images as well.


RayRoestenburg:
cc @RoryStokes

RayRoestenburg (approved):

Issue #297, opened 2020-04-23
Install and deploy Cloudflow in a single namespace.


michaelpnash:
Customers (BP, ING) report that company policy dictates they can only run in a single namespace.  Obviously this will not be a problem when we offer a managed service, but may be an issue for our “on-prem” business.

PR #949, opened 2021-01-13
Proposal to allow shorter K8S object names for pods, services, config-maps, etc.

thomasschoeftner:
For the past year, we've continously run into various object name length limitations issued by K8S - mostly the 63character limit.
Unfortunately, the object names generated by Cloudflow are very long as the usually consist of `applid-streamletname`.

A quick test with a manually modified CR (containing a few streamlets and HTTP endpoints) showed that
- the actual magic for shorter object names can be handled by the sbt-cloudflow plugin (by creating shorter names in the CRs)
- the names for K8S services are still pretty long due to to way how service-names are computed (function `Names.ofSerivce` is misleading here because changing it would not impact the actual service name).

This PR proposes 2 changes:
1. Use the deployment-name from the CR instead of calculating the K8S service name from the endpoint configuration in the CR. With the current implementation of Cloudflow (2.0.21 and earlier) the computed Service name would always amount to `appid-streamletname` (-prefix) - same as the deployment name.
2. add setting key `useShortK8sObjectNames` to sbt-cloudflow in order to yield shorter deployment names (which boil down to K8S object names of pods, services, config-maps) during CR creation.

The existing tests are still running through.
I added another test to actually verify the creation of short K8S object names.
Aside from that long object names (as used until now) remain the default.

### DIFF
--- a/core/cloudflow-blueprint/src/main/scala/cloudflow/blueprint/deployment/ApplicationDescriptor.scala
+++ b/core/cloudflow-blueprint/src/main/scala/cloudflow/blueprint/deployment/ApplicationDescriptor.scala
@@ -53,7 +53,7 @@ object ApplicationDescriptor {
             image: String,
             blueprint: VerifiedBlueprint,
             agentPaths: Map[String, String],
-            libraryVersion: String): ApplicationDescriptor = {
+            libraryVersion: String)(useShortK8sObjectNames: Boolean): ApplicationDescriptor = {
 
     val sanitizedApplicationId    = Dns1123Formatter.transformToDNS1123Label(appId)
     val namedStreamletDescriptors = blueprint.streamlets.map(streamletToNamedStreamletDescriptor)
@@ -61,7 +61,12 @@ object ApplicationDescriptor {
       namedStreamletDescriptors
         .map {
           case (streamlet, instance) =>
-            StreamletDeployment(sanitizedApplicationId, instance, image, portMappingsForStreamlet(streamlet, blueprint))
+            val sd = StreamletDeployment(sanitizedApplicationId, instance, image, portMappingsForStreamlet(streamlet, blueprint))
+            if (useShortK8sObjectNames) {
+              sd.withShortName()
+            } else {
+              sd
+            }
         }
 
     ApplicationDescriptor(sanitizedApplicationId,
@@ -119,7 +124,11 @@ object StreamletDeployment {
   val ServerAttributeName   = "server"
   val EndpointContainerPort = 3000
 
-  def name(appId: String, streamlet: String) = s"${appId}.${streamlet}"
+  implicit class StreamletDeploymentHelper(deployment: StreamletDeployment) {
+    def withShortName() = deployment.copy(name = deployment.streamletName)
+  }
+
+  def longName(appId: String, streamlet: String) = s"${appId}.${streamlet}"
 
   def apply(appId: String,
             streamlet: StreamletInstance,
@@ -129,7 +138,7 @@ object StreamletDeployment {
             replicas: Option[Int] = None): StreamletDeployment = {
     val (config, endpoint) = configAndEndpoint(appId, streamlet, containerPort)
     StreamletDeployment(
-      name(appId, streamlet.name),
+      longName(appId, streamlet.name),
       streamlet.descriptor.runtime.name,
       image,
       streamlet.name,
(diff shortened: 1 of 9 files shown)


core/cloudflow-operator-actions/src/main/scala/cloudflow/operator/action/EndpointActions.scala
```diff
@@ -40,17 +40,25 @@ object EndpointActions {
     def distinctEndpoints(app: CloudflowApplication.Spec) =
       app.deployments.flatMap(deployment => deployment.endpoint).toSet
 
+    def deploymentsOfEndpoints(app: CloudflowApplication.Spec) =
+      app.deployments
+        .flatMap(deployment => deployment.endpoint.map(ep => ep -> deployment.name))
+        .toMap // use deployment name rather than re-constructing the name again (by duplicating code)
+
+    val currentDeploymentsOfEndpoints = currentApp.map(current => deploymentsOfEndpoints(current.spec)).getOrElse(Map.empty)
```
thomasschoeftner, line 48: maybe naming these 2 maps in plural was not so smart...

core/cloudflow-operator-actions/src/main/scala/cloudflow/operator/action/Name.scala
```diff
@@ -125,7 +125,7 @@ object Name {
     makeDNS1039Compatible(fixDots(streamletDeploymentName))
 
   def ofConfigMap(streamletDeploymentName: String) =
-    makeDNS1123CompatibleSubDomainName(s"configmap-${fixDots(streamletDeploymentName)}")
+    makeDNS1123CompatibleSubDomainName(s"configmap-${fixDots(streamletDeploymentName)}") // TODO - why append '-service' (line 145) but prepend 'configmap-'? - suffixes seem the more k8s-idiomatic choice
```
thomasschoeftner, line 128: just a rant - because the naming seem not too concise.

changing that, however, will likely introduce a whole lot of dead config-maps lying around on K8S.

core/cloudflow-operator-actions/src/main/scala/cloudflow/operator/action/Name.scala
```diff
@@ -142,7 +142,7 @@ object Name {
   val ofContainerPrometheusExporterPort = max15Chars("prom-metrics")
 
   def ofService(streamletDeploymentName: String) =
-    truncateTo63CharactersWithSuffix(makeDNS1039Compatible(ofPod(streamletDeploymentName)), "-service")
+    truncateTo63CharactersWithSuffix(ofPod(streamletDeploymentName), "-service")
```
thomasschoeftner, line 145: DNS1039 compliance is already ensured by `ofPod`

core/cloudflow-blueprint/src/main/scala/cloudflow/blueprint/deployment/ApplicationDescriptor.scala
```diff
@@ -53,15 +53,20 @@ object ApplicationDescriptor {
             image: String,
             blueprint: VerifiedBlueprint,
             agentPaths: Map[String, String],
-            libraryVersion: String): ApplicationDescriptor = {
+            libraryVersion: String)(useShortK8sObjectNames: Boolean): ApplicationDescriptor = {
```
thomasschoeftner, line 56: an `apply` with 2 parameter lists is not pretty, but
- with objects it's hard to introduce configurable behavior (unless we repeatedly load a HOCON config in various objects, for instance)
- at some point the compiler expects a generic apply method `apply(?, ?, ?, ? ,?, ?, ?)`. Adding the new parameter to the other params would have confused the compiler here
- `useShortK8SObjectNames` is really configuration how to process and not a value for a data field, so a separate parameter list seems right.

RayRoestenburg:
Hi @thomasschoeftner after this PR we had a discussion that would improve naming and also make it possible to deploy many applications in one namespace. Is this interesting for you to pickup?

RayRoestenburg:
Thanks again @thomasschoeftner, closing this PR, since we came up with a better idea for how to solve this on Zulip recently. Let us know if you need any help with the design, would be great if you could add an issue for the new way of naming generated resources, also supporting https://github.com/lightbend/cloudflow/issues/297 (single namespace for cloudflow and all apps).

thomasschoeftner:
thanks for taking a look.
I'm planning to tackle the proposed solution soon, but I'll create a dedicated ticket to describe the what and how first.

RayRoestenburg:
Thanks @thomasschoeftner looking forward to it!

PR #951, opened 2021-01-13
upgrade to ccpa banner

rasummer:
@andreaTP this updates the cookie preference center to be ccpa compliant for the docs and the homepage.

screenshot-localhost_8000-2021 01 13-14_28_14


andreaTP:
Sorry I completely missed it @rasummer , thanks!

andreaTP:
@rasummer the new Cookie banner is pretty aggressive in local development, would you, please, reconsider merging/doing something like:
https://github.com/lightbend/antora-ui-lightbend-theme/pull/67
?

rasummer:
> @rasummer the new Cookie banner is pretty aggressive in local development, would you, please, reconsider merging/doing something like:
> [lightbend/antora-ui-lightbend-theme#67](https://github.com/lightbend/antora-ui-lightbend-theme/pull/67)
> ?

There is a solution already in place when building the docs locally. As long as you are using a local build that uses the [author-mode-site.yml](https://github.com/lightbend/cloudflow/blob/master/docs/docs-source/author-mode-site.yml#L19), then it will use the default layout of `default-dev`. This layout uses the test script and the cookie banner should work normally as long as you are viewing it on a localhost sever such as `python -m SimpleHTTPServer`. The cookie banner won't work if you just open the file directly in the browser, as it needs to drop a cookie against a domain, thus the need for localhost.

There is one other approach we could look at, one that the Akka Platform guide is doing. In it's case for the `html-author-mode` `make` command it is [removing and storing](https://github.com/akka/akka-platform-guide/blob/main/Makefile#L47) the `head-scripts.hbs` file so that it can build the docs with no cookie banner. It then places the [file back in the folder](https://github.com/akka/akka-platform-guide/blob/main/Makefile#L55) once the local build is complete.

andreaTP:
Thanks for the feedback @rasummer 
If it works as expected using `localhost` is already good enough! Thanks 👍

PR #962, opened 2021-01-22
Build Linux static binaries from republished graalvm docker image

andreaTP:
Since last night:
https://hub.docker.com/oracle -> 404


RayRoestenburg (approved):
Looks like a good quick fix! amazing that the Oracle image is not there anymore...

PR #969, opened 2021-01-26
Fix configuration propagation in the local runner

andreaTP:
Fix the propagation of streamlet and runtime conf in `runLocal`

To match the behavior of the application in the cluster.

No

Locally tested on an example project.


RayRoestenburg (approved):
LGTM!

PR #1004, opened 2021-03-18
Removing schema from cr

franciscolopezsancho:
#986
CloudflowApplication size reach etcd request size limit

the CR json hits the limit of etcd

no

unit testing and in GKE cluster checking the console enterprise was working fine.

### DIFF
--- a/core/cloudflow-blueprint/src/main/scala/cloudflow/blueprint/deployment/ApplicationDescriptor.scala
+++ b/core/cloudflow-blueprint/src/main/scala/cloudflow/blueprint/deployment/ApplicationDescriptor.scala
@@ -64,13 +64,18 @@ object ApplicationDescriptor {
             StreamletDeployment(sanitizedApplicationId, instance, image, portMappingsForStreamlet(streamlet, blueprint))
         }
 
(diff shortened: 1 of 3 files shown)


andreaTP:
🎉  love this! We have to coordinate when we merge but is not going to be a big issue.
Have you considered using a different `case class` here:
https://github.com/lightbend/cloudflow/blob/7988d1b3085d97e3f5c31d73c45f4da8d3417373/core/cloudflow-blueprint/src/main/scala/cloudflow/blueprint/deployment/ApplicationDescriptor.scala#L114
to make sure at compile time that this property is not used?

franciscolopezsancho:
:) Thanks @andreaTP! I did consider removing the schema parameter inside that class but it's used to compare inlets and outlets, I found. We can talk about it at any time.

RayRoestenburg (approved):

RayRoestenburg:
@andreaTP @franciscolopezsancho please coordinate to have this merged (possibly after the big changes in operator)?

franciscolopezsancho:
merging with fabric8-action-next

PR #1013, opened 2021-04-01
Descriptor generator to it's own library

andreaTP:
The `StreamletDescriptor`s extraction is now in a separate sub-project.
As a byproduct, the `sbt-plugin` now belongs to the `tools` project.

I'm open to suggestions for the new class names.

Modularize the `sbt-plugin` for easier re-use.

No

tests / scripted tests / runLocal is an example project

_NOTE:_ we do need to update the release process and the IT CI after this got merged.


tools/build.sbt
```diff
@@ -220,3 +229,55 @@ lazy val cloudflowOperator =
       dockerUsername := sys.props.get("docker.username"),
       dockerRepository := sys.props.get("docker.registry"),
       dockerBaseImage := "adoptopenjdk/openjdk11:alpine-jre")
+
+lazy val cloudflowDescriptorGenerator =
```
RayRoestenburg, line 233: Please rename to `cloudflowDescriptorExtractor` I think that is a better name. Use Extractor instead of Generatro everywhere?

tools/cloudflow-descriptor-generator/src/main/scala/akka/cloudflow/DescriptorGenerator.scala
```diff
@@ -0,0 +1,41 @@
+package akka.cloudflow
+
+import com.typesafe.config.{ Config, ConfigFactory, ConfigValueFactory }
+
+import java.io.File
+import java.net.{ URL, URLClassLoader }
+import scala.util.{ Failure, Success }
+
+object DescriptorGenerator {
```
RayRoestenburg, line 9: DescriptorExtractor

tools/cloudflow-descriptor-generator/src/main/scala/akka/cloudflow/Main.scala
```diff
@@ -0,0 +1,30 @@
...
+import com.typesafe.config.{ ConfigFactory, ConfigRenderOptions }
+
+import java.net.URL
+import scala.collection.JavaConverters._
+
+object Main {
+
+  def main(args: Array[String]): Unit = {
+    if (args.size != 1) {
+      Console.err.println("Need exactly one argument that is the configuration file")
```
RayRoestenburg, line 12: What is the format of this file?

RayRoestenburg: We probably want to tune the format of this a bit. Is it 
``
{ 
  projectId = ..., 
  classpath = [], 
  dockerImageName = ...
}
```
?

Comment (RayRoestenburg) [reply tools/cloudflow-descriptor-generator/src/main/scala/akka/cloudflow/Main.scala:12]: I guess I should see this `Main` as a POC at this point..

File: tools/cloudflow-descriptor-generator/src/main/scala/akka/cloudflow/StreamletScanner.scala [hunk 4]
```diff
@@ -14,22 +14,22 @@
  * limitations under the License.
  */
 
-package cloudflow.sbt
+package akka.cloudflow
 
 import java.lang.reflect.Modifier
 
-import io.github.lukehutch.fastclasspathscanner.FastClasspathScanner
+import io.github.classgraph.ClassGraph
```
RayRoestenburg, line 21: 👍

tools/cloudflow-descriptor-generator/src/main/scala/akka/cloudflow/StreamletScanner.scala
```diff
@@ -39,20 +39,30 @@ object StreamletScanner {
...
+      try {
+        Thread.currentThread().setContextClassLoader(classLoader)
+        val nonAbstractClasses: List[Class[_]] =
+          scanResult.map(nonAbstract(_, classLoader)).collect { case Success(x) => x }
+
+        nonAbstractClasses
+          .map(clazz => clazz.getName -> getDescriptor(clazz))
+          .toMap
+      } finally {
+        Thread.currentThread().setContextClassLoader(originalClassLoader)
```
RayRoestenburg, line 63: 👍

tools/cloudflow-descriptor-generator/src/main/scala/akka/cloudflow/StreamletScanner.scala
```diff
@@ -86,16 +96,25 @@ object StreamletScanner {
...
-      Try(streamletClass.getDeclaredConstructor().newInstance()).recoverWith {
-        case error => Failure(ConstructorFailure(streamletClass, error))
-      }
+  private def getInstanceFromDefaultConstructor(streamletClass: Class[_]): Try[Any] = {
+    getNoArgConstructor(streamletClass) match {
+      case Some(constructor) =>
+        try {
+          constructor.newInstance()
+        } catch {
+          case ex: Throwable => ex.getCause.printStackTrace()
```
RayRoestenburg, line 105: This does not seem right?

RayRoestenburg: Would have preferred to leave the original code as is.. I think in this case before it would result in a Failure?

RayRoestenburg: Another thing, that has been on some list.. we should prevent people from defining a Streamlet as `object`. It's not a problem on the cluster, but in local running it can be, if we fork a JVM per set of streamlets in the same library. (we can change that as well, so that we fork a jvm for every object, if needed, but just wanted to mention it FYI)

tools/cloudflow-descriptor-generator/src/test/scala/akka/cloudflow/StreamletScannerSpec.scala
```diff
@@ -14,24 +14,30 @@
...
+    val (valid, invalid) = results.partition {
+      case (_, triedDiscoveredStreamlet) => triedDiscoveredStreamlet.isSuccess
+    }
+    val validStreamlets = valid.collect { case (k, Success(discovered)) => (k, discovered) }
     val invalidStreamlets = invalid.toMap
 
     // These are all valid streamlets defined in TestStreamlets.scala
     "find all valid test streamlets" in {
-      valid must have size 6
+      valid must have size 5
```
RayRoestenburg, line 40: why

tools/cloudflow-descriptor-generator/src/test/scala/akka/cloudflow/StreamletScannerSpec.scala
```diff
@@ -62,37 +68,22 @@ final class StreamletScannerSpec extends WordSpec with TryValues with OptionValu
         .map { confParConf =>
           confParConf.getString("description") mustBe expected.GoldPrice.description
           confParConf.getString("key") mustBe expected.GoldPrice.key
-          confParConf.getString("validation_type") mustBe expected.GoldPrice.toDescriptor.validationType
-        }
-    }
-
-    "find streamlet with Java config parameters" in {
```
RayRoestenburg, line 69: Why is this dropped?

tools/cloudflow-descriptor-generator/src/main/scala/akka/cloudflow/DescriptorGenerator.scala
```diff
@@ -0,0 +1,41 @@
+package akka.cloudflow
```
RayRoestenburg, line 1: `package cloudflow.[something]`

RayRoestenburg (changes requested):
I see some changes in reflective calls to instantiate streamlets and changes in the test I would like to understand, why they were needed. Also found some other minor issues.

tools/cloudflow-descriptor-generator/src/main/scala/akka/cloudflow/Main.scala
andreaTP: yes, do you want to remove it?

tools/cloudflow-descriptor-generator/src/main/scala/akka/cloudflow/StreamletScanner.scala
andreaTP: is a leftover, fixing it

tools/cloudflow-descriptor-generator/src/main/scala/akka/cloudflow/DescriptorGenerator.scala
andreaTP: something = ?
😄

tools/cloudflow-descriptor-generator/src/main/scala/akka/cloudflow/Main.scala
RayRoestenburg: no I think it is OK.

tools/cloudflow-descriptor-generator/src/main/scala/akka/cloudflow/DescriptorGenerator.scala
RayRoestenburg: `cloudflow.descriptor` ?

tools/build.sbt
```diff
@@ -220,3 +229,55 @@ lazy val cloudflowOperator =
...
+  Project(id = "cloudflow-descriptor-generator", base = file("cloudflow-descriptor-generator"))
```
RayRoestenburg, line 234: maybe `cloudflow-descriptor` is good enough as name (also to reflect as cloudflow.descriptor as package)

tools/cloudflow-extractor/src/main/scala/cloudflow/extractor/DescriptorExtractor.scala
```diff
@@ -0,0 +1,45 @@
...
+    streamletDescriptors.flatMap {
+      case (streamletClassName, Success(descriptor)) =>
+        Some(streamletClassName -> descriptor)
+
+      case (_, Failure(error)) =>
+        None
+    }
+  }
+
+  def resolve(config: ResolveConfiguration, streamlets: Map[String, Config]) = {
```
RayRoestenburg, line 32: nice to have return type

RayRoestenburg (approved):
LGTM!

PR #1015, opened 2021-04-05
[RFC] Remove the Cr DTOs from the Blueprint

andreaTP:
The current state is more like a proposal rather than a real PR.
Here we remove the redundant CR specification from the `blueprint` making the blueprint itself a thin layer to check the streamlets connection layer.

The blueprint becomes based on the CR.


andreaTP:
No plans to implement this in the near future

PR #1017, opened 2021-04-08
Add a flag to disable Flink in the operator

andreaTP:


tools/cloudflow-operator/src/main/scala/cloudflow/operator/action/CloudflowStatus.scala
```diff
@@ -47,7 +47,7 @@ object CloudflowStatus {
   }
 
   private def podReady(ps: App.PodStatus) = {
-    ps.status == PodStatus.Running && ps.nrOfContainersReady == ps.nrOfContainers && ps.nrOfContainers > 0
+    (ps.status == PodStatus.Running && ps.nrOfContainersReady == ps.nrOfContainers && ps.nrOfContainers > 0) || (ps.nrOfContainers == 0 && ps.nrOfContainersReady == 0)
```
RayRoestenburg, line 50: This does not seem right..

tools/cloudflow-operator/src/main/scala/cloudflow/operator/action/CloudflowStatus.scala
```diff
@@ -57,7 +57,7 @@ object CloudflowStatus {
       nrOfContainersReady: Int,
       nrOfContainers: Int): App.PodStatus = {
     val ready =
-      if (nrOfContainersReady == nrOfContainers && nrOfContainers > 0) PodStatus.ReadyTrue else PodStatus.ReadyFalse
+      if ((nrOfContainersReady == nrOfContainers && nrOfContainers > 0)) PodStatus.ReadyTrue else PodStatus.ReadyFalse
```
RayRoestenburg, line 60: not needed?

tools/cloudflow-operator/src/main/scala/cloudflow/operator/action/CloudflowStatus.scala
```diff
@@ -170,11 +175,25 @@ object CloudflowStatus {
...
-        streamletName = deployment.streamletName,
-        expectedPodCount = Some(expectedPodCount),
-        podStatuses = Nil)
+      if (deployment.runtime == FlinkRunner.Runtime && !runners.contains(FlinkRunner.Runtime)) {
+        App.StreamletStatus(
+          streamletName = deployment.streamletName,
+          expectedPodCount = Some(0),
+          podStatuses = Seq(
+            App.PodStatus(
+              name = "<external>",
```
RayRoestenburg, line 184: why not depend on this `"<external>"` string to identify the podstatus that should be treated differently?

tools/cloudflow-operator/src/main/scala/cloudflow/operator/action/CloudflowStatus.scala
RayRoestenburg: IIRC the nrOfContainers can be 0 at first. If this is just for flink status when flink is disabled, rather use the "<external>" marker string to decide this.

RayRoestenburg (changes requested):
We need to discuss the ready logic.

tools/cloudflow-operator/src/main/scala/cloudflow/operator/action/CloudflowStatus.scala
```diff
@@ -102,8 +102,10 @@ object CloudflowStatus {
       nrOfContainers = nrOfContainers)
   }
 
-  private def getStatusFromContainerStates(containerStates: List[ContainerState], nrOfContainers: Int): String =
-    if (containerStates.nonEmpty) {
+  private def getStatusFromContainerStates(containerStates: List[ContainerState], nrOfContainers: Int): String = {
+    if (nrOfContainers == 0) {
```
RayRoestenburg, line 106: We would prefer to show as pending? not sure about this change.

tools/cloudflow-operator/src/main/scala/cloudflow/operator/action/CloudflowStatus.scala
andreaTP: fixed

tools/cloudflow-operator/src/main/scala/cloudflow/operator/action/CloudflowStatus.scala
andreaTP: yes

tools/cloudflow-operator/src/main/scala/cloudflow/operator/action/CloudflowStatus.scala
andreaTP: reverted

andreaTP:
should be good to go now @RayRoestenburg

RayRoestenburg (approved):

PR #1023, opened 2021-04-12
Introduce a flag in CLI to disable runtime checks

andreaTP:


tools/cloudflow-cli/src/test/scala/akka/cli/cloudflow/CliWorkflowSpec.scala
```diff
@@ -160,6 +160,18 @@ class CliWorkflowSpec extends AnyFlatSpec with Matchers with TryValues {
     res.failure.exception.getMessage.contains("required version") shouldBe true
   }
 
+  it should "succeed a mocked deploy if flink is un unamanged runtime" in {
```
RayRoestenburg, line 163:
```suggestion
  it should "succeed a mocked deploy if flink is an unmanaged runtime" in {
```

RayRoestenburg (approved):

PR #1037, opened 2021-04-27
Remove magnolia

andreaTP:
Ref: https://github.com/scala-open-letter/scala-open-letter.github.io

### DIFF
--- a/tools/build.sbt
+++ b/tools/build.sbt
@@ -69,6 +69,8 @@ lazy val cloudflowCli =
           // TODO: possibly to be removed
           "--allow-incomplete-classpath",
           "--initialize-at-run-time" + Seq(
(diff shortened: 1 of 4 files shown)


SethTisue:
LGTM (and I'm glad it appears to be that easy), but I see CI isn't quite on board yet.

SethTisue:
Have you measured any impact on compile times? That's typically been the hook for projects to adopt Magnolia, is decreased compile times.

andreaTP:
@SethTisue I just forgot to make the necessary changes into an auxiliary project, nothing major.

Yes, the compilation time will basically double on the cloudflow-config subproject but it doesn't spill to users in any way and, locally, the cache works well.

debasishg (approved):
LGTM

PR #1041, opened 2021-04-29
The glorious return back to core

andreaTP:
_"Everything must change for everything to remain the same"_
_Il Gattopardo_

TODO:
before this PR gets merged the release process should be tweaked and we should verify that all the artifacts are published as expected.


debasishg:
@andreaTP One general observation. Lots of changes (in fact most of them) comes from `scalafmt` version and rule changes. It would have been easier to check if these formatting changes were in a separate PR.

debasishg (approved):
Almost all of the PR is refactoring code from `tools` to `core` and changes in formatting through the new rules and version of `scalafmt`. LGTM

andreaTP:
@debasishg you are right but it would have taken me double the time to keep scalafmt in one commit, unfortunately 😢  sorry.

Minor caveats:
 - cannot get javaformatter to reliably work cross Java 8 - 11 have had to disable it
 - bumped the entire codebase to ScalaTest 3.2 -> this bump is possibly going to break user's tests and they are forced to bump
 - we would need to re-enforce back some `scalacOptions`.

PR #1043, opened 2021-05-05
extract a CR generator application

andreaTP:
Extract a self-contained CR Generator utility that can be used on its own.
It comes with a basic API `generate` method, and a simple and basic scopt command line helper.

Generate the uber-jar with:
```
sbt cloudflow-cr-generator/assembly
```

the application classpath can be extracted with a simple sbt task like:

```scala
val writeClasspath = taskKey[Unit]("write the application classpath")
writeClasspath := {
  val classpath =
    (Compile / fullClasspath).all(ScopeFilter(inAnyProject)).value.flatten.map(_.data.toURI.toURL).distinct.mkString("\n")

  IO.write(file("classpath.txt"), classpath)
}
```

and the command line can be invoked:

```
java -jar \
  cloudflow-cr-generator-assembly-<...>.jar \
  -n <application-name> \
  -v <application-version> \
  -b <blueprint.conf> \
  -c <classpath.txt> \
  -i <deployment-id>=<docker-image-name>,...
```


core/cloudflow-cr-generator/src/main/scala/cloudflow/cr/Generator.scala
```diff
@@ -0,0 +1,137 @@
...
+      import cloudflow.blueprint.deployment.CloudflowCRFormat._
+      val cloudflowCr = makeCR(appDescriptor, version)
+      cloudflowCr.toJson.compactPrint
+    } else {
+      throw new Exception(s"Found errors: ${blueprint.problems.mkString}")
+    }
+  }
+
+  case class Config(
+      name: String = null,
```
debasishg, line 85: nitpick: can make `final case class Config`. Purely my personal choice - I try to make everything `final` unless forced otherwise. But feel free to ignore :-)

core/cloudflow-cr-generator/src/main/scala/cloudflow/cr/Generator.scala
```diff
@@ -0,0 +1,137 @@
...
+          if (c.name == null || c.version == null || c.blueprint == null || c.classpath == null)
+            failure("All the options should be filled in")
+          else success))
+  }
+
+  def main(args: Array[String]): Unit = {
+
+    OParser.parse(parser, args, Config()) match {
+      case Some(config) =>
+        val result = generate(
```
debasishg, line 123: nitpick: (blame it on my love for pure FP). Can we make `generate` return an `Either` instead of throwing exceptions. Then the `main` can deterministically have an exit value under all circumstances. Currently when `generate` throws we don't have an exit value and the caller will have to catch the exception.

```scala
type Error = String
def generate(..): Either[Error, String] = //..
```

and we return a `Left` for error with detailed messages and a `Right` for the actual value.

Again: Just a suggestion - feel free to ignore

debasishg (approved):
Some minor (optional) changes suggested, otherwise LGTM

core/cloudflow-blueprint/src/main/scala/cloudflow/blueprint/deployment/ApplicationDescriptor.scala
```diff
@@ -54,14 +54,28 @@ object ApplicationDescriptor {
       blueprint: VerifiedBlueprint,
       agentPaths: Map[String, String],
       libraryVersion: String): ApplicationDescriptor = {
+    apply(appId, appVersion, (_: String) => image, blueprint, agentPaths, libraryVersion)
+  }
+
+  def apply(
+      appId: String,
+      appVersion: String,
+      images: String => String,
```
RayRoestenburg, line 63: ?

andreaTP: Is just more generic: I can implement this method with `Map[String, String].apply(key)` as well as a `TypesafeConfig.getString(key)`

core/cloudflow-cr-generator/src/main/scala/cloudflow/cr/Generator.scala
andreaTP: done

core/cloudflow-cr-generator/src/main/scala/cloudflow/cr/Generator.scala
andreaTP: encapsulated the call site into a `Try` with a nicer error handling.

A couple of considerations:
 - this is, of course, still half baked, the `Main` is probably going to be removed at the end of the process
 - the `generate` method might be called by Java code and I want to keep the API as simple and agnostic as possible

core/cloudflow-blueprint/src/main/scala/cloudflow/blueprint/deployment/ApplicationDescriptor.scala
RayRoestenburg: minor, would prefer a name like `imageSelector` or `getImageByKey` or something like that

core/cloudflow-cr-generator/src/main/scala/cloudflow/cr/Generator.scala
```diff
@@ -0,0 +1,144 @@
...
+      val appDescriptor = {
+        val appVersion = version
+        val agentPathsMap = Map("prometheus" -> "/prometheus/jmx_prometheus_javaagent.jar")
+        val libraryVersion = cloudflowVersion
+
+        ApplicationDescriptor(
+          appId = projectId,
+          appVersion = version,
+          (k) => dockerImages(k),
+          blueprint.verified.right.get,
```
RayRoestenburg, line 72: You could have used the either returned by verified instead of throwing exception?

RayRoestenburg: ah wait... I get it now... possible Java usage. never mind!

RayRoestenburg (approved):
LGTM

PR #1044, opened 2021-05-05
WIP: initial/partial maven support

andreaTP:
This is a single streamlet mvn project that can already perform `buildApp` (using `buildApp.sh`) and the generated CR can actually be deployed.

For building the docker image I used `io.fabric8.docker-maven-plugin`

Note that it depends on the #1043 jar


examples/mvn-hello-world/hello-world/src/main/scala/helloworld/HelloWorldShape.scala
```diff
@@ -0,0 +1,22 @@
...
+  def createLogic = new RunnableGraphStreamletLogic() {
+    def runnableGraph = {
+      Source
+        .cycle(() => Iterator.continually("hello world!"))
+        .throttle(1, 1.second)
+        .map(println)
+        .to(Sink.ignore)
+    }
+  }
+}
```
debasishg, line 22: May be I am missing something. But aren't we talking about maven support for Java projects ? If I can write Scala, then I can very well use sbt.

andreaTP: We want to support Maven builds whatever the target language is going to be, this can easily be in Java ...

RayRoestenburg (commented):
I love the direction this is going into!

PR #1051, opened 2021-05-18
Fix strimzi cr

andreaTP:
Fix documentation on installing Strimzi (outdated since last week).
(Reflect #1050 )


docs/shared-content-source/docs/modules/administration/pages/how-to-install-and-use-strimzi.adoc
```diff
@@ -63,15 +63,20 @@ spec:
     config:
       auto.create.topics.enable: false
       log.message.format.version: "2.3"
-      log.retention.bytes: 1073741824
-      log.retention.hours: 1
-      log.retention.check.interval.ms: 300000
+      log.retention.bytes: 112197632
```
RayRoestenburg, line 66: Less space?

RayRoestenburg (approved):
Not sure if needed to reduce size was this intentional? Otherwise LGTM

docs/shared-content-source/docs/modules/administration/pages/how-to-install-and-use-strimzi.adoc
andreaTP: restored, was accidental

PR #1059, opened 2021-05-27
Validate all example and sample applications

marcospereira:

Some of them were not compiled or tested. The script now discovers all apps and tests them.

To make sure all apps are in good shape, compiling, and building as expected.

No API changes at all.

Running `./scripts/build-all.sh test` so far.


examples/snippets/modules/ROOT/examples/akkastreams-grpc-java/src/main/java/sensordata/Logger.java
```diff
@@ -5,16 +5,18 @@
...
 import sensordata.grpc.SensorData;
 
 public class Logger extends AkkaStreamlet {
-    private final ProtoInlet<SensorData> inlet = new ProtoInlet<SensorData>(
+    private final ProtoInlet<SensorData> inlet = new ProtoInlet<>(
             "in",
             SensorData.class,
-            true
+            true,
+            CodecInlet::logAndSkip
```
marcospereira, line 19: The code here was not compiling. I'm not sure this is the correct fix for it, though.

examples/snippets/modules/ROOT/examples/base-images-override/build.sbt
```diff
@@ -3,8 +3,9 @@ import sbt.Keys._
 
     //tag::docs-projectSetup-example[]
 lazy val sampleApp = (project in file("."))
+    .enablePlugins(CloudflowApplicationPlugin)
     .settings(
-      cloudflowAkkaBaseImage := "myRepositoryUrl/myRepositoryPath:2.0.10-cloudflow-akka-2.6.6-scala-2.12",
+      cloudflowDockerBaseImage := "myRepositoryUrl/myRepositoryPath:2.0.10-cloudflow-akka-2.6.6-scala-2.12",
```
marcospereira, line 8: I'm not sure if there are other places in docs where we should recommend the new sbt setting instead of the deprecated ones.

scripts/build-all.sh
```diff
@@ -25,16 +25,16 @@ then
...
   echo "==================================================================================="
-  exit -1
+  exit 1
 fi
 echo "========================================================================="
 echo "Runs 'sbt $TARGET' for core and examples"
 echo "========================================================================="
 
 cd $DIR/../core
-sbt -mem 2048 --supershell=false "; scalafmtCheck ; $TARGET  ; +publishLocal"
```
marcospereira, line 35: I got an OOM error when running locally. Maybe this is a too aggressive bump?

scripts/build-all.sh
```diff
@@ -45,24 +45,32 @@ echo "Now starting building of examples..."
...
 
 for prj in $PROJECTS; do
   echo "========================================================================="
   echo "${TARGET}: $prj"
   echo "========================================================================="
 
-  cd $prj
+  pushd "$prj"
+
+  # Only run `verifyBlueprint` task if there are blueprint files in the project
```
marcospereira, line 58: Not all examples have a `blueprint.conf` file.

marcospereira (commented):
This was sitting on my machine since yesterday's night. There are some other minor changes to make all the samples run in the CI. I will push them when I have the time.

examples/snippets/modules/ROOT/examples/akkastreams-grpc-java/src/main/java/sensordata/Logger.java
RayRoestenburg: There is a Java API for this, `ProtoInlet.create` (it automatically uses `logAndSkip`), which is probably better for the example.

marcospereira: Resolved when rebasing with the main branch.

.github/workflows/build-pr.yaml
```diff
@@ -2,34 +2,30 @@ name: Build PRs
...
+    
     - name: Setup java
       uses: joschi/setup-jdk@v2
       with:
         java-version: ${{ matrix.java }}
         architecture: x64
+    
     - name: Cache SBT and Coursier cache
       uses: coursier/cache-action@v3
-    - name: Scalafmt check
```
marcospereira, line 20: This was duplicated with `push.yaml`, which also runs for pull requests.

.github/workflows/build-pr.yaml
```diff
@@ -2,34 +2,30 @@ name: Build PRs
...
-      run: |
-        cd core
-        sbt "scalafmtCheckAll; scalafmtSbtCheck"
-    - name: Build documentation
```
marcospereira, line 24: This was duplicated with `push.yaml`, which also runs for pull requests.

.github/workflows/build-pr.yaml
```diff
@@ -2,34 +2,30 @@ name: Build PRs
...
-        cd docs
-        make html-author-mode
-    - name: Run integration tests
+    
+    - name: Run scripted tests
       run: |
         git config --global user.email "cloudflow@lightbend.com"
         git config --global user.name "Cloudflow CI"
         cd core
-        sbt -mem 2048 +test
```
marcospereira, line 34: `core/tests` runs for `push.yaml` too.

.github/workflows/push.yml
```diff
@@ -23,23 +23,65 @@ jobs:
...
       env:
         SBT_OPTS: -Xms512M -Xmx2048M -Xss2M -XX:MaxMetaspaceSize=1024M
-      run: |
-        export PATH="$HOME/opt/sbt/bin:$PATH"
-        export SBT_OPTS="-Xss256M"
-        cd scripts
-        ./build-all.sh +test
+      run: ./scripts/build-core.sh +test
+
+  build-and-test-sbt-examples:
```
marcospereira, line 51: Split into different jobs so that they can run independently.

.github/workflows/push.yml
```diff
@@ -23,23 +23,65 @@ jobs:
...
+    name: "sbt samples"
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        java: [ '8', '11', '13' ]
```
marcospereira, line 57: Perhaps we don't want these samples to run on all these JDK versions.

core/build.sbt
```diff
@@ -176,6 +176,11 @@ setVersionFromTag := {
     .getOrElse("0.0.0-SNAPSHOT")}"""")
 }
 
+lazy val writeVersionToFile = taskKey[Unit]("Write project version to file")
```
marcospereira, line 179: This helps to have a consistent version that will be used when running the tests for all the samples. The file will later be read by the build shell scripts.

The goal here is to avoid tasks that change files to break dynver (used as a fallback by the samples) since it generates another version if the git repository is dirty.

RayRoestenburg:
Sure, let's look at these as possible improvements and pick it up in a support cycle. @andreaTP might also be interested in these changes

.github/workflows/build-pr.yaml
```diff
@@ -2,34 +2,30 @@ name: Build PRs
 on:
   pull_request:
 jobs:
-  build-all:
+  run-scripted-tests:
     runs-on: ubuntu-latest
     strategy:
+      fail-fast: false
```
andreaTP, line 8: I don't think we want to avoid the `fail-fast` semantic here

.github/workflows/build-pr.yaml
andreaTP: I agree on removing the duplicated `scalafmt` and `docs` build, but the `+test` is intentional and guarantees compatibility on the different JVMs

.github/workflows/build-test-tools.yml
```diff
@@ -6,14 +6,13 @@ on:
     branches: [main]
 
 jobs:
-  build:
-    runs-on: ubuntu-18.04
+  build-binary:
+    runs-on: ubuntu-latest
```
andreaTP, line 10: is this change needed, we should prefer stable identifiers

.github/workflows/build_and_publish.yaml
```diff
@@ -8,8 +8,7 @@ jobs:
   build-docs:
     runs-on: ubuntu-latest
     steps:
-    - name: Checkout
-      uses: actions/checkout@v2.3.3
+    - uses: actions/checkout@v2
```
andreaTP, line 11: Have you tested this workflow in isolation to verify this change?
I'm pretty sure we experienced a few regressions on the versioning here ...

.github/workflows/push.yml
```diff
@@ -23,23 +23,65 @@ jobs:
...
     steps:
-    - uses: actions/checkout@v2
 
-    - name: Setup environment
-      run: |
-        make -C .github/workflows/setup setup-local-all
```
andreaTP, line 33: good catch we can remove the entire `Makefile`:
https://github.com/lightbend/cloudflow/blob/main/.github/workflows/setup/Makefile

.github/workflows/push.yml
```diff
@@ -23,23 +23,65 @@ jobs:
...
-      run: |
-        export PATH="$HOME/opt/sbt/bin:$PATH"
-        export SBT_OPTS="-Xss256M"
-        cd scripts
-        ./build-all.sh +test
+      run: ./scripts/build-core.sh +test
+
+  build-and-test-sbt-examples:
+    name: "sbt samples"
+    runs-on: ubuntu-latest
```
andreaTP, line 53: we should prefer fixed versions

.github/workflows/push.yml
andreaTP: oh, it's done, nice, thanks!

core/build.sbt
andreaTP: Please remove this customization.

> The goal here is to avoid tasks that change files to break dynver

That's the reason why we should keep it as it was, the tasks should not change files to while running, otherwise, we want the CI to fail.

core/cloudflow-it/swiss-knife/project/cloudflow-plugins.sbt
```diff
@@ -1,5 +1,5 @@
 val latestVersion = {
-  sys.env.get("CLOUDFLOW_VERSION").fold(
+  sys.env.get("CLOUDFLOW_VERSION").filter(_.nonEmpty).fold(
```
andreaTP, line 2: have you done a global search and replace for this?

examples/snippets/modules/ROOT/examples/base-images-override/.scalafmt.conf
```diff
@@ -0,0 +1,44 @@
+version = "2.3.2"
```
andreaTP, line 1: is the `scalafmt` configuration all the same across the repo?
was, this one, missing?

examples/snippets/modules/ROOT/examples/base-images-override/build.sbt
andreaTP, line 8: I'm not sure what is going on here ...

examples/snippets/modules/ROOT/examples/base-images-override/project/build.properties
```diff
@@ -0,0 +1 @@
+sbt.version = 1.4.4
```
andreaTP, line 1: the rest of the repo standardize on `1.4.9`

scripts/functions.sh
```diff
@@ -0,0 +1,33 @@
+#!/usr/bin/env bash
+
+SCRIPTS_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd -P )"
+ROOT_DIR=$(dirname "$SCRIPTS_DIR")
+
+# Generates a stable version that will be used instead of reading 
+# from dynver. This is useful when running tasks that can change the
+# code for all the samples, for example, when running format tools.
+function echo_cf_version() {
```
andreaTP, line 9: if there are issues with dynver, let's analyze them and agree on solutions, but this should not be done in this way.

scripts/build-mvn-examples.sh
```diff
@@ -0,0 +1,72 @@
...
+
+for prj in "${PARENT_PROJECTS[@]}"; do
+  show_message "mvn ${TARGET}: $prj"
+  cd "$prj"
+
+  # Only run `cloudflow:verify-blueprint` task if there are blueprint files in the project
+  if [[ -n $(find . -name "blueprint.conf") ]]; then
+    # For more details about Maven workflow, see the docs:
+    # https://developer.lightbend.com/docs/cloudflow/shared-content/2.1.2/develop/maven-support.html
+    if ! mvn package cloudflow:extract-streamlets -DskipTests; then
```
andreaTP, line 57: Can we remove `-DskipTests` ?

andreaTP (changes requested):
I will review this better tomorrow, overall I really like the separation of the bash scripts, thanks @marcospereira !

The current issue with this PR is how it plays with `dynver` we should be able to avoid fixing the version.

.github/workflows/build-pr.yaml
marcospereira: Should we make the change there instead? I mean, make this run for multiple JVMs:

https://github.com/lightbend/cloudflow/blob/e9a48e29630df7f9629108539ed996d14eb89184/.github/workflows/push.yml#L26-L45

core/build.sbt
marcospereira: It will fail anyway, but instead of failing with unresolved artifacts for the next sample, it would fail for the right reason if that is the case.

This is also relevant when running the Maven examples where we need the `CLOUDFLOW_VERSION` env since we don't have dynver there. Or is there an option for Maven samples?

.github/workflows/build-pr.yaml
marcospereira: Fixed here: 637ca491f5b3efe7457d9186e6187d8ae9179e4d

.github/workflows/build_and_publish.yaml
marcospereira: I have not. How can I do that?

core/cloudflow-it/swiss-knife/project/cloudflow-plugins.sbt
marcospereira: Yes. :-)

examples/snippets/modules/ROOT/examples/base-images-override/.scalafmt.conf
marcospereira: They are not the same across the repo, but I'm pretty sure they are the same for all the samples. This one was copied from one of the other samples.

examples/snippets/modules/ROOT/examples/base-images-override/build.sbt
marcospereira: Avoid using the deprecated setting in the example.

examples/snippets/modules/ROOT/examples/base-images-override/project/build.properties
marcospereira: The vast majority of the examples are using 1.4.4, and this was probably just following one of them. I thought about changing all of them to use `1.4.9`, but this PR already has too many changes (this is why I've started https://github.com/lightbend/cloudflow/pull/1075).

scripts/functions.sh
marcospereira: There is no issue with dynver: if the repo is dirty it will generate a version that contains the time.

scripts/build-mvn-examples.sh
marcospereira: Tests will run later (see a little below where we run the target), so we don't need to run them here.

.github/workflows/build-pr.yaml
andreaTP: I have no strong opinions here, I think I'll prefer it here, but I won't disagree on moving it to `push.yml`

.github/workflows/push.yml
andreaTP: agreed samples can run on java 8 only

core/build.sbt
andreaTP: for Maven something like:
```bash
export CLOUDFLOW_VERSION=$(sbt -no-colors --error 'print version')
```
should make it work.

> It will fail anyway

If the git repo is not dirty the version is always the same.
But maybe I'm misunderstanding, can you elaborate? or maybe is faster to go through in a sync meeting?

.github/workflows/build_and_publish.yaml
andreaTP: you can hack the workflow in a branch to be triggered on `push` and verify that the output (e.g. the version) is correct and cosistent

examples/snippets/modules/ROOT/examples/base-images-override/.scalafmt.conf
andreaTP: thanks!

examples/snippets/modules/ROOT/examples/base-images-override/project/build.properties
andreaTP: fine with it

scripts/functions.sh
andreaTP: correct, and building/running examples should not result in a dirty repo

scripts/build-sbt-examples.sh
```diff
@@ -0,0 +1,60 @@
...
+
+  case "$prj" in
+    *-java)
+      if ! sbt -mem 4096 --supershell=false "; $TARGET; $blueprintTask"; then
+        show_message "Failed to run $TARGET and other checks for $prj"
+        exit 1
+      fi
+      ;;
+    *)
+      if ! sbt -mem 4096 --supershell=false "; scalafmtCheckAll ; $TARGET; $blueprintTask"; then
```
andreaTP, line 54: minor, can we append `scalafmtCheckAll` the same way as `blueprintTask`? I find it much cleaner.

marcospereira: Do you mean to run `scalafmtCheckAll` only if there is a `.scalafmt.conf` file?

andreaTP: I mean using this pattern:
```bash
  # Only run `verifyBlueprint` task if there are blueprint files in the project
  blueprintTask=""
  if [[ -n $(find . -name "blueprint.conf") ]]; then
    blueprintTask="verifyBlueprint"
  fi
```

and removing the `  case "$prj" in`

marcospereira: Done. See https://github.com/lightbend/cloudflow/pull/1059/commits/0a60944d528585bbbb4df9367037ca4319391c00.

scripts/functions.sh
marcospereira: Done. See https://github.com/lightbend/cloudflow/pull/1059/commits/2b356c499ba8328da59804f1a05f3f174b133337.

andreaTP: nice 👍

scripts/build-sbt-examples.sh
andreaTP: nice 👍

andreaTP (approved):
Added one commit to test examples only on Java 8.

Thanks for the hard work @marcospereira ! LGTM!

PR #1068, opened 2021-06-18
bump: Akka-gRPC to 2.0.0

andreaTP:
Note that this is going to be incompatible with previous versions, we need at least to bump `minor` after we merge.

### DIFF
--- a/examples/snippets/modules/ROOT/examples/build-flink-streamlets-java/build.sbt
+++ b/examples/snippets/modules/ROOT/examples/build-flink-streamlets-java/build.sbt
@@ -37,22 +37,6 @@ lazy val step2 = appModule("step2")
       Test / fork := true
     )
 
-// [info] * org.scala-lang:scala-library:2.12.11
-// [info] * org.apache.avro:avro:1.8.2
-// [info] * com.google.protobuf:protobuf-java:3.11.4
-// [info] * com.lightbend.akka.grpc:akka-grpc-runtime_2.12:1.0.2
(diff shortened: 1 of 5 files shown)


examples/snippets/modules/ROOT/examples/build-flink-streamlets-java/build.sbt
```diff
@@ -40,6 +40,7 @@ lazy val step2 = appModule("step2")
 lazy val step3 = appModule("step3")
     .enablePlugins(CloudflowFlinkPlugin)
     .settings(
+      dependencyOverrides += "com.lightbend.akka.grpc" %% "akka-grpc-runtime" % "1.0.3",
```
andreaTP, line 43: since we are going to bump the version and deprecate Flink/Spark integrations this should be ok.

RayRoestenburg: Have you tested that this still works?

RayRoestenburg: The 'deprecate Flink/Spark integrations' need some nuance, since we will point users to cloudflow-contrib and use that. Is the issue with the example?

andreaTP: > Have you tested that this still works in any case?

with this dependency override the tests are running and passing (haven't tested on a cluster).

>  Is the issue with the example?

The classpath

RayRoestenburg (approved):

PR #1075, opened 2021-07-23
Use sbt 1.5

marcospereira:

Update sbt version to 1.5.5 and use its new slash syntax.

No.

### DIFF
--- a/core/build.sbt
+++ b/core/build.sbt
@@ -316,7 +316,7 @@ lazy val cloudflowAkkaTestkit =
       crossScalaVersions := Vector(Dependencies.Scala212, Dependencies.Scala213),
       scalafmtOnCompile := true,
       javacOptions ++= Seq("-Xlint:deprecation", "-Xlint:unchecked"),
-      (sourceGenerators in Test) += (avroScalaGenerateSpecific in Test).taskValue)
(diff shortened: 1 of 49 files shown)


RayRoestenburg:
@marcospereira Looking at it again, the currently deprecated, but still existing `cloudflow-spark` library depends on Spark 2.4.5. The new cloudflow-contrib library depends on Spark 3.1.2. `sparksql-scalapb` has specific versions for both of these Spark versions, the version in this PR is for Spark 3.x. Once we remove cloudflow-spark completely, in `3.x`, we can remove these dependencies. I guess in the meantime we will have to deal with some eviction errors.

marcospereira:
> I guess in the meantime we will have to deal with some eviction errors, and not update the sparksql-scalapb support, since otherwise the deprecated support will not work.

I will take a look at it later because what the eviction is doing is basically warning us that the newer version is used instead of what is declared by Spark 2.4.5. Meaning that, in the end, we may need to instead do some dependencies override.

RayRoestenburg:
@marcospereira yes that sounds like a solution. (though it has to work such that if you use cloudflow-contrib, it overrides it with the spark 3.1.2 compatible libs, of course)

ennru:
Superseded by #1115 AFAICS

PR #1078, opened 2021-07-27
Fix GraalVM configuration for EKS

andreaTP:
List needed reflection configuration for GraalVM

Fix default access to EKS clusters (using `ExecCredentials`)

No

Tested against an EKS cluster with k8s 1.20 (validated on GKE as well)


core/tooling/src/main/resources/akka-microservices-crd.yml
```diff
@@ -0,0 +1,466 @@
+apiVersion: apiextensions.k8s.io/v1
+kind: CustomResourceDefinition
```
RayRoestenburg, line 2: Why is this crd here?

core/tooling/src/main/scala/cli/CodepathCoverageMain.scala
```diff
@@ -46,6 +46,10 @@ object CodepathCoverageMain extends App {
   }
   preparationClient.resourceList(sparkPvc).inNamespace("swiss-knife").createOrReplace()
 
+  val microservicesCRD =
```
RayRoestenburg, line 49: There is also a cloudflow crd now you could maybe use?

RayRoestenburg (commented):
Left some comments, can you use the cloudflow crd for the test?

core/tooling/src/main/resources/akka-microservices-crd.yml
andreaTP: Because the GraalVM automatic configuration should reach as many code-paths as possible, I forgot to add this step to the training while integrating the Microservices support

core/tooling/src/main/scala/cli/CodepathCoverageMain.scala
andreaTP: We should use both to ensure coverage.

andreaTP:
> Left some comments, can you use the cloudflow crd for the test?

That's what we are doing, adding the microservices CRD only to fix the GraalVM configuration generation (e.g. it's in the `tooling` sub-project)

core/tooling/src/main/resources/akka-microservices-crd.yml
RayRoestenburg: ok

RayRoestenburg (approved):

PR #1088, opened 2021-08-18
Fix docs publishing

marcospereira:

Fix docs publishing. It was silently failing in our CI jobs.

The job for publishing the docs is failing due to an error on a `yq` expression. 

No.

Running `cd docs && make all` locally.

### DIFF
--- a/.github/workflows/build_and_publish.yaml
+++ b/.github/workflows/build_and_publish.yaml
@@ -21,6 +21,7 @@ jobs:
     - name: Build documentation  
       run: |
         sudo snap install yq
+        yq --version
         cd docs
         make all
     - name: Publish to cloudflow.io  
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -114,7 +114,7 @@ list-todos: html
 # Generate the ScalaDoc and the JavaDoc, for the old versions
 all_javascaladocs:
 	@echo "Building old java and scala docs"
-	(yq e '.content.sources[0].branches.*' docs-source/site.yml | xargs -n1 ./build-javascaladocs.sh)
+	./build-javascaladocs.sh
 
 print-site:
 	# The result directory with the contents of this build:
--- a/docs/build-javascaladocs.sh
+++ b/docs/build-javascaladocs.sh
@@ -1,23 +1,58 @@
-#! /bin/bash
+#!/usr/bin/env bash
 
-VERSION=$1
+SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+ROOT_DIR=$( cd "${SCRIPT_DIR}/.." && pwd )
+DOCS_DIR="$ROOT_DIR/docs"
 
-echo $VERSION
+# Keep this consistent with the `work_dir` in the Makefile.
+TARGET_DIR="$DOCS_DIR/target"
+CLOUDFLOW_CLONE_DIR="${TARGET_DIR}/cloudflow"
 
-if [[ -z "${VERSION}" ]]; then
-  echo "Version not defined"
-else
-  (cd target && \
-    rm -rf cloudflow && \
-    git clone --depth=100 --branch "$VERSION" https://github.com/lightbend/cloudflow.git && \
-    cd cloudflow && \
-    export DIR=$(yq e '.version' docs/docs-source/docs/antora.yml) && \
-    echo $DIR && \
-    cd core && \
-    (sbt -mem 2048 clean unidoc || true) && \
-    cd ../../../ && \
-    ls -al "./target/cloudflow/core/target" && \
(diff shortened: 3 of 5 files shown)


marcospereira:
Failure example: https://github.com/lightbend/cloudflow/runs/3314872960?check_suite_focus=true#step:5:130

.github/workflows/build_and_publish.yaml
```diff
@@ -21,6 +21,7 @@ jobs:
     - name: Build documentation  
       run: |
         sudo snap install yq
+        yq --version
```
marcospereira, line 24: To debug the exact version that was installed.

docs/Makefile
```diff
@@ -114,7 +114,7 @@ list-todos: html
 # Generate the ScalaDoc and the JavaDoc, for the old versions
 all_javascaladocs:
 	@echo "Building old java and scala docs"
-	(yq e '.content.sources[0].branches.*' docs-source/site.yml | xargs -n1 ./build-javascaladocs.sh)
```
marcospereira, line 117: The fix starts here: `.content.sources[0].branches.*` is not a valid expression. It should be `.content.sources[0].branches[]` as done here:

https://github.com/lightbend/cloudflow/pull/1088/files#diff-c9354f203689690fae42eccaf90c1b0879bb7a4efe5ba5d5666a69e0b6e0e4efR17

But I'm moving all of this to the script.

docs/build-javascaladocs.sh
```diff
@@ -1,23 +1,58 @@
...
+
+  mv -v "$CLOUDFLOW_CLONE_DIR/core/target/scala-2.12/unidoc" "$TARGET_DIR/staging/docs/$DIR/api/scaladoc"
+  mv -v "$CLOUDFLOW_CLONE_DIR/core/target/javaunidoc"        "$TARGET_DIR/staging/docs/$DIR/api/javadoc"
+done
+
+# So that we can at least log the errors, but it won't fail the script.
+if [ ${#FAILURES[@]} -gt 0 ]; then
+  printf "Failed to generate docs for version %s\n" "${FAILURES[@]}"
+  cd "$ROOT_DIR" || true # get back home
+  exit 0
```
marcospereira, line 57: Should we fail the script? I guess it is better than keeping the publishing process going and fail silently.

docs/shared-content-source/docs/modules/develop/pages/build-flink-streamlets.adoc
```diff
@@ -33,7 +33,7 @@ Java::
 --
 [source,java]
 ----
-include::{cloudflow-examples-version}@docsnippets:ROOT:example$build-flink-streamlets-java/step0/src/main/java/com/example/ReportPrinter.java[]
```
marcospereira, line 36: Fixing some includes.

docs/build-javascaladocs.sh
andreaTP: > the publishing process going and fail silently

this was done (sadly) on purpose, some of the old versions of Cloudflow return an error code from the `sbt unidoc` command despite producing the actual documentation.

Now we removed some old versions and it might work, have you tested it in CI @marcospereira ?

docs/build-javascaladocs.sh
```diff
@@ -1,23 +1,58 @@
...
+mkdir -p -v "$TARGET_DIR"
+
+git clone https://github.com/lightbend/cloudflow.git "$CLOUDFLOW_CLONE_DIR"
+
+FAILURES=()
+
+for version in $versions; do
+  echo "Building docs for version $version"
+  cd "$CLOUDFLOW_CLONE_DIR" || exit 1 # Maybe if failed to clone the cloudflow repo.
+  git checkout --quiet "$version"
```
andreaTP, line 36: Changing the version using a simple `checkout` would be nice, have you checked that there are no residual files in the target directories? I recall having troubles with cleanups ...

andreaTP (changes requested):
Thanks for looking into this issue @marcospereira !
Left a couple of comments, but let's fix this!

docs/build-javascaladocs.sh
marcospereira: Yeah, I would say let's remove these old versions since they are not even maintained, right?

> have you tested it in CI @marcospereira ?

No. Shall we remove the old-failing versions before doing that?

docs/build-javascaladocs.sh
andreaTP: "should" ok, let's give this approach a go.

docs/build-javascaladocs.sh
andreaTP: > No. Shall we remove the old-failing versions before doing that?

I don't know if we should remove version e.g. `1.3.3` can you check if/on which versions the script is failing?

PR #1094, opened 2021-08-30
Option for the operator to control a single namespace

andreaTP:
With: https://github.com/lightbend/cloudflow-helm-charts/pull/23

### DIFF
--- a/core/cloudflow-operator/src/main/scala/cloudflow/operator/Settings.scala
+++ b/core/cloudflow-operator/src/main/scala/cloudflow/operator/Settings.scala
@@ -23,6 +23,7 @@ import com.typesafe.config._
 import io.fabric8.kubernetes.api.model.Quantity
 
 import scala.io.{ BufferedSource, Source }
+import scala.util.{ Success, Try }
 
 object Settings extends ExtensionId[Settings] with ExtensionIdProvider {
   override def lookup = Settings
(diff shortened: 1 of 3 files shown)


core/cloudflow-operator/src/main/scala/cloudflow/operator/Settings.scala
```diff
@@ -131,6 +132,11 @@ final case class Settings(config: Config) extends Extension {
   val flinkEnabled = config.getBoolean(s"$root.flink-enabled")
   val sparkEnabled = config.getBoolean(s"$root.spark-enabled")
 
+  lazy val controlledNamespace: Option[String] = Try(sys.env("CONTROLLED_NAMESPACE")) match {
```
RayRoestenburg, line 135: what about getting it from config and adding a default in the config lie we do in other places?

andreaTP: We can, but, in this specific case, I decided to use an env variable to make it easier and clean in the helm chart.
e.g. https://github.com/lightbend/cloudflow-helm-charts/pull/23/files#diff-08bec0d8da17f5482f6840b57e144767f8250cafd31b80a37e13e494d2236b71R5

Having the default config would require some additional logic to be implemented there.

RayRoestenburg: ok

RayRoestenburg: (the env var could have been in the reference.conf so it is known what is used, but this is also in the code so it does not really matter)

RayRoestenburg (approved):

core/cloudflow-operator/src/main/scala/cloudflow/operator/Settings.scala
RayRoestenburg: (I guess I'm thinking of this as, the reference.conf tells me what can be configured, or which env props can be used to change things, but this is a minor detail)

andreaTP: agree that would be nicer, I'll bind it to the operator configuration 👍

andreaTP: @RayRoestenburg done here:
https://github.com/lightbend/cloudflow/pull/1094/commits/c7e5386844e0dfa178ea53117c24e90884304969

RayRoestenburg: Thanks!

PR #1097, opened 2021-08-30
Encode the protocol version in the CRD

andreaTP:
Move the version validation to the CRD from the ConfigMap.

### DIFF
--- a/core/cloudflow-cli/src/main/scala/akka/cli/cloudflow/Cli.scala
+++ b/core/cloudflow-cli/src/main/scala/akka/cli/cloudflow/Cli.scala
@@ -16,9 +16,7 @@ object Cli {
   def defaultKubeClient(config: Option[File], logger: CliLogger) =
     new KubeClientFabric8(config)(logger)
 
-  val ProtocolVersion = "6"
-
-  val SupportedApplicationDescriptorVersion = 6
+  val ApplicationDescriptorVersion = 6
 
   val RequiredSparkVersion = "v1beta2"
 
--- a/core/cloudflow-cli/src/main/scala/akka/cli/cloudflow/kubeclient/KubeClientFabric8.scala
+++ b/core/cloudflow-cli/src/main/scala/akka/cli/cloudflow/kubeclient/KubeClientFabric8.scala
@@ -211,25 +211,16 @@ class KubeClientFabric8(
 
   def getOperatorProtocolVersion(): Try[String] = withClient { client =>
     for {
-      protocolVersionCM <- Try {
+      crd <- Try {
         client
-          .configMaps()
-          .inAnyNamespace()
-          .withLabel(KubeClient.CloudflowProtocolVersionConfigMap)
-          .list()
-          .getItems()
+          .apiextensions()
+          .v1beta1()
+          .customResourceDefinitions()
+          .withName(App.ResourceName)
+          .get()
       }
-      protocolVersion <- {
-        protocolVersionCM.size() match {
-          case 1 => Success(protocolVersionCM.get(0))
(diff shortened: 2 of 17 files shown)


core/cloudflow-cli/src/main/scala/akka/cli/cloudflow/kubeclient/KubeClientFabric8.scala
```diff
@@ -212,25 +212,16 @@ class KubeClientFabric8(
...
-        protocolVersionCM.size() match {
-          case 1 => Success(protocolVersionCM.get(0))
-          case x if x > 1 =>
-            Failure(
-              CliException("Multiple Cloudflow operators detected in the cluster. This is not supported. Exiting"))
-          case x if x < 1 => Failure(CliException("No Cloudflow operators detected in the cluster. Exiting"))
-        }
+          .apiextensions()
+          .v1beta1()
+          .customResourceDefinitions()
```
RayRoestenburg, line 219: Does this need cluster admin / wide permissions?

RayRoestenburg: It's probably a good idea to have a yaml that specifies exactly which permissions CLI needs to function, similar to https://github.com/lightbend/cloudflow-helm-charts/blob/main/cloudflow/templates/02-cloudflow-operator-clusterrole.yaml, what do you think?

core/cloudflow-crd/src/main/scala/akka/datap/crd/App.scala
```diff
@@ -46,6 +46,9 @@ object App {
 
   final val Scope = "Namespaced"
 
+  final val ProtocolVersionKey = "protocol-version"
+  final val ProtocolVersion = "7"
```
RayRoestenburg, line 50: 6, not 7?

RayRoestenburg: (or update all to 7, is maybe better, since the way that the protocol version is checked has changed)

core/cloudflow-cli/src/main/scala/akka/cli/cloudflow/kubeclient/KubeClientFabric8.scala
andreaTP: that's a good idea, have to think how to produce it, maybe running the cli from within the cluster and checking permissions.

core/cloudflow-crd/src/main/scala/akka/datap/crd/App.scala
andreaTP: done

core/cloudflow-cli/src/main/scala/akka/cli/cloudflow/Cli.scala
```diff
@@ -16,9 +16,7 @@ object Cli {
   def defaultKubeClient(config: Option[File], logger: CliLogger) =
     new KubeClientFabric8(config)(logger)
 
-  val ProtocolVersion = "6"
-
-  val SupportedApplicationDescriptorVersion = 6
+  val ApplicationDescriptorVersion = 6
```
RayRoestenburg, line 19: 7?

andreaTP: even `ApplicationDescriptor` ?

RayRoestenburg: Oh wait. No you are right, this stays the same.

PR #1101, opened 2021-09-03
Move the Protocol version to a secret

andreaTP:
This change will enable having different Cloudflow versions deployed in separate namespaces in a cluster

### DIFF
--- a/core/cloudflow-cli/src/main/scala/akka/cli/cloudflow/kubeclient/KubeClientFabric8.scala
+++ b/core/cloudflow-cli/src/main/scala/akka/cli/cloudflow/kubeclient/KubeClientFabric8.scala
@@ -227,7 +227,7 @@ class KubeClientFabric8(
           case 1 => Success(protocolVersionSecret.get(0))
           case x if x > 1 =>
             Failure(
-              CliException("Multiple Cloudflow operators detected in the cluster. This is not supported. Exiting"))
+              CliException("Multiple Cloudflow operators detected in the cluster. Specify an 'operator-namespace' to select the correct one. Exiting"))
           case x if x < 1 => Failure(CliException("No Cloudflow operators detected in the cluster. Exiting"))
         }
       }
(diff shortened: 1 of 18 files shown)


core/cloudflow-cli/src/main/scala/akka/cli/cloudflow/kubeclient/KubeClientFabric8.scala
```diff
@@ -209,18 +209,30 @@ class KubeClientFabric8(
...
+          .withLabel(App.CloudflowProtocolVersion)
+          .list()
+          .getItems()
+      }
+      protocolVersion <- {
+        protocolVersionSecret.size() match {
+          case 1 => Success(protocolVersionSecret.get(0))
+          case x if x > 1 =>
+            Failure(
+              CliException("Multiple Cloudflow operators detected in the cluster. This is not supported. Exiting"))
```
RayRoestenburg, line 230: It sort of is supported, but only when using `operatorNamespace`? (maybe specify that they must use the operatorNamespace flag instead?)

core/cloudflow-operator/src/main/scala/cloudflow/operator/Main.scala
```diff
@@ -61,6 +61,7 @@ object Main extends {
       checkCRD(settings, client)
 
       val ownerReferences = getDeploymentOwnerReferences(settings, client)
+      installProtocolVersion(settings, client, ownerReferences)
```
RayRoestenburg, line 64: What about adding this to helm chart?

RayRoestenburg: I guess the cloudflow currently already has permissions to create secrets in its own namespace, so is fine like this too. But feels a bit like a helm thing. (I'm ok what is done here, just a comment)

RayRoestenburg (approved):
One comment about feedback when more than one protocol version is found, if they did not specify the operatorNamespace, we could provide better feedback.

core/cloudflow-operator/src/main/scala/cloudflow/operator/Main.scala
andreaTP: make sense, I just restored the old way of doing it, for now, doing it in code makes us faster to change it, but helm makes sense too.

RayRoestenburg:
fails on format

PR #1105, opened 2021-09-09
Fix namespace and operatorNamespace propagation

andreaTP:


core/cloudflow-cli/src/main/scala/akka/cli/cloudflow/execution/DeployExecution.scala
```diff
@@ -204,7 +204,7 @@ final case class DeployExecution(d: Deploy, client: KubeClient, logger: CliLogge
         if (d.noRegistryCredentials) Success(())
         else {
           client.createImagePullSecret(
-            namespace = name,
+            namespace = namespace,
```
andreaTP, line 207: this was the bug

RayRoestenburg: Ah good catch @andreaTP !

PR #1108, opened 2021-10-19
fix: bumping up version

franciscolopezsancho:
As per documentation https://github.com/strimzi/strimzi-kafka-operator/releases with the new release, the Kafka CR doesn't work any more with version 2.7.1. Deploy a CR with that version raises an exception that states that 2.7.1 is not supported anymore and you need to choose among 2.8.0, 2.8.1, and 3.0.0.
I have chosen 3.0.0 only to be the latest, for no other reason.


franciscolopezsancho:
hey, @andreaTP for some reason I thought the CI would do that. No worries I can test that now.

RayRoestenburg:
@franciscolopezsancho please run the integration tests to be sure, the CI only runs scripted tests, so it will not hit the strimzi install.

PR #1111, opened 2021-10-20
removing version so it sets it to latest

franciscolopezsancho:
I tested it with `preparing-cluster`. The created default version was 3.0.0 in this case.


franciscolopezsancho:
I'm fine either way, too. Let's make @RayRoestenburg  decide then :). The other PR is [here](https://github.com/lightbend/cloudflow/pull/1110) Ray.

PR #1114, opened 2021-11-06
Support for ConfigMap volumes

leozilla:

Adds support for using k8s `ConfigMap` for k8s `Volume`s. Atm only `Secret` and `Pvc` are possible.

See: zulip thread: [Why is logging conf stored in a secret instead of ConfigMap](https://cloudflow.zulipchat.com/#narrow/stream/263236-cloudflow/topic/Why.20is.20logging.20conf.20stored.20in.20a.20secret.20instead.20of.20ConfigMap.3F)

Yes, it adds a new config feature.

Only tested via unit tests.

### Documentation

I did not adjust the documentation yet cos I first want to get some feedback on the PR if this feature makes sense in the way I am proposing it.

### Open questions

1. I did not test the writing of the `CloudflowConfig` yet and I am pretty sure this will not work at the moment cos `exportWriter` most likey does not treat the `ConfigMapVolumeKeyToPath` correctly. I did only see very few tests for writing the `CloudflowConfig`, thats why I am wondering if its even necessary?

### DIFF
--- a/core/cloudflow-config/src/test/scala/akka/cloudflow/config/CloudflowConfigSpec.scala
+++ b/core/cloudflow-config/src/test/scala/akka/cloudflow/config/CloudflowConfigSpec.scala
@@ -9,9 +9,9 @@ import scala.jdk.CollectionConverters._
 import scala.annotation.nowarn
 import akka.datap.crd.App
 import com.fasterxml.jackson.module.scala.DefaultScalaModule
-import com.typesafe.config.ConfigFactory
+import com.typesafe.config.{Config, ConfigFactory, ConfigObject, ConfigValueType}
 import io.fabric8.kubernetes.client.utils.Serialization
-import org.scalatest.{ OptionValues, TryValues }
+import org.scalatest.{Assertion, OptionValues, TryValues}
 import org.scalatest.flatspec.AnyFlatSpec
 import org.scalatest.matchers.should.Matchers
 import pureconfig.ConfigSource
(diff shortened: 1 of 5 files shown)


core/cloudflow-operator/src/test/scala/cloudflow/operator/action/runner/AkkaRunnerSpec.scala
```diff
@@ -105,7 +105,7 @@ class AkkaRunnerSpec
       crd.getSpec.getTemplate.getMetadata.getLabels.get("key2") mustBe "value2"
     }
 
-    "read from config custom annotaions and add them to the pod spec" in {
+    "read from config custom annotations and add them to the pod spec" in {
```
RayRoestenburg, line 108: nice

RayRoestenburg:
Thanks for the PR @leozilla , looks like it is going into the right direction! Please see the other export writers, and yes, likely you need to do something extra. It would be great if you could add a test for writing out a config which has config map volumes as well.

core/cloudflow-config/src/test/scala/akka/cloudflow/config/CloudflowConfigSpec.scala
```diff
@@ -893,6 +924,75 @@ class CloudflowConfigSpec extends AnyFlatSpec with Matchers with OptionValues wi
...
+    def shouldBeConfigMap(
+        configMapCfg: Config,
+        name: String,
+        optional: Boolean = false,
+        items: Map[String, String] = Map.empty): Assertion = {
+      val configMap = configMapCfg.getConfig("config-map")
+      configMap.getString("name") shouldBe name
+      configMap.getBoolean("optional") shouldBe optional
+
+      if (items.nonEmpty) {
```
leozilla, line 973: This assertion is a bit ugly, I have not found a nicer way to write it.

RayRoestenburg (approved):

